PR #207, opened 2022-12-27
PRT-206 add lava txs to e2e lava over lava txs validate they happened on chain

kajeagentspi:
This PR does the following.
- Create proposal, vote and add providers and clients for a new network.
- Check if the txs are reflected onchain.
- Change the kill style for the proxy routine to interrupt so it properly gives up the port.
- Only print the errors from cmd.Wait() if the test is not done successfully. The current one prints out signal is killed errors because finishTestSuccessfully() kills them first before realizing the test is done because of the cmd.Wait().


Aleksao998 (dismissed):
Thanks for fixing and answering my comments 🚀

ranlavanet:
@kajeagentspi Please resolve conflicts so we can merge

ranlavanet (approved):

PR #219, opened 2023-01-03
Add common allow-list for epochErrors

Aleksao998:
# Description

If no pairings available error happens, that would mean that the pairings will not be available at least till the next epoch.
When we send the first request (when we don't have available pairings) we would get 5 error logs:

1. context deadline exceeded
2. context deadline exceeded
3. No pairings available.
4. No pairings available.
5. No pairings available.

Context deadline exceeded error will appear only a certain number of times (i think 10 per epoch) after which the portal would stop sending req to the provider (would blocklist it) and then we would only see no pairings available errors

Ideally, we would log these errors only once, and for every other call in the epoch we would still return N**o pairings available error** to the user but we don't want to print these errors every time. The reason for this is that the portal can get overwhelmed with error logs.

# Changes include

- [x] Bugfix (non-breaking change that solves an issue)
- [ ] Hotfix (change that solves an urgent issue, and requires immediate attention)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (change that is not backwards-compatible and/or changes current functionality)

### Manual tests

1. Start a chain
2. init_chain_command
3. stop providers for specific chain (for example ETH)
4. Send relays to the ETH portal multiple times, zou see no pairing available **only first time**
`curl -X POST -H "Content-Type: application/json"  127.0.0.1:3333/1 --data '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}'`

Bonus:
1. Start a chain
2. Start portal without providers
3. Rename file to .sh and run the script it will spam requests and you should only see for the first request no pairing available and for others in the epoch only context deadline
[spam.txt](https://github.com/lavanet/lava/files/10337366/spam.txt)


ranlavanet (changes requested):
Lets investigate the possibility of context attributes or a utils singleton.

Aleksao998:
Closing this PR, we decided to go with completely different solution

PR #227, opened 2023-01-08
Add files via upload

KokosP:
Addr book Lava_Network


omerlavanet:
we don't want to store chain specific data on the main branch of the repository, we have release branches, and a separate repo for genesis, and seed node data

PR #235, opened 2023-01-11
CNS-140,CNS-150: dockerize lavad build and execution

orenl-lava:
Update the Makefile and intorduce Dockerfile to improve lavad build.
    
Notable make targets (for all targets - see Makefile):
    
  make build                    - local build (output in ./build/)
  make test                     - run unit-tests
  make lint                     - run the linter
    
  make docker-build             - local build with docker image (host arch)
   
  make build-images             - build and produce docker images (all archs)
  make build-image-amd64        - build and produce docker image (linux/amd64)
  make build-image-amd64        - build and produce docker image (linux/arm64)
    
Other notable changes:
    
- See Makefile for options to turn on/off via LAVA_BUILD_OPTIONS variable
- Use LAVA_BUILD_OPTIONS="static" to build static binary
- Use LAVA_BUILD_OPTIONS="static,release" to build static binary for release
    
Also updated .dockerignore (notably, added .git/ to avoid the cost). Anyway, the copy only affects the builder stage in docker - nothing is copied to the final image.

Also updates .gitignote (primarily some cleanup).

- [ ] Commit message follows the Contribution Guidelines
- [ ] Tests ran locally and added/modified if needed

* **Other information**:


docker/start_node.sh
```diff
@@ -0,0 +1,99 @@
...
+
+    # remove old data (if any)
+    rm -rf ${setup_config_dir}
+
+    # download setup configuration
+    git clone --depth 1 ${SETUP_CONFIG_GIT_URL} ${setup_config_dir} || \
+	error "setup: failed to clone setup configuration"
+
+    cd ${setup_config_dir}/testnet-1
+    . setup_config/setup_config.sh
```
SeanZoR, line 29: setup config sh, is too blackboxed.
How does the user select which network to connect?

docker/start_node.sh
```diff
@@ -0,0 +1,99 @@
+#!/bin/sh
+# vim:sw=4:ts=4:et
+
+set -e
+
+SETUP_CONFIG_GIT_URL='https://github.com/K433QLtr6RA9ExEq/GHFkqmTzpdNLDd6T.git'
```
SeanZoR, line 6: The git, along with it's structure, may change. Lets wait for that first.it will effect the script lines below also I believe

docker/start_node.sh
orenl-lava: The setup follows the public instructions at https://docs.lavanet.xyz/testnet-manual-cosmovisor
I'd like to get this in so users can enjoy it, and we can add customization later.

SeanZoR (dismissed):

docker/start_node.sh
orenl-lava: I agree that it would be helpful to make it more transparent - and allow more control.
Let's get this first version in, get more feedback, and then make improvements.

docker/start_node.sh
orenl-lava: Yes, that should be improved in the next iteration.

Dockerfile
```diff
@@ -0,0 +1,173 @@
+# syntax = docker/dockerfile:1.2
+# WARNING! Use with `docker buildx ...` or `DOCKER_BUILDKIT=1 docker build ...`
+# to enable --mount feature used below.
+
+########################################################################
+# Dockerfile for reproducible build of lavad binary and docker image
+########################################################################
+
+ARG GO_VERSION="1.18.2"
+ARG RUNNER_IMAGE="debian:11-slim"
```
praveenperera, line 10: @orenl-lava great Dockerfile! Thoughts on using alpine instead of debian-slim for the runner image? We'll get a smaller final image.

orenl-lava: Yes, that's the plan; I prefer to not delay this PR from getting to main, so deferred optimization until after the merge. 

Currently, our image size is ~256MB, of which the debian slim base (before `apt-get install ...`) is ~81MB; For reference, the alpine base alpine is ~10MB (before `apk add ...`).

praveenperera: sounds good, approved

praveenperera (approved):

PR #236, opened 2023-01-12
PRT-273 when we have 0 providers left in pairing un block all providers and retry

ranlavanet:

Adding support for multiple no pairings resets. (after no pairings are available we try to go over the pairing list again)

currently, we reach a point where there is an issue and we ban the provider for the remaining of the epoch, this can result in many idle epochs waiting for an epoch to change to get a different pairing list. 

after the entire provider list is blocked, we reset them and try again, instead of waiting for the remaining of the epoch. 

added unit tests to stress test the system 

* **Other information**:

### DIFF
--- a/relayer/lavasession/consumer_session_manager.go
+++ b/relayer/lavasession/consumer_session_manager.go
@@ -47,7 +47,7 @@ func (csm *ConsumerSessionManager) UpdateAllProviders(ctx context.Context, epoch
 	csm.atomicWriteCurrentEpoch(epoch)
 
 	// Reset States
-	csm.validAddresses = make([]string, pairingListLength)
+	// csm.validAddresses length is reset in setValidAddressesToDefaultValue
 	csm.pairingAddresses = make([]string, pairingListLength)
 	csm.addedToPurgeAndReport = make(map[string]struct{}, 0)
 	csm.pairingAddressesLength = uint64(pairingListLength)
@@ -61,9 +61,14 @@ func (csm *ConsumerSessionManager) UpdateAllProviders(ctx context.Context, epoch
 		csm.pairingAddresses[idx] = provider.Acc
 		csm.pairing[provider.Acc] = provider
 	}
-	copy(csm.validAddresses, csm.pairingAddresses) // the starting point is that valid addresses are equal to pairing addresses.
+	csm.setValidAddressesToDefaultValue() // the starting point is that valid addresses are equal to pairing addresses.
 
 	return nil
+}
+
+func (csm *ConsumerSessionManager) setValidAddressesToDefaultValue() {
+	csm.validAddresses = make([]string, len(csm.pairingAddresses))
+	copy(csm.validAddresses, csm.pairingAddresses)
 }
 
 // reads cs.currentEpoch atomically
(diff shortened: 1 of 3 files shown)


relayer/lavasession/consumer_session_manager.go
```diff
@@ -75,11 +76,34 @@ func (csm *ConsumerSessionManager) atomicReadCurrentEpoch() (epoch uint64) {
 	return atomic.LoadUint64(&csm.currentEpoch)
 }
 
+// validating we still have providers, otherwise reset valid addresses list
+func (csm *ConsumerSessionManager) validatePairingListNotEmpty() uint64 {
+	csm.lock.RLock() // lock read to validate length
```
omerlavanet, line 81: this function layout is prune for errors, need to do:
bigFunc(){
should_reset := readLockedFunc(){
Rlock
defer Runlock
}
if should_reset {
WritelockedFunc(){
lock 
defer unlock
check again
do it
}

relayer/lavasession/consumer_session_manager.go
```diff
@@ -75,11 +76,34 @@ func (csm *ConsumerSessionManager) atomicReadCurrentEpoch() (epoch uint64) {
...
+	numberOfResets := csm.numberOfResets
+	if len(csm.validAddresses) == 0 {
+		csm.lock.RUnlock()                // unlock read
+		csm.lock.Lock()                   // lock write
+		if len(csm.validAddresses) == 0 { // re verify it didn't change while waiting for lock.
+			utils.LavaFormatWarning("Provider pairing list is empty, resetting state.", nil, nil)
+			csm.validAddresses = make([]string, csm.pairingAddressesLength)
+			copy(csm.validAddresses, csm.pairingAddresses) // reset the list of valid addresses
```
omerlavanet, line 89: initialising should be done with the same code from update providers, not a copy of the same lines, create a small function that creates a new valid addresses from pairingAddresses and call it here and in updateAllProviders

relayer/lavasession/consumer_session_manager.go
```diff
@@ -75,11 +76,34 @@ func (csm *ConsumerSessionManager) atomicReadCurrentEpoch() (epoch uint64) {
...
+	}
+	return numberOfResets
+}
+
 // GetSession will return a ConsumerSession, given cu needed for that session.
 // The user can also request specific providers to not be included in the search for a session.
 func (csm *ConsumerSessionManager) GetSession(ctx context.Context, cuNeededForSession uint64, initUnwantedProviders map[string]struct{}) (
 	consumerSession *SingleConsumerSession, epoch uint64, providerPublicAddress string, reportedProviders []byte, errRet error,
 ) {
+	numberOfResets := csm.validatePairingListNotEmpty() // if pairing list is empty we reset the state.
```
omerlavanet, line 105: no need to check it on every session, you slow down the main path, do it after an error happens in getting a session, it will run only if we really need this

omerlavanet (changes requested):

relayer/lavasession/consumer_session_manager.go
ranlavanet: We need to read the number of resets every time, its just a read lock at most cases which doesnt really slow it down as we only write in 2 occasions

relayer/lavasession/consumer_session_manager.go
```diff
@@ -76,23 +76,36 @@ func (csm *ConsumerSessionManager) atomicReadCurrentEpoch() (epoch uint64) {
...
+		csm.validAddresses = make([]string, csm.pairingAddressesLength)
+		copy(csm.validAddresses, csm.pairingAddresses) // reset the list of valid addresses
+		csm.numberOfResets += 1
+		numberOfResets = csm.numberOfResets // update numberOfResets with the new value
+	}
+	return numberOfResets
+}
+
+// validating we still have providers, otherwise reset valid addresses list
+func (csm *ConsumerSessionManager) validatePairingListNotEmpty() uint64 {
```
omerlavanet, line 105: nice and clean :)

relayer/lavasession/consumer_session_manager.go
```diff
@@ -76,23 +76,36 @@ func (csm *ConsumerSessionManager) atomicReadCurrentEpoch() (epoch uint64) {
...
+	csm.lock.Lock() // lock write
+	defer csm.lock.Unlock()
+	if len(csm.validAddresses) == 0 { // re verify it didn't change while waiting for lock.
+		utils.LavaFormatWarning("Provider pairing list is empty, resetting state.", nil, nil)
+		csm.validAddresses = make([]string, csm.pairingAddressesLength)
+		copy(csm.validAddresses, csm.pairingAddresses) // reset the list of valid addresses
+		csm.numberOfResets += 1
+		numberOfResets = csm.numberOfResets // update numberOfResets with the new value
+	}
+	return numberOfResets
```
omerlavanet, line 101: bug: if len != 0 number of resets is not the right value

omerlavanet (changes requested):

omerlavanet (approved):

PR #239, opened 2023-01-16
PRT-173 Make logging async using singleton

Aleksao998:

This is PR creates a singleton Logger which accepts log messages async. It also has a allow list, which can block some error messages to be print inside epoch with error code

Before adding this we had spam of error logs for some edge casses. FOr example if there are not available providers for an epoch, we know that no new providers will be available till next epoch so we don't need to print it every time

Inside logger we can allow-list some error codes

```
Start a chain
init_chain_command
stop providers for specific chain (for example ETH)
Send relays to the ETH portal multiple times, zou see no pairing available only first time
curl -X POST -H "Content-Type: application/json"  127.0.0.1:3333/1 --data '{"jsonrpc":"2.0","method":"eth_blockNumber","params":[],"id":1}'
```

* **Other information**:


utils/logger/logger.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/lavanet/lava/utils/allowList"
+	"github.com/rs/zerolog"
+	zerologlog "github.com/rs/zerolog/log"
+)
+
+var (
+	instance *Logger
+	once     sync.Once
+)
+
```
ranlavanet, line 16: lets add a comment describing what these numbers are for

utils/logger/logger.go
```diff
@@ -0,0 +1,120 @@
...
+func (l *Logger) printLogs(description string, logEvent *zerolog.Event) {
+	logEvent.Msg(description)
+}
+
+// Log function is to push the log messages to the channel
+func (l *Logger) Log(msg LogMessage) {
+	// Print a log if a channel gets full
+	if len(l.logChan) == cap(l.logChan) {
+		logEvent := zerologlog.Warn()
+		logEvent.Msg("Log channel is full")
```
ranlavanet, line 116: lets, for now, return without printing the log.

ranlavanet: lets also add everything to a go routine

ranlavanet (changes requested):

relayer/sentry/sentry.go
```diff
@@ -612,6 +613,12 @@ func (s *Sentry) Start(ctx context.Context) {
 				utils.LavaFormatInfo("New Epoch Event", nil)
 				utils.LavaFormatInfo("New Epoch Info:", &map[string]string{"Height": strconv.FormatInt(data.Block.Height, 10)})
 
+				// Get logger instance
+				logger := singletonLogger.GetInstance()
```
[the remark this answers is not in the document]
Aleksao998: Fixed, 1354966e1c5a66a0ad783c96aeff6f1b8defeffe

utils/logger/logger.go
Aleksao998: Fixed returning 02950e1f5a7f71ae9755d820a33b1488a0cda5b1, also changed the log to the error level so we would easier catch it.

The idea with go routines has a major flaw, if we put this in go routine, we can not guarantee that the logs will be printed in the same order as they arrived. This can mess up an order of logs which will make debugging impossible

utils/logger/logger.go
Aleksao998: Fixed 3dca57a241b5d5c718a4202326fd4c7ea711dae7

utils/allowList/allowList.go
```diff
@@ -0,0 +1,59 @@
...
+func (al *AllowList) IsErrorSet(code uint32) bool {
+	// Check if the pointer is nil.
+	if al == nil {
+		// If the pointer is nil, return false.
+		return false
+	}
+
+	// Check if the value for the specified error name is set to true.
+	if val, ok := (*al)[code]; ok && val {
+		return true
```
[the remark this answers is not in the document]
Aleksao998: FIxed ff1b98818ab2447a84b94aa17c79f9a3aa725a01

ranlavanet (dismissed):

Aleksao998:
Closing this PR for now, after creating a new rpcConsumer we don't have a need for allow list for logs and we don't want to add extra complexity without need for it.

If in the future we again get a need for it, this PR is completed so we can always re open it

PR #240, opened 2023-01-16
Update go.mod

freshe4qa:
important patches


ranlavanet:
go.mod is updated every time we merge a feature, no need to make a specific pull request for that :)

ranlavanet:
we are working with very specific versions, we cannot just upgrade cosmos-SDK without adjusting to all the requirements.

PR #245, opened 2023-01-19
PRT-318 adding print when max cu reached + adjusting tests

ranlavanet:


relayer/lavasession/consumer_session_manager.go
```diff
@@ -138,6 +138,7 @@ func (csm *ConsumerSessionManager) GetSession(ctx context.Context, cuNeededForSe
 				return nil, 0, "", nil, err
 			} else if MaxComputeUnitsExceededError.Is(err) {
 				// This provider doesn't have enough compute units for this session, we block it for this session and continue to another provider.
+				utils.LavaFormatError("Max Compute Units Exceeded For provider", err, &map[string]string{"providerAddress": providerAddress})
```
omerlavanet, line 141: add the CU used and CU requested and the CU limit here

omerlavanet (changes requested):

relayer/lavasession/consumer_session_manager.go
ranlavanet: cant, its inside the cswp. i can add another print if it finds an error and wrap it inside the error.

omerlavanet (approved):

PR #253, opened 2023-01-23
PRT-292 incorrect account sequence no payments to providers

ranlavanet:

Retry mechanism for sequence number errors on payment requests by the expected sequence number. 
Also, splitting the payment requests into a random block in an epoch instead of all payment requests on the first block.
currently fails and doesnt parse the errors correctly and always asking rewards on the first block in an epoch 

Retry mechanism for sequence number errors on payment requests by the expected sequence number. 
Also, splitting the payment requests into a random block in an epoch instead of all payment requests on the first block.

testing multiple payment requests from the same provider in different processes, resulting in reproducing the problem and fixing it 

* **Other information**:

### DIFF
--- a/relayer/server.go
+++ b/relayer/server.go
@@ -11,6 +11,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"regexp"
 	"strconv"
 	"strings"
 	"sync"
@@ -48,7 +49,7 @@ import (
 )
 
 const (
-	RETRY_INCORRECT_SEQUENCE      = 3
+	RETRY_INCORRECT_SEQUENCE      = 5
 	TimeWaitInitializeChainSentry = 10
 	RetryInitAttempts             = 10
 )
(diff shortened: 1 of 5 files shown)


relayer/server.go
```diff
@@ -313,7 +316,17 @@ func askForRewards(staleEpochHeight int64) {
...
 
-func summarizeTransactionResult(transactionResult string) (string, []string) {
+// extract requested sequence number from tx error.
+func findSequenceNumber(sequence string) (int, error) {
+	re := regexp.MustCompile(`expected (\d+), got (\d+)`)
+	match := re.FindStringSubmatch(sequence)
+	if match == nil {
+		return 0, utils.LavaFormatWarning("Failed to parse sequence number from error", nil, &map[string]string{"sequence": sequence})
+	}
+	return strconv.Atoi(match[1]) // atoi return 0 upon error, so it will be ok when sequenceNumberParsed uses it
```
omerlavanet, line 326: check match length before accessing index 1

relayer/server.go
```diff
@@ -322,7 +335,17 @@ func summarizeTransactionResult(transactionResult string) (string, []string) {
...
 		}
 	}
-	return summarizedResult, transactionResults
+
+	re := regexp.MustCompile(`code:(\d+)`) // extracting code from transaction result (in format code:%d)
+	match := re.FindStringSubmatch(transactionResult)
+	if match == nil {
+		return summarizedResult, 1 // not zero
+	}
+	retCode, err := strconv.Atoi(match[1]) // extract return code.
```
omerlavanet, line 344: check match length

omerlavanet (changes requested):

omerlavanet (approved):

PR #275, opened 2023-01-31
PRT-403 fixing the tendermint response to be identical to what the user expects

ranlavanet:


relayer/chainproxy/tendermintRPC.go
```diff
@@ -428,30 +526,30 @@ func (nm *TendemintRpcMessage) Send(ctx context.Context, ch chan interface{}) (r
...
+		if err != nil {
+			return nil, "", nil, utils.LavaFormatError("Failed parsing ID when getting rpc error", err, nil)
 		}
-		replyMsg.Error = &rpcclient.JsonError{
-			Code:    1,
-			Message: fmt.Sprintf("%s", err),
+		replyMsg = &RPCResponse{
+			JSONRPC: nm.msg.Version,
+			ID:      id,
+			Error:   convertErrorToRPCError(err),
```
omerlavanet, line 543: BUG: error here is always nil because you overwrite it with id, err := idFromRawMessage(nm.msg.ID) and bail out, did you mean to use the original error from the callContext?

omerlavanet (changes requested):

omerlavanet (dismissed):

omerlavanet (approved):

PR #288, opened 2023-02-05
CNS 250: Introduce Spec inheritance syntax

orenl-lava:
CNS-250: introduce spec inheritance syntax
    
Extend the spec syntax to add the ability for a spec to inherit/import another
spec. With spec inheritance, changes in an ancestor spec (a.k.a base spec)
would immediately propagate to all descendants specs that inherit from it.
    
Spec inheritance is recursive, and applied only the APIs defined (not general
spec parameters). APIs in current spec override inherited APIs, but duplicates
(from multiple ancestors) are not allowed. An override with 'Enabled: false'
effectively removes an API). Spec inheritance is valitaded when submitting a
spec proposal.

Tested with unit tests: `go test ./x/spec/keeper -v`
Tested with e2e tests: `go test ./testutil/e2e -v -timeout 1200s`
Tested with `scripts/init_chain_commands.sh`


proto/spec/spec.proto
```diff
@@ -11,6 +11,7 @@ import "spec/service_api.proto";
 message Spec {
   string index = 1; 
   string name = 2; 
+  repeated string imports = 12;
```
Yaroms, line 14: we have some conflict from main. some more fields were added to spec and the number 12 is taken :).
merge main to the branch and decide on the number, than generate the proto buffs.

x/spec/types/spec.go
```diff
@@ -7,6 +7,97 @@ import (
...
+type getterFunc func(index string) (Spec, bool)
+ 
+// ExpandSpec expands the "imports" field of the spec -if needed, recursively-
+// and adds to the current Spec those additional APIs from the imported Spec(s).
+// It takes as argument a getter function to fetch imported Spec(s) by name from
+// the keeper store, on demand. (Can't import the keeper module due to circular
+// dependencies)
+func (spec *Spec) ExpandSpec(getSpec getterFunc) (map[string]string, error) {
+	depends := map[string]bool{spec.Index: true}
+	details := map[string]string{"imports": fmt.Sprintf("%s", spec.Index)}
```
Yaroms, line 19: why do we need this as a map? it looks like it contains only one key value pair

x/spec/types/spec.go
```diff
@@ -7,6 +7,97 @@ import (
...
+		}
+	}
+
+	currentApis := make(map[string]bool)
+	for _, api := range spec.Apis {
+		currentApis[api.Name] = true
+	}
+
+	// merge collected APIs into current spec's APIs (unless overridden)
+	for _, api := range mergedApis {
```
Yaroms, line 92: we are missing the api deletion if it is disabled.
if the api is disabled dont add it to the list

Yaroms (changes requested):
amazing job

x/spec/types/spec.go
orenl-lava: That's what the caller expects, convention for reporting details of errors/events for printing.
Internally in `doExpandSpec()` we can maintain a single string - will fix.

x/spec/types/spec.go
orenl-lava: If an API is disabled in the parent, then it won't be added to the `mergedApis` list - see L75 above.

proto/spec/spec.proto
orenl-lava: Done

x/spec/types/spec.go
```diff
@@ -7,6 +7,97 @@ import (
...
 
+type getterFunc func(index string) (Spec, bool)
+
+// ExpandSpec expands the "imports" field of the spec -if needed, recursively-
+// and adds to the current Spec those additional APIs from the imported Spec(s).
+// It takes as argument a getter function to fetch imported Spec(s) by name from
+// the keeper store, on demand. (Can't import the keeper module due to circular
+// dependencies)
+func (spec *Spec) ExpandSpec(getSpec getterFunc) (map[string]string, error) {
+	depends := map[string]bool{spec.Index: true}
```
omerlavanet, line 18: refactor: this is a confusing layout and complex implementation, passing callbacks without added value (it's always the same one too). instead of adding it as a method to the spec object and then have circular dependencies, if it accesses the keeper it should be implemented as a keeper function

x/spec/types/spec.go
```diff
@@ -7,6 +7,97 @@ import (
...
+	for _, index := range spec.Imports {
+		imported, found := getSpec(index)
+		// import of unknown Spec not allwoed
+		if !found {
+			details += fmt.Sprintf("%s%s(unknown)", comma, index)
+			return details, fmt.Errorf("imported spec unknown: %s", index)
+		}
+		// import of disabled Spec not allowed
+		if !imported.Enabled {
+			details += fmt.Sprintf("%s%s(disabled)", comma, index)
```
omerlavanet, line 45: importing disabled specs should be allowed, in case we want to inherit from some base spec that doesn't hold on it's own

x/spec/types/spec.go
```diff
@@ -7,6 +7,97 @@ import (
...
+	}
+
+	var parents []Spec
+
+	// visual markers when import deepens
+	details += "->["
+
+	// recursion to get all parent specs
+	comma := ""
+	for _, index := range spec.Imports {
```
omerlavanet, line 36: should be extracted in a method for readability

x/spec/types/spec.go
```diff
@@ -7,6 +7,97 @@ import (
...
+	details += "]"
+
+	var mergedApis []ServiceApi
+	mergedApisMap := make(map[string]ServiceApi)
+
+	// collect all parents' Specs' APIs
+	for _, imported := range parents {
+		for _, api := range imported.Apis {
+			if api.Enabled {
+				// duplicate API(s) not allowed
```
omerlavanet, line 76: there's a case we might have conflict but it's overwritten anyway by the inheriting spec.Apis, there is no need to bail here, since those won't be used anyway

omerlavanet (changes requested):

x/spec/types/spec.go
orenl-lava: ok

x/spec/types/spec.go
orenl-lava: Not sure I understand your comment - this is unrelated to overrides that the current Spec may (or may not) have.
The code detects cases where a Spec imports at once two or more Specs, which have APIs with the same name (but different details). The result of such import may be ambiguous (e.g. depending on the order), so we prohibit it.

x/spec/types/spec.go
orenl-lava: Can you detail you have in mind?  Moving the inside of this loop to a helper function, means that the recursion will ping-pong between this function and the helper, and I think that's less readable.

x/spec/proposal_handler.go
```diff
@@ -74,31 +74,61 @@ func NewSpecProposalsHandler(k keeper.Keeper) govtypes.Handler {
 	}
 }
 
+func validateSpec(ctx sdk.Context, k keeper.Keeper, spec types.Spec) (map[string]string, error) {
+	// NOTE: spec.ExpandSpec may modify spec.Apis and affect the caller's copy;
```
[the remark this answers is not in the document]
orenl-lava: Ok, will do. (For the record, It has to be subtle either way: if `spec.ExpandSpec()` returns a new `spec`, then when doing so it has to remember to make a copy of spec's APIs list).

x/spec/types/spec.go
omerlavanet: i would move it to a function that performs fetch parent apiInterfaces, its cleaner and clearer, there is no reason to recurse the combination of the apis to one list so you can only recurse that one

x/spec/types/spec.go
omerlavanet: in the case of both functions importing the same API and then its ambiguous but discarded due to an override, then it didn't matter there was ambiguity and therefore this check can fail when in fact we could have continued

x/spec/proposal_handler.go
```diff
@@ -74,31 +75,53 @@ func NewSpecProposalsHandler(k keeper.Keeper) govtypes.Handler {
 	}
 }
 
+func validateSpec(ctx sdk.Context, k keeper.Keeper, spec types.Spec) (map[string]string, error) {
```
Yaroms, line 78: let this be function of the spec keeper instead of passing him as an argument

Yaroms (changes requested):

x/spec/types/spec.go
orenl-lava: Done.

x/spec/types/spec.go
orenl-lava: The recursion is standard DFS and I prefer to not split into two functions that call each other.  Will try to move the spec fetch and validation to a helper function.

x/spec/types/spec.go
orenl-lava: Ok.

Yaroms (approved):

x/spec/keeper/spec.go
```diff
@@ -62,6 +65,108 @@ func (k Keeper) GetAllSpec(ctx sdk.Context) (list []types.Spec) {
...
+
+	currentApis := make(map[string]bool)
+	for _, api := range spec.Apis {
+		currentApis[api.Name] = true
+	}
+
+	// merge collected APIs into current spec's APIs (unless overridden)
+	for _, api := range mergedApis {
+		if _, found := currentApis[api.Name]; !found {
+			spec.Apis = append(spec.Apis, api)
```
omerlavanet, line 148: it's still replacing the input spec in-place, please create a separate spec that is expanded

x/spec/keeper/spec.go
```diff
@@ -62,6 +65,108 @@ func (k Keeper) GetAllSpec(ctx sdk.Context) (list []types.Spec) {
...
+	var mergedApis []types.ServiceApi
+	mergedApisMap := make(map[string]types.ServiceApi)
+
+	// collect all parents' Specs' APIs
+	for _, imported := range parents {
+		for _, api := range imported.Apis {
+			if api.Enabled {
+				// duplicate API(s) not allowed
+				if _, found := mergedApisMap[api.Name]; found {
+					return details, fmt.Errorf("duplicate imported api: %s (in spec: %s)", api.Name, imported.Index)
```
omerlavanet, line 132: overriden apis can still error out

x/spec/types/spec.go
```diff
@@ -7,6 +7,97 @@ import (
...
 const minCU = 1
 
+type getterFunc func(index string) (Spec, bool)
+
+// ExpandSpec expands the "imports" field of the spec -if needed, recursively-
+// and adds to the current Spec those additional APIs from the imported Spec(s).
+// It takes as argument a getter function to fetch imported Spec(s) by name from
+// the keeper store, on demand. (Can't import the keeper module due to circular
+// dependencies)
+func (spec *Spec) ExpandSpec(getSpec getterFunc) (map[string]string, error) {
```
omerlavanet, line 17: remove the expand code from types, and the getter func

omerlavanet (changes requested):

x/spec/keeper/spec.go
orenl-lava: No it does not. 
`ExpandSpec()` takes a copy of the original spec and returns a new copy.
`DoExpandSpec()` takes a reference to a spec and extends the API list of that spec; When it calls itself, it does so using a copy of the imported spec obtained from `k.GetSpec()`, so again not modifying an original.

x/spec/types/spec.go
orenl-lava: Sigh, no idea how it got back there.

omerlavanet (approved):

PR #298, opened 2023-02-08
PRT-XYZ Fix invalid parsing and add unit tests

Aleksao998:

Bug Fix

The response error data is not JSON compatible so clients can not unmarshal objects

fix the issue so the error data is json compatible

* **Other information**:


protocol/common/rpcconsumerlogs.go
```diff
@@ -57,14 +58,17 @@ func (pl *RPCConsumerLogs) GetMessageSeed() string {
 
 // Input will be masked with a random GUID if returnMaskedErrors is set to true
 func (pl *RPCConsumerLogs) GetUniqueGuidResponseForError(responseError error, msgSeed string) string {
-	var ret string
-	ret = "Error GUID: " + msgSeed
-	utils.LavaFormatError("UniqueGuidResponseForError", responseError, &map[string]string{"msgSeed": msgSeed})
+	data := map[string]interface{}{
```
omerlavanet, line 61: let's do this a deterministic serialization so we can do data reliability on it, so let's not marshal a map rather instead lets marshal a list of key value pairs

omerlavanet (changes requested):

protocol/common/rpcconsumerlogs.go
Aleksao998: I considered this while i was writing the code but the reason i decided to use map and not array of key value is how the user would need to unmarshal this error. If we use only map the response which the user will get will look like this:
```
{"Error":"height missing from map map[data:0a3f6f736d6f3138377033776e7268666b30376e756a367868386139637867643368686a6e687978327272786b3067786e666e72726e35723439737433676d616c path:/cosmwasm.wasm.v1.Query/ContractInfo prove:%!s(bool=false)]","Error_GUID":"GUID_7018205163"}
```
```
And if we use array of key value it wil look like this:
`[{"Error_GUID":"GUID_7604293667"},{"Error":"Failed all retries -- map[errors:Errors: [rpc error: code = DeadlineExceeded desc = Sending nodeMsg failed ErrMsg: Failed Sending Message ErrMsg: context deadline exceeded: context deadline exceeded: Failed Sending Message ErrMsg: context deadline exceeded: context deadline exceeded No pairings available.]]"}]
```

The map made more sense because this is how the users could unmarshall GUID and ERROR
```
parsedError = JSON.parse(err.message);
data = JSON.parse(parsedError.data);
console.log("Error", data.Error)
console.log("GUID", data.Error_GUID)
```

And if we go with array approach the unmarshalling will look like this:
```
parsedError = JSON.parse(err.message);
data = JSON.parse(parsedError.data)
console.log("Error", data[0].Error_GUID)
console.log("GUID", data[1].Error)
```
Accessing the array using index + referencing by key did not seem standard

On the other side i completely understand the argument that we want it to be deterministic, so if we decide with array approach i only wanted to point this

Aleksao998: also for data realibility we can unmarshall the struct and make sure that the values are same and not depend on ordering of the serialized data

ranlavanet: waiting for this @omerlavanet to merge.

omerlavanet: data reliability doesn't unmarshal the content of the response, deterministic is a requirement we have to fulfil, you can do any inner layout that makes sense, hardcoding the indexes looks pretty bad, why not use a proto marshaller? this is deterministic and also has named properties

Aleksao998: Fixed + added unit tests to make sure it is deterministic and also if something changes in the future we would see failed test c0019e09c5ab399fedbca468858d4a8e5801b303

protocol/common/rpcconsumerlogs_test.go
```diff
@@ -0,0 +1,94 @@
+package common
```
ranlavanet, line 1: need to add these to protocol.yaml

ranlavanet (changes requested):

protocol/common/rpcconsumerlogs_test.go
Aleksao998: Fixed 560efde2ef47fcaab162a1d8e14c032f62fe2c44

ranlavanet (approved):

omerlavanet (approved):

ranlavanet (approved):

PR #301, opened 2023-02-14
Feature/update relay metrics

tabakuj:
- [ ] Commit message follows the Contribution Guidelines
- [ ] Tests ran locally and added/modified if needed

* **Other information**:


relayer/chainproxy/chainproxy.go
```diff
@@ -316,11 +317,11 @@ func SendRelay(
...
 }
 
-func ConstructFiberCallbackWithDappIDExtraction(callbackToBeCalled fiber.Handler) fiber.Handler {
+func ConstructFiberCallbackWithHeaderAndParameterExtraction(callbackToBeCalled fiber.Handler) fiber.Handler {
 	webSocketCallback := callbackToBeCalled
 	handler := func(c *fiber.Ctx) error {
-		dappId := ExtractDappIDFromFiberContext(c)
-		c.Locals("dappId", dappId)
+		c.Locals("dappId", ExtractDappIDFromFiberContext(c))
+		c.Locals(refererHeaderKey, ExtractHeaderFiberContext(c, refererHeaderKey))
```
nirtayeb, line 324: where do you get refererHEaderKey from?

relayer/chainproxy/chainproxy.go
```diff
@@ -342,6 +343,18 @@ func ExtractDappIDFromFiberContext(c *fiber.Ctx) (dappID string) {
 	return dappID
 }
 
+func ExtractHeaderFiberContext(c *fiber.Ctx, headerKey string) string {
+	headers := c.GetReqHeaders()
+	headerValue, exist := headers[headerKey]
+	if !exist {
+		headerValue, exist = headers[strings.ToLower(refererHeaderKey)]
```
nirtayeb, line 350: where do you get refererHEaderKey from?

relayer/chainproxy/portalLogs.go
```diff
@@ -96,9 +103,38 @@ func (pl *PortalLogs) LogStartTransaction(name string) {
...
+		pl.metricService.SendData(*data)
 	}
 }
+
+func (pl *PortalLogs) ShouldCountMetric(c *fiber.Ctx) bool {
+	refererHeaderValue := c.Get(refererHeaderKey, "")
+	if refererHeaderValue == "" {
+		refererHeaderValue = c.Get(strings.ToLower(refererHeaderKey), "")
+	}
+	if len(pl.toExcludeMetricsReferrers) > 0 && refererHeaderKey != "" {
```
nirtayeb, line 118: this condition should be in one function and used by the 3 -"shouldCount" functions, so no code duplication and if we add/remove conditions we will do it on one place only

relayer/metrics/metricsService.go
```diff
@@ -89,11 +89,16 @@ func prepareArrayForProject(projectData map[string]map[string]*AggregatedMetric,
 	var toSendData []RelayAnalyticsDTO
 	for chainKey, chainData := range projectData {
 		for apiTypekey, apiTypeData := range chainData {
+			var averageLatency int64
+			if apiTypeData.RelaysCount > 0 {
```
nirtayeb, line 93: if all relays are failed, you still have division by zero

tabakuj: yes i misstyped the  if there

relayer/chainproxy/chainproxy.go
tabakuj: its a const on this package.

relayer/chainproxy/chainproxy.go
tabakuj: its a const on this package.

relayer/chainproxy/portalLogs.go
```diff
@@ -96,9 +106,42 @@ func (pl *PortalLogs) LogStartTransaction(name string) {
...
-		data.Success = !isNotSuccessful
-		pl.MetricService.SendData(*data)
+func (pl *PortalLogs) AddMetric(data *metrics.RelayMetrics, isSuccessful bool) {
+	if pl.storeMetricData {
+		data.Success = isSuccessful
+		pl.metricService.SendData(*data)
+	}
+}
+
+func (pl *PortalLogs) ShouldCountMetric(c *fiber.Ctx) bool {
```
nirtayeb, line 116: ShouldCountMetricForHTTP

relayer/metrics/metricsService.go
```diff
@@ -171,15 +177,15 @@ func (m *MetricService) storeChainIdData(projectData map[string]map[string]*Aggr
 	}
 }
 
-func (m *MetricService) storeApiTypeData(chainIdData map[string]*AggregatedMetric, data RelayMetrics, successCount int64) {
+func (m *MetricService) storeApiTypeData(chainIdData map[string]*AggregatedMetric, data RelayMetrics, successCount int64, successLatencyValue int64) {
 	apiTypesData, exists := chainIdData[data.APIType]
 	if exists {
-		apiTypesData.TotalLatency += data.Latency
+		apiTypesData.TotalLatency += successLatencyValue
```
ranlavanet, line 183: we are scared of overflow here. 
calculating average could be done differently by taking the current average with weight of successrelays and adding the latest relay latency with weight 1  
use this algorithm instead 

# initialize variables
n = 0  # number of samples seen so far
avg_latency = 0.0  # current average latency

# for each new latency value
n += 1
delta = latency - avg_latency  # compute difference from previous average
avg_latency += delta / n  # update the average latency

# the final average latency is stored in avg_latency

protocol/chainlib/rest.go
```diff
@@ -240,7 +242,9 @@ func (apil *RestChainListener) Serve(ctx context.Context) {
 		analytics := metrics.NewRelayAnalytics(dappID, chainID, apiInterface)
 
 		reply, _, err := apil.relaySender.SendRelay(ctx, path, query, http.MethodGet, dappID, analytics)
-		go apil.logger.AddMetric(analytics, err != nil)
+		if apil.logger.ShouldCountMetricForHttp(c) {
```
ranlavanet, line 245: make a wrapper method to wrap this check and runn addmetric inside, this method should run in a go routine instead the add metric one. so we do the check in a routine as well.

protocol/common/rpcconsumerlogs.go
```diff
@@ -110,9 +118,42 @@ func (pl *RPCConsumerLogs) LogStartTransaction(name string) {
...
+func (pl *RPCConsumerLogs) AddMetric(data *metrics.RelayMetrics, isSuccessful bool) {
 	if pl.StoreMetricData {
-		data.Success = !isNotSuccessful
+		data.Success = isSuccessful
 		pl.MetricService.SendData(*data)
 	}
 }
+
+func (pl *RPCConsumerLogs) ShouldCountMetricForHttp(c *fiber.Ctx) bool {
+	refererHeaderValue := c.Get(refererHeaderKey, "")
```
ranlavanet, line 129: why do we need to test both lower case and upper case? I wouldn't do both of them every time. 

can we decide on one of them that is valid and the other isn't?

protocol/chainlib/common.go
```diff
@@ -177,3 +179,15 @@ func verifyTendermintEndpoint(endpoints []string) (websocketEndpoint string, htt
...
+
+func ExtractHeaderFiberContext(c *fiber.Ctx, headerKey string) string {
+	headers := c.GetReqHeaders()
+	headerValue, exist := headers[headerKey]
+	if !exist {
+		headerValue, exist = headers[strings.ToLower(refererHeaderKey)]
+		if !exist {
+			headerValue = "NoValue"
+		}
+	}
```
nirtayeb, line 191: can be replaced with calling Get

https://docs.gofiber.io/api/ctx/#get

protocol/chainlib/common.go
```diff
@@ -177,3 +179,15 @@ func verifyTendermintEndpoint(endpoints []string) (websocketEndpoint string, htt
 	}
 	return websocketEndpoint, httpEndpoint
 }
+
+func ExtractHeaderFiberContext(c *fiber.Ctx, headerKey string) string {
+	headers := c.GetReqHeaders()
```
ranlavanet, line 184: @nirtayeb said there is a better way to extract the headers. please discuss

protocol/chainlib/common.go
```diff
@@ -59,9 +60,10 @@ func extractDappIDFromFiberContext(c *fiber.Ctx) (dappID string) {
 	return dappID
 }
 
-func constructFiberCallbackWithDappIDExtraction(callbackToBeCalled fiber.Handler) fiber.Handler {
+func constructFiberCallbackWithHeaderAndParameterExtraction(callbackToBeCalled fiber.Handler) fiber.Handler {
 	webSocketCallback := callbackToBeCalled
 	handler := func(c *fiber.Ctx) error {
+		c.Locals(refererHeaderKey, ExtractHeaderFiberContext(c, refererHeaderKey))
```
ranlavanet, line 66: do we even need the headers? we need to validate the environment variable.

protocol/chainlib/rest.go
```diff
@@ -204,7 +204,9 @@ func (apil *RestChainListener) Serve(ctx context.Context) {
 		utils.LavaFormatInfo("in <<<", &map[string]string{"path": path, "dappID": dappID, "msgSeed": msgSeed})
 		requestBody := string(c.Body())
 		reply, _, err := apil.relaySender.SendRelay(ctx, path, requestBody, http.MethodPost, dappID, metricsData)
-		go apil.logger.AddMetric(metricsData, err != nil)
+		if apil.logger.ShouldCountMetricForHttp(c) {
```
ranlavanet, line 207: for each one of the should please check the env if nil or not other wise we don't need to run it (another wrapper inside)

ranlavanet (changes requested):

relayer/metrics/metricsService.go
tabakuj: why do we need to calculate the avg everytime we have an insert that doesn't seem neccessary

protocol/chainlib/rest.go
tabakuj: i put everything inside add metric will push the changes now

relayer/metrics/metricsService.go
nirtayeb: @ranlavanet this is a formula for approximate an average and not for the actual average. running the number through excel gives wrong values.

We are not going to overflow the int64, 
we have a limit of maxCU per epoch, which limit the amount of relays
 and we are storing totals for 5min which is less than an epoch.
 
 we will change the type to uint instead int so we will have increased range.

relayer/metrics/metricsService.go
```diff
@@ -89,11 +89,16 @@ func prepareArrayForProject(projectData map[string]map[string]*AggregatedMetric,
 	var toSendData []RelayAnalyticsDTO
 	for chainKey, chainData := range projectData {
 		for apiTypekey, apiTypeData := range chainData {
+			var averageLatency int64
+			if apiTypeData.SuccessCount > 0 {
+				averageLatency = int64(apiTypeData.TotalLatency / uint64(apiTypeData.SuccessCount))
```
nirtayeb, line 94: mixing int and uint is a bad practice

nirtayeb: also, it should be a float...

tabakuj: sure i can make it all uint, but as for float its not necessary here since we are talking about ms (the floating part would be nano s ). do we need that precise ?

protocol/chainlib/common.go
```diff
@@ -177,3 +181,10 @@ func verifyTendermintEndpoint(endpoints []string) (websocketEndpoint string, htt
 	}
 	return websocketEndpoint, httpEndpoint
 }
+
+func storeRefererHeaderIfNeeded(c *fiber.Ctx) {
+	isMetricEnabled, _ := strconv.ParseBool(os.Getenv("IS_METRICS_ENABLED"))
```
ranlavanet, line 186: let's not parse this every relay, this is a handler you create one time but every time use. you can create it with the parsing and without it when you don't need it. parse the IS_METRICS_ENABLED outside the constructFiberCallbackWithHeaderAndParameterExtraction
and create the callback accordingly 

will be more efficient that way

protocol/chainlib/grpc.go
```diff
@@ -200,11 +200,12 @@ func (apil *GrpcChainListener) Serve(ctx context.Context) {
 		var relayReply *pairingtypes.RelayReply
 		metricsData := metrics.NewRelayAnalytics("NoDappID", apil.endpoint.ChainID, apiInterface)
 		if relayReply, _, err = apil.relaySender.SendRelay(ctx, method, string(reqBody), "", "NoDappID", metricsData); err != nil {
-			go apil.logger.AddMetric(metricsData, err != nil)
+			go apil.logger.AddMetricForGrpc(metricsData, false, ctx)
```
ranlavanet, line 203: can move this to right after sendRelay (before error handling) 
and parse the error inside, do this to all API's 

go apil.logger.AddMetricForGrpc(metricsData, err, ctx)

that way we can call it only once

protocol/chainlib/jsonRPC.go
```diff
@@ -227,7 +227,8 @@ func (apil *JsonRPCChainListener) Serve(ctx context.Context) {
 			defer cancel() // incase there's a problem make sure to cancel the connection
 			metricsData := metrics.NewRelayAnalytics(dappID, chainID, apiInterface)
 			reply, replyServer, err := apil.relaySender.SendRelay(ctx, "", string(msg), http.MethodGet, dappID, metricsData)
-			go apil.logger.AddMetric(metricsData, err != nil)
+			go apil.logger.AddMetricForWebSocket(metricsData, err == nil, c)
```
ranlavanet, line 230: if error handling is inside we dont need to parse the error here

protocol/chainlib/grpc.go
ranlavanet: saw you did that for the other api's so here as well maybe you forgot

protocol/chainlib/grpc.go
```diff
@@ -200,11 +200,12 @@ func (apil *GrpcChainListener) Serve(ctx context.Context) {
...
 			errMasking := apil.logger.GetUniqueGuidResponseForError(err, msgSeed)
 			apil.logger.LogRequestAndResponse("http in/out", true, method, string(reqBody), "", errMasking, msgSeed, err)
 			return nil, utils.LavaFormatError("Failed to SendRelay", fmt.Errorf(errMasking), nil)
 		}
+		go apil.logger.AddMetricForGrpc(metricsData, true, ctx)
```
ranlavanet, line 208: also, one thing to note here is that the context is canceled when we return the response (by the user) as they run in a routine, if you need a context. you would need to create a new background context inside the AddMetric.

ranlavanet: I saw you need this only for grpc for the metadata. as the context is canceled by the user you will need to extract the meta data before you start the routine, we are going to use the meta data later on anyway so you can just unload the entire meta data object to when we get the request and use the meta data in your go routine.

ranlavanet (changes requested):

protocol/chainlib/grpc.go
nirtayeb: do you mean to change the code like that?
```
relayReply, _, err = apil.relaySender.SendRelay(...)
go apil.logger.AddMetric(metricsData, err != nil)
if err != nill {
...
}
```

protocol/chainlib/common.go
nirtayeb: we should have i think on the MetricsService level, no?

tabakuj: we have on portal level ,will put it there also.

protocol/chainlib/grpc.go
tabakuj: nope i didn't forget, the code has a return inside the if (line 202)

nirtayeb:
lgtm - @ranlavanet wdyt?

ranlavanet (approved):

nirtayeb (approved):

PR #328, opened 2023-02-28
CNS-169: create module subscription

orenl-lava:
This PR holds the the result of the scaffolding module subscription.

The code was auto-generated using:
```ignite scaffold module subscription```
Pushed in a separate PR to simplify the review of the upcoming changes.

* **Please check if the PR fulfills these requirements**
- [x] Commit message follows the Contribution Guidelines
- [x] Tests ran locally and added/modified if needed
- [x] Docs have been added/updated, if applicable
- [x] If applicable - JIRA ticket ID was added

* **What kind of change does this PR introduce?** (Bug fix, feature, unit tests, docs update, ...)

* **What is the current behavior?** (You can also link to an open issue here)

* **What is the new behavior (if this is a feature change)?**

* **Please describe what manual tests you ran, if applicable**

* **Other information**:

PR #329, opened 2023-02-28
CNS-169 implement module subscription

orenl-lava:
Implement module subscription basics and the "subscribe" message,
This module depends on the Packages module, so tested with a local mock (not included in PR).
Note that the PR depends on #328 .

* **Other information**:


x/subscription/keeper/subscription.go
```diff
@@ -0,0 +1,198 @@
...
+	}
+
+	k.cdc.MustUnmarshal(b, &val)
+	return val, true
+}
+
+// RemoveSubscription removes the subscription (of a consumer) from the store
+func (k Keeper) RemoveSubscription(ctx sdk.Context, consumer string) {
+	sub, _ := k.GetSubscription(ctx, consumer)
+	packages.PutPackage(ctx, sub.PackageIndex, sub.Block)
```
oren-lava, line 39: In the latest "plans" implementation, we didn't include a plan wrapper for the PutEntry function (like you wrote). You obviously couldn't have known this, but commenting so you can fix after "plans" is merged.
Do you think we should implement plan wrappers to the common library's API?

x/subscription/keeper/subscription.go
```diff
@@ -0,0 +1,198 @@
...
+
+	return
+}
+
+// GetAllSubscription returns all subscription (of all consumers)
+func (k Keeper) GetAllSubscription(ctx sdk.Context) []types.Subscription {
+	return k.GetCondSubscription(ctx, nil)
+}
+
+func endOfMonth(date time.Time) time.Time {
```
oren-lava, line 68: maybe general time-related functions should be under common? (and not in subscription module)

oren-lava, line 68: Also, more of a general comment, in the "plans" implementation we've defined a month as 1440 blocks (assuming 1 block/min). I think both modules (and all other/future modules) should use time constants the same way. Which is the right way in your opinion?

x/subscription/keeper/subscription.go
```diff
@@ -0,0 +1,198 @@
...
+		if date.Day() == 28 && leapYear(date) {
+			next = date.AddDate(1, 0, 1)
+		}
+	}
+
+	return next
+}
+
+// CreateSubscription creates a subscription for a consumer
+func (k Keeper) CreateSubscription(ctx sdk.Context, sub types.Subscription) error {
```
oren-lava, line 112: who creates the "sub" argument? since this function get a subscription object, verifies it and sets it to the store, maybe call the function "VerifyAndSetSubscription"? or "SetSubscription" (the original "SetSubscription" will turn to "setSubscription").
The current name is kind of confusing since this function doesn't create a subscription object.

x/subscription/keeper/subscription.go
oren-lava: also, deal with error? what if a package (a.k.a. plan) with (sub.PackageIndex, sub.Block) doesn't exist?

x/subscription/types/keys.go
```diff
@@ -17,6 +17,22 @@ const (
 	MemStoreKey = "mem_subscription"
 )
 
+const (
+	// SubscriptionKeyPrefix is the prefix to retrieve all Subscription
+	SubscriptionKeyPrefix = "Subscribe/"
```
oren-lava, line 22: normally we'll use underscores (change to "Subscribe_")

x/subscription/types/keys.go
```diff
@@ -17,6 +17,22 @@ const (
...
 	return []byte(p)
 }
+
+// SubscriptionKey returns the store key to retrieve a Subscription from the consumer field
+func SubscriptionKey(consumer string) []byte {
+	var key []byte
+
+	indexBytes := []byte(consumer)
+	key = append(key, indexBytes...)
+	key = append(key, []byte("/")...)
```
oren-lava, line 35: change to "_"

x/subscription/types/message_subscribe.go
```diff
@@ -0,0 +1,58 @@
...
+
+func (msg *MsgSubscribe) Route() string {
+	return RouterKey
+}
+
+func (msg *MsgSubscribe) Type() string {
+	return TypeMsgSubscribe
+}
+
+func (msg *MsgSubscribe) GetSigners() []sdk.AccAddress {
```
oren-lava, line 31: why do you return an array if there's always one creator?

x/subscription/keeper/subscription_test.go
```diff
@@ -0,0 +1,187 @@
...
+		{
+			name:      "insufficient funds",
+			index:     "testpkg",
+			creator:   1,
+			consumers: []int{2},
+			success:   false,
+		},
+		{
+			name:      "invalid package",
+			index:     "",
```
oren-lava, line 153: this check will fail in the ValidateBasic() function since the index is empty. Maybe add a test with a non-empty plan index that doesn't exist?

x/subscription/types/subscription.go
```diff
@@ -0,0 +1,33 @@
...
+	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
+)
+
+func (sub Subscription) IsExpired(date time.Time) bool {
+	expiry := time.Unix(int64(sub.ExpiryTime), 0).UTC()
+	return expiry.Before(date)
+}
+
+// ValidateSubscription validates a subscription object fields
+func (sub Subscription) ValidateSubscription() error {
```
oren-lava, line 16: isn't this a duplicate of your ValidateBasic() function?

oren-lava (changes requested):

x/subscription/keeper/epoch_start.go
```diff
@@ -0,0 +1,22 @@
...
+	"github.com/lavanet/lava/x/subscription/types"
+)
+
+// EpochStart runs the functions that are supposed to run in epoch start
+func (k Keeper) EpochStart(ctx sdk.Context) {
+	// On epoch start we need to iterate through all the subscriptions and
+	// check if they just expired. Those that expire get deleted.
+
+	date := ctx.BlockTime().UTC()
+	subExpired := k.GetCondSubscription(ctx, func(sub types.Subscription) bool {
```
Yaroms, line 14: instead of iterating over all the subscriptions i think we should create an ordered list that contains the next subscription to expire, that way no iterating is required.

x/subscription/keeper/subscription.go
Yaroms: the definition of 1440 blocks are for 24 hours, this is the memory that a provider can claim rewards and the reason we keep old entries for that long

x/subscription/keeper/subscription.go
```diff
@@ -0,0 +1,198 @@
...
+
+func leapYear(date time.Time) bool {
+	year := date.Year()
+	return year%400 == 0 || (year%4 == 0 && year%100 != 0)
+}
+
+func nextMonth(date time.Time) time.Time {
+	next := date.AddDate(0, 1, 0)
+
+	if date.Day() == endOfMonth(date).Day() {
```
Yaroms, line 80: do we need this? maybe next := date.AddDate(0, 1, 0) is enough?
add a small explanation for this

x/subscription/keeper/subscription.go
```diff
@@ -0,0 +1,198 @@
...
+	return k.GetCondSubscription(ctx, nil)
+}
+
+func endOfMonth(date time.Time) time.Time {
+	return date.AddDate(0, 1, -date.Day())
+}
+
+func leapYear(date time.Time) bool {
+	year := date.Year()
+	return year%400 == 0 || (year%4 == 0 && year%100 != 0)
```
Yaroms, line 74: lets define constants for the 400, 4, 100, it is not clear to me what is this calculation

x/subscription/keeper/subscription.go
```diff
@@ -0,0 +1,198 @@
...
+		return endOfMonth(next)
+	}
+
+	return next
+}
+
+func nextYear(date time.Time) time.Time {
+	next := date.AddDate(1, 0, 0)
+
+	if date.Month() == 2 {
```
Yaroms, line 99: add some comments for this

x/subscription/types/message_subscribe.go
Yaroms: this is scaffolded function

x/subscription/types/subscription.go
Yaroms: validate basic is done for messages and proposals

proto/subscription/subscription.proto
```diff
@@ -0,0 +1,17 @@
...
+
+option go_package = "github.com/lavanet/lava/x/subscription/types";
+
+message Subscription {
+  
+  string creator = 1; 
+  string consumer = 2; 
+  int64 block = 3; 
+  string packageIndex = 4; 
+  int64 packageBlock = 5;
```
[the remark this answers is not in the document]
orenl-lava: To know which version of the plan we refer to. 
We don't keep a copy of the plan - don't need one. We took a reference with GetPlan() and when the subscription expires we will drop it with PutPlan().

proto/subscription/subscription.proto
```diff
@@ -0,0 +1,17 @@
...
+  bool isYearly = 6; 
+  uint64 expiryTime = 7; 
+  uint64 usedCU = 8;
```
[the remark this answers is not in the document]
orenl-lava: total = used + remaining
so we can keep either pair.

x/subscription/keeper/keeper.go
```diff
@@ -26,21 +30,31 @@ func NewKeeper(
...
+		epochstorageKeeper: epochstorageKeeper,
 	}
 }
 
 func (k Keeper) Logger(ctx sdk.Context) log.Logger {
 	return ctx.Logger().With("module", fmt.Sprintf("x/%s", types.ModuleName))
 }
+
+func (k Keeper) IsEpochStart(ctx sdk.Context) bool {
+	return k.epochstorageKeeper.GetEpochStart(ctx) == uint64(ctx.BlockHeight())
```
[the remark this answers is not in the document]
orenl-lava: Cannot, because epochstoragekeeper is a private member of (subscription) Keeper.
This code is copied from x/conflict and x/pairing.

x/subscription/keeper/subscription.go
orenl-lava: Yes, please add wrappers - see my review on the Plan PR.
An error here is a bad bug - we took a reference, we should always be able to drop it; And anyway, there is nothing we can do if that happens: panic or ignore. 
Likewise, PutPlan() should not return an error. If anything it - or the underlying FixateStore - should generate a log message error.

x/subscription/keeper/subscription.go
orenl-lava: This is standard way to determine a leap year.
Const here only used here, so adding them elsewhere will make it less readable.
Instead, I will add a comment and a reference for how this is calculated.

x/subscription/keeper/subscription.go
orenl-lava: date.AddDate() behavior is slightly different on month's ends. This makes sure that we don't advance to next-next month unintentionally. Will add comment and reference,

x/subscription/keeper/subscription.go
orenl-lava: Done

x/subscription/keeper/subscription.go
orenl-lava: It gets a partially filled Subscription object - with 4 fields: creator, consumer, plan-index, is-yearly; And it fills the rest. Will change it to accept the arguments explicitly.

x/subscription/keeper/subscription_test.go
orenl-lava: Sure.

x/subscription/types/keys.go
orenl-lava: copied from x/pairing/.., x/spec/..., x/conflict/...

x/subscription/types/keys.go
orenl-lava: Other modules use this too - see x/spec/key_spec.go for example.

x/subscription/keeper/epoch_start.go
orenl-lava: Deferred to a separate issue.

proto/subscription/subscription.proto
```diff
@@ -0,0 +1,17 @@
+syntax = "proto3";
+package lavanet.lava.subscription;
+
+option go_package = "github.com/lavanet/lava/x/subscription/types";
+
+message Subscription {
```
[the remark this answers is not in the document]
orenl-lava: done

Yaroms (approved):

PR #339, opened 2023-03-04
CNS-311 fixation store prefix bug

orenl-lava:
1. Sanitize indices stored FixationStore (fix bug with entries that are prefix of other entries)
2. Drop block arg from GetEntry() 
3. FindEntry() to skip stale entres
4. Test whether stale based on time since refcount becomes zero
5. Some code clean-up


common/fixation_entry.go
```diff
@@ -61,18 +70,49 @@ type FixationStore struct {
 	prefix   string
 }
 
+// sanitizeIdnex checks that a string contains only visible ascii characters
+// (i.e. Ascii 32-126), and appends a (ascii) DEL to the index; this ensures
+// that an index can never be a prefix of another index.
+func sanitizeIndex(index string) (string, error) {
+	for i := 0; i < len(index); i++ {
+		if index[i] < 32 || index[i] > 126 {
```
Yaroms, line 78: lets make those numbers constants

common/fixation_entry.go
```diff
@@ -61,18 +70,49 @@ type FixationStore struct {
...
+			return index, types.ErrInvalidIndex
+		}
+	}
+	// '\177' is octal for 127, which is Ascii DEL
+	return index + string('\177'), nil
```
Yaroms, line 83: also this

Yaroms (changes requested):
very clean code

common/fixation_entry.go
orenl-lava: done

common/fixation_entry.go
orenl-lava: done

common/fixation_entry.go
```diff
@@ -230,13 +241,22 @@ func (fs *FixationStore) getUnmarshaledEntryForBlock(ctx sdk.Context, index stri
...
 
+	latest := true
+
 	for ; iterator.Valid(); iterator.Next() {
 		var entry types.Entry
 		fs.cdc.MustUnmarshal(iterator.Value(), &entry)
 
 		if entry.GetBlock() <= block {
+			// TODO: remove exception for "latest" when latest gets refcount
+			if !latest && entry.IsStale(ctx) {
```
Yaroms, line 252: just return from the function, "latest " is not needed

Yaroms (approved):

common/fixation_entry.go
orenl-lava: Indeed `latest` is not needed - done.
The `break` is useful to ensure one path for "not found".

Yaroms (approved):

common/fixation_entry_index.go
```diff
@@ -6,49 +6,42 @@ import (
...
+// efficiently retrieve a list of all Entry indices that are stored. (Without
+// it, we would need to iterate through all the Entries *including* all their
+// versions - which may be wasteful).
+//
+// EntryIndex objects are stored with a distinct prefix to avoid confusion
+// when iterating over Entries. See example in `x/common/fixation_entry.go`.
+
+// setEntryIndex stores an Entry index in the store
+func (fs FixationStore) setEntryIndex(ctx sdk.Context, safeIndex string) {
+	store := prefix.NewStore(ctx.KVStore(fs.storeKey), types.KeyPrefix(fs.createEntryIndexStoreKey()))
```
omerlavanet, line 23: we can add another store utility function here for EntryIndex store and remove repeated calls to store create also here like you did in the prev file

common/fixation_entry_index.go
```diff
@@ -6,49 +6,42 @@ import (
...
 
-// RemoveEntryIndex removes an EntryIndex from the store
-func (fs FixationStore) removeEntryIndex(ctx sdk.Context, index string) {
-	// get the index store with the fixation key
-	store := prefix.NewStore(ctx.KVStore(fs.GetStoreKey()), types.KeyPrefix(fs.createEntryIndexStoreKey()))
-
-	// remove the index from the store
-	store.Delete(types.KeyPrefix(fs.createEntryIndexKey(index)))
+// removeEntryIndex removes an Entry index from the store
+func (fs FixationStore) removeEntryIndex(ctx sdk.Context, safeIndex string) {
```
omerlavanet, line 29: how about everyTime we assume safe index we check the last ascii marker is indeed set? will make it easier to find mistakes, its as easy as adding a utility function and checking the last byte (doesn't affect perf)

common/types/fixationEntry.go
```diff
@@ -0,0 +1,16 @@
...
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+// IsEntryStale tests whether an entry is stale, i.e. has refcount zero _and_
+// has passed its stale_at time (more than STALE_ENTRY_TIME since deletion).
+func (entry Entry) IsStale(ctx sdk.Context) bool {
+	if entry.GetRefcount() == 0 {
+		if entry.StaleAt < uint64(ctx.BlockHeight()) {
```
omerlavanet, line 11: no need for nesting, can use &&

omerlavanet (changes requested):

common/types/fixationEntry.go
orenl-lava: Yes. This form intended for the reader - emphasize the two tests.

common/fixation_entry.go
```diff
@@ -61,48 +71,106 @@ type FixationStore struct {
...
 		// if the new entry's block is equal to the latest entry, overwrite the latest entry
 		if block == latestEntry.GetBlock() {
 			return fs.ModifyEntry(ctx, index, block, entryData)
 		}
+
+		// if the old latest entry has refcount of 0, then update its "stale_at" time
+		// TODO: remove this when the latest entry gets its own refcount.
+		if latestEntry.Refcount == 0 {
+			// never overflows because because ctx.BlockHeight is int64
+			latestEntry.StaleAt = uint64(ctx.BlockHeight()) + uint64(types.STALE_ENTRY_TIME)
```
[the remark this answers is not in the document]
orenl-lava: This check is for the entry that was "latest entry" just prior to appending the current one. So StaleAt could not have been set for this entry - it was in use by definition, due to being the latest.
If nobody else took reference to it (e.g. either no-one did GetPlan() to got the latest plan, or someone did but later also called PutPlan() on it) - then it is unneeded from exactly now, so we set the StaleAt.
If there is a reference to it, because it is in use by someone, then the ref-count won't be zero and we won't set StaleAt at all.

omerlavanet (approved):

PR #343, opened 2023-03-07
CNS-175 continued: improve subscription tests and CLI

orenl-lava:
1. Fix bug in subscription nextMonth() on Jan 30,31 (spilled over to March)
2. Add unit tests for calculation of ExpiryTime (regular and yearly plans)
3. Add unit tests for calculation of plan cost (yearly with/without discount)
4. Improve CLI for (tx) subscribe and (query) current-subscription
    make the "consumer" and "is_yearly" args options
    (defaults are "creator" and "false" respectively).


orenl-lava:
Supersedes by #351

PR #348, opened 2023-03-09
Replace unused specHash with BlockLastUpdated

orenl-lava:
Use Spec.BlockLastUpdated (see commit a476ab00814a2dee9a1fd521e16720e775ea7b88 "added trait of block that spec was updated") to determine the spec version instead of calculating hash.

Also improve logic for server to wait until the desired spec is obtained, and respect the Context while doing so.


Yaroms:
this code is deprecated

PR #351, opened 2023-03-09
CNS-334: Subscription fixes, tests, and CLI

orenl-lava:
1. Fix bug in subscription nextMonth() on Jan 30,31 (spilled over to March)
2. Add unit tests for calculation of ExpiryTime (regular and yearly plans)
3. Add unit tests for calculation of plan cost (yearly with/without discount)
4. Replace `is_yearly` with `duration`, all plans are for 1 month (remove plan `Duration`)
5. Rename CLI from `tx subscription subscribe` to `tx subscription buy`
6. Improve CLI for (tx) `buy` and (query) `current-subscription`
7. Make the "consumer" and "duration" args options (defaults are "creator" and "1" respectively).


proto/plans/plan.proto
```diff
@@ -8,16 +8,15 @@ import "cosmos/base/v1beta1/coin.proto";
...
-    uint64 servicers_to_pair = 7; // max providers to be paired in an epoch
-    bool allow_overuse = 8; // allow CU overuse flag
-    uint64 overuse_rate = 9; // price of CU overuse
-    string name = 10; // plan name (non-unique "human" name)
-    string description = 11; // plan description (for humans)
-    string type = 12; // plan type
-    uint64 annual_discount_percentage = 13; // discount for buying the plan for a year
-}
\ No newline at end of file
+    uint64 block = 2; // the epoch that this plan was created
```
Yaroms, line 11: lets not change the prorobuff index, this version is already live and it might cause problems, leave 2 empty (you can mark it as reserved)

x/subscription/client/cli/query_current_subscription.go
```diff
@@ -1,29 +1,28 @@
...
+		Args:  cobra.RangeArgs(0, 1),
 		RunE: func(cmd *cobra.Command, args []string) (err error) {
-			reqConsumer := args[0]
-
 			clientCtx, err := client.GetClientTxContext(cmd)
 			if err != nil {
 				return err
 			}
 
+			reqConsumer := clientCtx.GetFromAddress().String()
```
Yaroms, line 21: query doesnt have "--from" flag

x/subscription/client/cli/tx_buy.go
```diff
@@ -11,31 +9,35 @@ import (
 	"github.com/spf13/cobra"
 )
 
-var _ = strconv.Itoa(0)
-
-func CmdSubscribe() *cobra.Command {
+func CmdBuy() *cobra.Command {
 	cmd := &cobra.Command{
-		Use:   "subscribe [consumer] [index] [is-yearly]",
```
Yaroms, line 18: if params are optional, explain it in the use

Yaroms (changes requested):

proto/plans/plan.proto
orenl-lava: ok

x/subscription/client/cli/query_current_subscription.go
orenl-lava: True; Reverted - so "[consumer]" arg is mandatory again.

x/subscription/client/cli/tx_buy.go
orenl-lava: ok.

Yaroms (approved):

x/subscription/client/cli/tx_buy.go
```diff
@@ -11,31 +9,35 @@ import (
...
-		Short: "Subscribe to a service plan",
-		Args:  cobra.ExactArgs(3),
+		Use:   "buy [index] [optional: consumer] [optional: duration]",
```
omerlavanet, line 14: lets add an example, also since duration is uint64 and not time.Duration you have to specify units (months)

x/subscription/keeper/subscription.go
```diff
@@ -172,47 +188,34 @@ func (k Keeper) CreateSubscription(
...
+
+	if duration < 12 {
+		price.Amount = price.Amount.MulRaw(int64(duration))
+		for i := 0; i < int(duration); i++ {
+			expiry = nextMonth(expiry)
+		}
 	}
 
-	if isYearly && duration < 12 {
+	if duration == 12 {
```
omerlavanet, line 215: can buy more than 12, and discount should apply, also can be an else case for the less than 12 if case prior

x/subscription/keeper/subscription.go
```diff
@@ -172,47 +188,34 @@ func (k Keeper) CreateSubscription(
...
 
-	duration := int(plan.GetDuration())
-	for i := 0; i < duration; i++ {
-		expiry = nextMonth(expiry)
+	// use current block's timestamp for subscription start-time
+	timestamp := ctx.BlockTime()
+	expiry := timestamp
+
+	if duration < 12 {
+		price.Amount = price.Amount.MulRaw(int64(duration))
```
omerlavanet, line 209: let's do the price depending on the months amount, and the multiplication of the price with duration outside the if case both both cases, let's do 12 as a const in the beginning of the file and not repeated in code numbers

x/subscription/types/subscription.go
```diff
@@ -7,6 +7,10 @@ import (
 	sdkerrors "github.com/cosmos/cosmos-sdk/types/errors"
 )
 
+const (
+	MAX_SUBSCRIPTION_DURATION = 12 // max duration of subscription in months
```
omerlavanet, line 11: why 12 is the max? if i want 13 months i can't?

omerlavanet (changes requested):

x/subscription/keeper/subscription.go
orenl-lava: FYI, it was decided to limit duration to 12 months per purchase, and also DurationLeft may not exceed 13 months (max duration + 1 for current month in which we renew before expiry). But I'll adjust to code in case that limit changes.

x/subscription/types/subscription.go
orenl-lava: That's the design decision.
Rationale: this is similar to standard SaaS offering, you can pay per month or per year, but not for years onward in advance (so users won't try to hedge on costs going up in coming years, and providers won't be caught in unreasonable commitment to provide services for very long time at fixed, low cost).

omerlavanet (dismissed):

omerlavanet (approved):

PR #352, opened 2023-03-12
CNS 315: implement subscription renewal

orenl-lava:
Implement subscription renewal:

Renewal takes place when buying a subscription for a consumer with an existing subscription, with the same plan (index and version). In this case, the existing subscription gets extended by the (new) duration requested. The total duration is limited to 1 year - and 1 additional month of grace to allow renewal with yearly discount before an existing subscription ends.

The first commit changes subscription logic to track CUs and remaining duration monthly.
The second commit adds support for subscription renewal.

(PR based on branch #351 for https://github.com/lavanet/lava/tree/CNS-334-subscription-fixes-tests-and-cli)


x/subscription/keeper/epoch_start.go
```diff
@@ -1,22 +1,61 @@
...
+			k.RemoveSubscription(ctx, sub.Consumer)
+			continue
+		}
+
+		sub.DurationLeft -= 1
+
+		if sub.DurationLeft > 0 {
+			date = nextMonth(date)
+			sub.MonthExpiryTime = uint64(date.Unix())
+			sub.MonthCuLeft = sub.MonthCuTotal
```
Yaroms, line 44: i feel like fixation will be a better fit and will solve all of this. wdyt?
how do you know where to add the payment if they are received from the last month?

x/subscription/keeper/subscription.go
```diff
@@ -176,71 +123,100 @@ func (k Keeper) CreateSubscription(
...
+				"err": err.Error(),
+			}
+			return utils.LavaError(ctx, logger, "CreateSubscription", details, "failed to create default project")
+		}
+	} else {
+		// allow renewal with the same plan ("same" means both plan index,block match);
+		// otherwise, only one subscription per consumer
+		if !(plan.Index == sub.PlanIndex && plan.Block == sub.PlanBlock) {
+			details := map[string]string{"consumer": consumer}
+			return utils.LavaError(ctx, logger, "CreateSubscription", details, "consumer has existing subscription")
```
Yaroms, line 177: "consumer has existing subscription with a different plan"

x/subscription/keeper/subscription.go
```diff
@@ -176,71 +123,100 @@ func (k Keeper) CreateSubscription(
...
+		// For now, allow renewal only by the same creator.
+		// TODO: after adding fixation, we can allow different creators
+		if creator != sub.Creator {
+			details := map[string]string{"creator": consumer}
+			return utils.LavaError(ctx, logger, "CreateSubscription", details, "existing subscription has different creator")
+		}
+
+		// The total duration may not exceed MAX_SUBSCRIPTION_DURATION, but allow an
+		// extra month to account for renwewals before the end of current subscription
+		if sub.DurationLeft+duration > types.MAX_SUBSCRIPTION_DURATION+1 {
```
Yaroms, line 189: if sub.DurationLeft+duration >= types.MAX_SUBSCRIPTION_DURATION

proto/subscription/subscription.proto
```diff
@@ -6,11 +6,14 @@ option go_package = "github.com/lavanet/lava/x/subscription/types";
...
-  string plan_index = 4;
-  uint64 plan_block = 5;
-  uint64 duration = 6; // duration in months
-  uint64 expiry_time = 7;
-  uint64 usedCU = 8;
-  uint64 remainingCU = 9;
+  uint64 block = 3; // when the subscription was created
+  string plan_index = 4; // index (name) of plan
+  uint64 plan_block = 5; // when the plan was created
+  uint64 duration_total = 6; // total requested duration in months
```
Yaroms, line 12: why do we need the total? looks like it is only used in tests

Yaroms (changes requested):

x/subscription/types/subscription.go
```diff
@@ -9,10 +9,11 @@ import (
 
 const (
 	MAX_SUBSCRIPTION_DURATION = 12 // max duration of subscription in months
+	PAYMENT_GRACE_TIME_HOURS  = 24 // time to wait for payments in hours
 )
 
-func (sub Subscription) IsExpired(date time.Time) bool {
-	expiry := time.Unix(int64(sub.ExpiryTime), 0).UTC()
+func (sub Subscription) IsMonthExpired(date time.Time) bool {
+	expiry := time.Unix(int64(sub.MonthExpiryTime), 0).UTC()
```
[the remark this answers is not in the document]
Yaroms: thanks for the clearance :)

x/subscription/keeper/epoch_start.go
Yaroms: but how do we know which is which? if a provider gives a relay payment with blockheight X, how do you know to which month to add it?

x/subscription/keeper/subscription.go
Yaroms: i get it, im just saying that this is confusing, why not remove the +1 and use >= same result, more readable

proto/subscription/subscription.proto
Yaroms: so if someone will buy 4 month you will auto buy for him another 4 month? seems odd. auto-renewel should add just 1 month i think, but it is the products decision

x/subscription/keeper/epoch_start.go
orenl-lava: If the payment is for this current month, then we use MonthCuLeft.
If the payment is for previous month, then we use PrevCuLeft.
We can tell which case it is because we know PrevExpiryBlock - the barrier between previous and current month.

x/subscription/keeper/subscription.go
orenl-lava: They aren't equivalent... 
(DurationLeft=1 with duration=11 (or 12) are both valid, but 1+11 >= 12 (or 13) both evaluate to true)
Besides, the explicit "+1" makes the logic visible: MAX plus ONE

proto/subscription/subscription.proto
orenl-lava: Not me personally :)
But, consider there is mechanism to auto-renew (on chain, or 3rd party service, doesn't matter).
Suppose I bought 1 month subscription - I'd like it to auto-renew by 1 month at a time going forward.
Suppose I bought yearly subscription, - I'd like it to auto-renew by 1 year at a time going forward (or I will lose the discount).

Yaroms (dismissed):

omerlavanet (changes requested):
dup line, add zero check for duration

omerlavanet (approved):

PR #356, opened 2023-03-14
Add safe random number generator to avoid potential security vulnerab…

DedicatedDev:
There is a small issues to use weak rand functions.

### DIFF
--- a/protocol/lavasession/consumer_types.go
+++ b/protocol/lavasession/consumer_types.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"math"
-	"math/rand"
 	"sort"
 	"strconv"
 	"sync/atomic"
@@ -213,10 +212,7 @@ func (cswp *ConsumerSessionsWithProvider) getConsumerSessionInstanceFromEndpoint
 		return nil, 0, MaximumNumberOfSessionsExceededError
 	}
 
-	randomSessionId := int64(0)
-	for randomSessionId == 0 { // we don't allow 0
-		randomSessionId = rand.Int63()
-	}
+	randomSessionId := utils.SafeRand()
 
 	consumerSession := &SingleConsumerSession{
 		SessionId: randomSessionId,
--- a/utils/safernd.go
+++ b/utils/safernd.go
@@ -1,0 +1,23 @@
+package utils
+
+import (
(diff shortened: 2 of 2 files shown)


protocol/lavasession/consumer_types.go
```diff
@@ -213,10 +212,7 @@ func (cswp *ConsumerSessionsWithProvider) getConsumerSessionInstanceFromEndpoint
 		return nil, 0, MaximumNumberOfSessionsExceededError
 	}
 
-	randomSessionId := int64(0)
-	for randomSessionId == 0 { // we don't allow 0
-		randomSessionId = rand.Int63()
-	}
+	randomSessionId := utils.SafeRand()
```
omerlavanet, line 215: we are using rand here to make sessions statistically unique between unsynced parties, and not random, the generating party (consumer) has an interest for them not to collide to save up on probing requests to the provider
there is no damage if it is predictable or craftable by either party, this do however need to be as performant as we can, so let's continue with whatever generator is faster

utils/safernd.go
```diff
@@ -0,0 +1,23 @@
...
+)
+
+func SafeRand() int64 {
+	var safeRndVal int64
+	err := binary.Read(rand.Reader, binary.LittleEndian, &safeRndVal)
+	if err != nil {
+		LavaFormatInfo(err.Error(), nil)
+	}
+	safeRndVal &= 0x7fffffffffffffff // set the highest bit to 0 to get a positive number
+	for safeRndVal == 0 {            // we don't allow 0
```
omerlavanet, line 15: let's remove the denying of 0 to outside this function since other places that will need a stronger rand from crypto/rand might actually allow it

utils/safernd.go
```diff
@@ -0,0 +1,23 @@
...
+		err = binary.Read(rand.Reader, binary.LittleEndian, &safeRndVal)
+		if err != nil {
+			LavaFormatInfo(err.Error(), nil)
+		}
+		safeRndVal &= 0x7fffffffffffffff // set the highest bit to 0 to get a positive number
```
omerlavanet, line 20: let's use a go const instead of manually writing the number ourselves

omerlavanet (changes requested):

PR #372, opened 2023-03-23
CNS-348: do not jail providers with too short history

orenl-lava:
Providers just recently staked, or just recently unfrozen, do not have enough history to properly judge their performance and QoS. Therefore, exclude them from consideration of unresponsive providers.

### DIFF
--- a/x/pairing/keeper/unresponsive_provider.go
+++ b/x/pairing/keeper/unresponsive_provider.go
@@ -29,14 +29,14 @@ func (k Keeper) UnstakeUnresponsiveProviders(ctx sdk.Context, epochsNumToCheckCU
 		largerEpochsNumConst = epochsNumToCheckCUForUnresponsiveProvider
 	}
 
-	// To check for punishment, we have to go back recommendedEpochNumToCollectPayment+max(epochsNumToCheckCUForComplainers,epochsNumToCheckCUForUnresponsiveProvider) epochs from the current epoch. If there isn't enough memory, do nothing
-	epochTemp := currentEpoch
-	for counter := uint64(0); counter < largerEpochsNumConst+recommendedEpochNumToCollectPayment; counter++ {
-		previousEpoch, err := k.epochStorageKeeper.GetPreviousEpochStartForBlock(ctx, epochTemp)
-		if err != nil {
-			return nil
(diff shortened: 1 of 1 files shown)


x/pairing/keeper/unresponsive_provider.go
```diff
@@ -79,6 +80,19 @@ func (k Keeper) UnstakeUnresponsiveProviders(ctx sdk.Context, epochsNumToCheckCU
...
 }
 
+// getBlockEpochsAgo returns the block numEpochs back from the given blockHeight
+func (k Keeper) getBlockEpochsAgo(ctx sdk.Context, blockHeight uint64, numEpochs uint64) uint64 {
+	for counter := 0; counter < int(numEpochs); counter++ {
+		var err error
+		blockHeight, err = k.epochStorageKeeper.GetPreviousEpochStartForBlock(ctx, blockHeight)
+		if err != nil {
+			// too early in the chain life: bail without an error
+			return uint64(0)
```
omerlavanet, line 90: how about returning an error if it doesn't exist instead of 0? otherwise you are counting on the caller function to remember always to check for 0, which is more prone to errors than checking for err (that is enforced by go)

x/pairing/keeper/unresponsive_provider.go
```diff
@@ -47,28 +47,29 @@ func (k Keeper) UnstakeUnresponsiveProviders(ctx sdk.Context, epochsNumToCheckCU
...
+	minPaymentBlock := k.getBlockEpochsAgo(ctx, currentEpoch, recommendedEpochNumToCollectPayment)
+	if minPaymentBlock == 0 {
+		// not enough history, do nothiing
+		return nil
 	}
 
 	// Go over the staked provider entries (on all chains)
 	for _, providerStakeStorage := range providerStakeStorageList {
 		for _, providerStakeEntry := range providerStakeStorage.GetStakeEntries() {
+			if minHistoryBlock < providerStakeEntry.StakeAppliedBlock {
```
omerlavanet, line 59: simple, yet efficient :)

omerlavanet (changes requested):

omerlavanet (approved):

PR #380, opened 2023-03-28
CNS-229: set consumer policies for project

oren-lava:
- [V] Commit message follows the Contribution Guidelines
- [V] Tests ran locally and added/modified if needed
- [V] Docs have been added/updated, if applicable
- [V] If applicable - JIRA ticket ID was added

feature

Currently, there is no way for a user to change the policy of its project

I've added a new TX that lets the user set a new policy to its project. The policy is detailed in a YAML file. An example file is in cookbook/project-policies

Tested with local node and wrote a unit test

* **Other information**:
Currently, there is no validation of the CU limits and max providers to pair fields of the policy. These fields should be restricted by the subscription's plan policy, but this is not implemented yet (design required)


app/app.go
```diff
@@ -473,6 +478,7 @@ func New(
 		app.ProjectsKeeper,
 		app.PlansKeeper,
 	)
+
```
orenl-lava, line 481: nit: unneeded newline

oren-lava: done

PR #383, opened 2023-03-29
GROW-687 - Fix newrelic reporting

nirtayeb:


protocol/chainlib/jsonRPC.go
```diff
@@ -291,7 +292,8 @@ func (apil *JsonRPCChainListener) Serve(ctx context.Context) {
 	app.Get("/:dappId/websocket", websocketCallbackWithDappID) // catching http://HOST:PORT/1/websocket requests.
 
 	app.Post("/:dappId/*", func(fiberCtx *fiber.Ctx) error {
-		apil.logger.LogStartTransaction("jsonRpc-http post")
+		endTx := apil.logger.LogStartTransaction("jsonRpc-http post")
+		defer endTx()
```
omerlavanet, line 296: nil check so we won;'t crash if the handler is dead

omerlavanet (changes requested):

ranlavanet (approved):

protocol/chainlib/jsonRPC.go
```diff
@@ -208,7 +208,8 @@ func (apil *JsonRPCChainListener) Serve(ctx context.Context) {
 	app.Use(favicon.New())
 
 	app.Use("/ws/:dappId", func(c *fiber.Ctx) error {
-		apil.logger.LogStartTransaction("jsonRpc-WebSocket")
+		endTx := apil.logger.LogStartTransaction("jsonRpc-WebSocket")
```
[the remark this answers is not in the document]
nirtayeb: i removed the reporting for websocket

protocol/chainlib/jsonRPC.go
nirtayeb: add nil check inside the returned function from LogStartTransaction

omerlavanet (approved):

PR #395, opened 2023-04-03
CNS-352: Reset CU at month end

orenl-lava:
This PR ensures that subscriptions at the end of each month reset/recharge the CU usage of their projects. For this, we store (in the Projects module) the list of projects for each subscription.

The PR also fixes calls to DeleteProject() which used the bare project name, rather than the outcome of ProjectIndex() function (which combines the subscription address with the project name).

NOTE: this requires a store-migration to ensure that existing subscriptions get their projects tracked in the Projects module store.


x/pairing/types/expected_keepers.go
```diff
@@ -67,6 +67,6 @@ type BankKeeper interface {
 
 type ProjectsKeeper interface {
 	GetProjectDevelopersPolicy(ctx sdk.Context, developerKey string, blockHeight uint64) (policy projectstypes.Policy, err error)
-	AddComputeUnitsToProject(ctx sdk.Context, developerKey string, blockHeight uint64, cu uint64) (err error)
```
Yaroms, line 70: orenS is in the middle of using/changing this function and this can lead to a problem when he does his PR, lets keep the renaming to another PR or let him do it

x/projects/keeper/creation.go
```diff
@@ -38,6 +38,8 @@ func (k Keeper) CreateProject(ctx sdk.Context, subscriptionAddress string, proje
 	}
 
 	project.Enabled = enable
+
+	k.AppendSubscriptionProject(ctx, subscriptionAddress, project.Index)
```
Yaroms, line 42: we dont need this

x/projects/keeper/project.go
```diff
@@ -4,11 +4,55 @@ import (
...
 
+	"golang.org/x/exp/slices"
+
+	"github.com/cosmos/cosmos-sdk/store/prefix"
 	sdk "github.com/cosmos/cosmos-sdk/types"
 	"github.com/lavanet/lava/utils"
 	"github.com/lavanet/lava/x/projects/types"
 )
 
+func (k Keeper) GetSubscriptionProjects(ctx sdk.Context, subscriptionAddr string) []string {
```
Yaroms, line 15: we dont need all this functionality, just iterate on all the fixations with the subscription prefix, this is redundent

x/projects/types/keys.go
```diff
@@ -19,10 +19,24 @@ const (
 	// prefix for the projects fixation store
 	ProjectsFixationPrefix = "prj-fs"
 
-	// prefix for the projects fixation store
+	// prefix for the developer keys fixation store
 	DeveloperKeysFixationPrefix = "dev-fs"
+
+	// prefix for the subscription-to-projects mapping
+	SubscriptionProjectsPrefix = "sub/projs/"
```
Yaroms, line 26: same

Yaroms (changes requested):

proto/projects/project.proto
```diff
@@ -43,4 +43,9 @@ message ChainPolicy {
 message ProtoDeveloperData {
     string projectID = 1;
     string vrfpk = 2;
-}
\ No newline at end of file
+}
+
+// map subscription to a list of its projects
+message Subscription {
```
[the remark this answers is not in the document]
orenl-lava: my concern was whether it could happen that a (subscription) key is a prefix of another (subscription) key. I don't know if that's possible -and if so, probably very hard to create intentionally- but if it is then I believe we must use an extra map.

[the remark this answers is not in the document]
orenl-lava: actually, an alternative is to force a delimiter after the subscription key (in the current scheme), and then will eliminate the risk of unintended prefix issue. (Similar to how we did so in FixationStore). That's indeed better than adding the extra map.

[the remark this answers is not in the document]
orenl-lava: From bech32 spec at https://en.bitcoin.it/wiki/BIP_0173#Specification: it as two parts, the latter may contain alphanumeric characters only. Projects currently use "-" (to join the subscription address to the project name), which suffices to prevent prefix collisions with addresses.

x/subscription/keeper/subscription_test.go
```diff
@@ -317,6 +320,49 @@ func TestSubscriptionAdminProject(t *testing.T) {
 	require.Nil(t, err)
 }
 
+func TestMonthlyRechargeCU(t *testing.T) {
```
Yaroms, line 323: in the tests we want to use the message server (ts.servers) to simulate the whole process and not call the keepers functions

x/subscription/keeper/subscription_test.go
```diff
@@ -317,6 +320,49 @@ func TestSubscriptionAdminProject(t *testing.T) {
...
+	require.Equal(t, sub.PrevCuLeft, sub.MonthCuTotal-1000)
+	err = projectKeeper.AddComputeUnitsToProject(ts.ctx, creator, block, 1000)
+	require.Nil(t, err)
+	// verify that project used the CU
+	proj, _, err := projectKeeper.GetProjectForDeveloper(ts.ctx, creator, block)
+	require.Nil(t, err)
+	require.Equal(t, uint64(1000), proj.UsedCu)
+
+	// fast-forward one months
+	sub = ts.expireSubscription(sub)
```
Yaroms, line 350: also lets add a fixation in the middle of the month for the project and check that the old ones are ok

Yaroms (changes requested):

x/subscription/keeper/subscription_test.go
orenl-lava: Unsure what you propose.
Monthly recharge is an internal activity that is carried out by the code periodically and not as a result of a message to a server. Specifically, the subscription code checks for subscription whose current month just expires, and if so runs the needed logic.

x/subscription/keeper/subscription_test.go
orenl-lava: done

Yaroms (approved):

PR #398, opened 2023-04-04
CNS-376: projects module upgrade: v2 to v3

orenl-lava:
The projects module now tracks the list of projects per subscription. 
This migration scans the existing projects in the store, and generates the data to tracks projects per subscription.


orenl-lava:
The projects module logic now implemented using existing store format- migration no longer needed.

PR #400, opened 2023-04-05
 PRT-167-Data-reliability-for-rest-tendermint-grpc-optional-params

candostyavuz:
The requested block has been added to the `newChainMessage`, replacing the prior usage of `spectypes.NOT_APPLICABLE` for both REST and gRPC interfaces.

### DIFF
--- a/protocol/chainlib/chainproxy/rpcInterfaceMessages/restMessage_test.go
+++ b/protocol/chainlib/chainproxy/rpcInterfaceMessages/restMessage_test.go
@@ -2,6 +2,8 @@ package rpcInterfaceMessages
 
 import (
 	"testing"
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestRestMessage(t *testing.T) {
@@ -14,9 +16,7 @@ func TestRestMessage(t *testing.T) {
 
 	// Test GetParams method
 	params := restMessage.GetParams()
-	if params == nil {
-		t.Errorf("Expected params, but got nil")
-	}
+	require.Nil(t, params)
 
 	// Test GetResult method
 	result := restMessage.GetResult()
--- a/protocol/chainlib/chainproxy/rpcInterfaceMessages/restMessage_test.go
+++ b/protocol/chainlib/chainproxy/rpcInterfaceMessages/restMessage_test.go
@@ -9,9 +9,8 @@ import (
 func TestRestMessage(t *testing.T) {
 	// Test GetParams method
 	restMessage := RestMessage{
-		Path:     "eth_getTransactionByHash",
-		Msg:      []byte{1, 2, 3, 4, 5},
-		SpecPath: "eth_getTransactionByHash",
+		Path:     "blocks/latest",
+		SpecPath: "blocks/latest",
 	}
 
 	// Test GetParams method
--- a/protocol/chainlib/chainproxy/rpcInterfaceMessages/restMessage_test.go
+++ b/protocol/chainlib/chainproxy/rpcInterfaceMessages/restMessage_test.go
@@ -7,14 +7,15 @@ import (
 func TestRestMessage(t *testing.T) {
 	// Test GetParams method
 	restMessage := RestMessage{
-		Path: "eth_getTransactionByHash",
-		Msg:  []byte{1, 2, 3, 4, 5},
+		Path:     "eth_getTransactionByHash",
+		Msg:      []byte{1, 2, 3, 4, 5},
+		SpecPath: "eth_getTransactionByHash",
 	}
 
 	// Test GetParams method
 	params := restMessage.GetParams()
-	if params != nil {
-		t.Errorf("Expected nil, but got %v", params)
+	if params == nil {
+		t.Errorf("Expected params, but got nil")
 	}
 
 	// Test GetResult method
--- a/protocol/chainlib/common_test.go
+++ b/protocol/chainlib/common_test.go
@@ -25,7 +25,7 @@ func TestMatchSpecApiByName(t *testing.T) {
 		expectedOk  bool
 	}{
 		{
-			name: "test0",
+			name: "test1",
 			serverApis: map[string]spectypes.ServiceApi{
 				"/blocks/[^\\/\\s]+": {
 					Name: "/blocks/{height}",
@@ -56,7 +56,7 @@ func TestMatchSpecApiByName(t *testing.T) {
 			expectedOk:  true,
 		},
 		{
-			name: "test1",
+			name: "test2",
 			serverApis: map[string]spectypes.ServiceApi{
 				"/cosmos/base/tendermint/v1beta1/blocks/[^\\/\\s]+": {
 					Name: "/cosmos/base/tendermint/v1beta1/blocks/{height}",
@@ -87,7 +87,7 @@ func TestMatchSpecApiByName(t *testing.T) {
 			expectedOk:  true,
 		},
 		{
-			name: "test2",
+			name: "test3",
 			serverApis: map[string]spectypes.ServiceApi{
 				"/cosmos/base/tendermint/v1beta1/blocks/latest": {
 					Name: "/cosmos/base/tendermint/v1beta1/blocks/latest",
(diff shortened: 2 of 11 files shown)


protocol/chainlib/chainproxy/rpcInterfaceMessages/restMessage_test.go
```diff
@@ -7,14 +7,15 @@ import (
...
+		Path:     "eth_getTransactionByHash",
+		Msg:      []byte{1, 2, 3, 4, 5},
+		SpecPath: "eth_getTransactionByHash",
 	}
 
 	// Test GetParams method
 	params := restMessage.GetParams()
-	if params != nil {
-		t.Errorf("Expected nil, but got %v", params)
+	if params == nil {
```
ranlavanet, line 17: you dont need to do this check you can do 
require.Nil(t, params)

ranlavanet: no need for print as well the require is enough.

protocol/chainlib/common_test.go
```diff
@@ -25,35 +25,128 @@ func TestMatchSpecApiByName(t *testing.T) {
...
+				},
 			},
-			inputName:   "test3-match",
-			expectedApi: spectypes.ServiceApi{},
-			expectedOk:  false,
+			inputName:   "/cosmos/base/tendermint/v1beta1/blocks/latest",
+			expectedApi: spectypes.ServiceApi{Name: "/cosmos/base/tendermint/v1beta1/blocks/latest"},
+			expectedOk:  true,
 		},
+		// {
```
ranlavanet, line 120: lets not leave commented code in PR

protocol/chainlib/grpc.go
```diff
@@ -69,19 +70,33 @@ func (apip *GrpcChainParser) ParseMsg(url string, data []byte, connectionType st
...
+	// Extract default block parser
+	blockParser := serviceApi.BlockParsing
+
 	apiInterface := GetApiInterfaceFromServiceApi(serviceApi, connectionType)
 	if apiInterface == nil {
 		return nil, fmt.Errorf("could not find the interface %s in the service %s", connectionType, serviceApi.Name)
 	}
 
+	// Check if custom block parser exists in the api interface
+	// Use custom block parser only for URI calls
```
ranlavanet, line 82: we don't have URI in grpc, no need to overwrite block parsing

protocol/chainlib/rest.go
```diff
@@ -63,11 +64,20 @@ func (apip *RestChainParser) ParseMsg(url string, data []byte, connectionType st
...
+	blockParser := serviceApi.BlockParsing
+
 	apiInterface := GetApiInterfaceFromServiceApi(serviceApi, connectionType)
 	if apiInterface == nil {
 		return nil, fmt.Errorf("could not find the interface %s in the service %s", connectionType, serviceApi.Name)
 	}
 
+	// Check if custom block parser exists in the api interface
+	// Use custom block parser only for URI calls
+	if apiInterface.GetOverwriteBlockParsing() != nil {
```
ranlavanet, line 77: Here as well no need to overwrite block parsing. this is only for tendermint. URI calls

protocol/chainlib/rest_test.go
```diff
@@ -111,8 +111,9 @@ func TestRestParseMessage(t *testing.T) {
 	assert.Equal(t, msg.GetServiceApi().Name, apip.serverApis["API1"].Name)
 
 	restMessage := rpcInterfaceMessages.RestMessage{
-		Msg:  []byte("test message"),
-		Path: "API1",
+		Msg:      []byte("test message"),
+		Path:     "API1",
+		SpecPath: "API1",
```
ranlavanet, line 116: if Spec path is not used in this test you don't need to specify it, it will just be empty by default.

protocol/parser/parser.go
```diff
@@ -307,7 +307,7 @@ func ParseDictionary(rpcInput RPCInput, input []string, dataSource int) ([]inter
 	case map[string]interface{}:
 		// If attribute with key propName exists return value
 		if val, ok := unmarshalledDataTyped[propName]; ok {
-			return appendInterfaceToInterfaceArray(val), nil
+			return appendInterfaceToInterfaceArray(fmt.Sprintf("%v", val)), nil
```
ranlavanet, line 310: this is not the right way to do it, if grpc has objects unmarshalled not as strings which is why I assume you need it changed you can have a method to check the type and convert only the necessary and known types
```
convertInterfaceToString(input interface{}) (string, error) {
switch inp := input .(type) {
	case string:
		return info
	case int:
	       return strconv.formatInt(info,10)
	default:
	       return utils.LavaFormatError("Failed to convert to string, unknown type", info: input , inp  etc..)
}
```

ranlavanet: that way we know which types we support and which we do not

protocol/rpcprovider/rewardserver/reward_server.go
```diff
@@ -70,21 +70,22 @@ type RewardsTxSender interface {
 	EarliestBlockInMemory(ctx context.Context) (uint64, error)
 }
 
-func (rws *RewardServer) SendNewProof(ctx context.Context, proof *pairingtypes.RelaySession, epoch uint64, consumerAddr string) (existingCU uint64, updatedWithProof bool) {
+func (rws *RewardServer) SendNewProof(ctx context.Context, proof *pairingtypes.RelaySession, epoch uint64, consumerAddr string, apiInterface string) (existingCU uint64, updatedWithProof bool) {
 	rws.lock.Lock() // assuming 99% of the time we will need to write the new entry so there's no use in doing the read lock first to check stuff
 	defer rws.lock.Unlock()
+	consumerRewardsKey := proof.SpecId + apiInterface + consumerAddr
```
ranlavanet, line 76: this should be a function, that gets these as inputs and returns them unified that way no mistakes can happen in the order.

protocol/rpcprovider/rewardserver/reward_server.go
```diff
@@ -100,19 +101,20 @@ func (rws *RewardServer) SendNewProof(ctx context.Context, proof *pairingtypes.R
 	return 0, true
 }
 
-func (rws *RewardServer) SendNewDataReliabilityProof(ctx context.Context, dataReliability *pairingtypes.VRFData, epoch uint64, consumerAddr string) (updatedWithProof bool) {
+func (rws *RewardServer) SendNewDataReliabilityProof(ctx context.Context, dataReliability *pairingtypes.VRFData, epoch uint64, consumerAddr string, specId string, apiInterface string) (updatedWithProof bool) {
 	rws.lock.Lock() // assuming 99% of the time we will need to write the new entry so there's no use in doing the read lock first to check stuff
 	defer rws.lock.Unlock()
+	consumerRewardsKey := specId + apiInterface + consumerAddr
```
ranlavanet, line 107: as well here

ranlavanet (changes requested):

protocol/chainlib/chainproxy/rpcInterfaceMessages/restMessage_test.go
candostyavuz: fixed in : fd83ad126719d12e3ec38756fd82b04c61f4af1d

protocol/chainlib/common_test.go
candostyavuz: fixed: fd83ad126719d12e3ec38756fd82b04c61f4af1d

protocol/chainlib/grpc.go
candostyavuz: fixed: 524cb1833ba85f59885c6f6e829b7f9d6d08e451

protocol/chainlib/rest.go
candostyavuz: fixed: 524cb1833ba85f59885c6f6e829b7f9d6d08e451

protocol/parser/parser.go
candostyavuz: Yes the reason was grpc being unmarshalled not as strings. Fixed here: c261e1ecef87053813dee7dc1761a4ff6fe5b86c

protocol/rpcprovider/rewardserver/reward_server.go
candostyavuz: fixed: 57dff5307fc3945fee74060280ef65ceaa19e079

protocol/rpcprovider/rewardserver/reward_server.go
candostyavuz: fixed: https://github.com/lavanet/lava/commit/57dff5307fc3945fee74060280ef65ceaa19e079

protocol/parser/parser.go
```diff
@@ -408,3 +424,14 @@ func appendInterfaceToInterfaceArray(value interface{}) []interface{} {
 	retArr = append(retArr, value)
 	return retArr
 }
+
+func convertInterfaceToString(input interface{}) (string, error) {
+	switch inp := input.(type) {
+	case string:
+		return inp, nil
+	case float64:
+		return fmt.Sprintf("%v", inp), nil
```
ranlavanet, line 433: there is a better way to convert float64 to string use strconv.FormatFloat() instead.

ranlavanet (changes requested):

protocol/chainlib/rest_test.go
candostyavuz: It is actually used in the unit tests because `ParseMsg()` assigns `SpechPath` to `msg` parameter, so it needs to be specified here as well.

protocol/parser/parser.go
candostyavuz: Sure: 41f80bad125e2f5f36fc678183fe4c806426151e

protocol/chainlib/chainproxy/rpcInterfaceMessages/restMessage_test.go
```diff
@@ -2,20 +2,21 @@ package rpcInterfaceMessages
...
+
+	"github.com/stretchr/testify/require"
 )
 
 func TestRestMessage(t *testing.T) {
 	// Test GetParams method
 	restMessage := RestMessage{
-		Path: "eth_getTransactionByHash",
-		Msg:  []byte{1, 2, 3, 4, 5},
+		Path:     "eth_getTransactionByHash",
```
ranlavanet, line 12: eth has no rest messages so it might be confusing ethereum uses only jsonrpc

ranlavanet (changes requested):

protocol/chainlib/chainproxy/rpcInterfaceMessages/restMessage_test.go
candostyavuz: Fixed: ae1eabb9451c17943a95cdbe083634adc9dfb073

ranlavanet (approved):

ranlavanet:
waiting for @omerlavanet approval as well before merge

PR #401, opened 2023-04-06
Enable snapshots

freshe4qa:
Enable snapshots for testnet to enable validators to more quickly deploy nodes/run experiments


Yaroms:
thanks for the effort, we just added it in another PR

PR #402, opened 2023-04-07
Binaries to the CDN registry at tag release

luis02lopez:
Add step to additionally upload the binary to the CDN. We need to define what we consider as "latest" is the latests on pre-release or the one stable?

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -7,6 +7,11 @@ on:
 permissions:
   contents: write
 
+env:
+  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+  AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+  AWS_DEFAULT_REGION: 'us-east-1'
+
 jobs:
   release:
     name: 'release'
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -66,6 +71,11 @@ jobs:
           -H "Content-Type: $(file -b --mime-type build/lavad)" \
           --data-binary @build/lavad \
           $(echo '${{ github.event.release.upload_url }}' | sed 's/{?name,label}/?name=lavad-${{ github.event.release.tag_name }}-linux-amd64/g')
+
+          #Upload to binary registry
+          aws s3 cp build/lavad s3://network-upgrades-zarak/"${{ github.event.release.tag_name }}"/linux-amd64/lavad
+          #Latest -- We need to define what is the latest for us
+          # aws s3 cp build/lavad s3://network-upgrades-zarak/latest/linux-amd64/lavad
```
luis02lopez, line 78: We need to define what we consider as "latest" is the latests on pre-release or the one stable?

This automation is triggered on modification of any Release entry, if we consider latest being the higher tag number, I'll be adding that condition.

.github/workflows/release.yml
```diff
@@ -7,6 +7,11 @@ on:
 permissions:
   contents: write
 
+env:
+  AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
```
luis02lopez, line 11: We'll need to add these keys to the project. After the PR has been approved, I'll ask to have admin priviledges on this repo to add the sensitive information into the secrets.

Yaroms: @luis02lopez @praveenperera , any follow up on that?

praveenperera: @luis02lopez you can use

```
          role-to-assume: arn:aws:iam::529724559987:role/github-cdk-ci-ecr-role-repo-lava
```

PR #403, opened 2023-04-08
CNS-368: fixation: improve API for GetEntry,FindEntry,PutEntry

orenl-lava:
Change GetEntry,FindEntry to not return an error, use MustUnmarshal() instead of Unmarshal(). Change PutEntry to not return anything and not fill the entry whose refcount is being dropped (caller must already have it).

Previously they returned an error if the Ummarshal of the entry data failed. However that was bad for two reasons: first, such a failure would indicate a severe logical bug because the very same data must have been marshaled and saved before - hence it must be either corrupted or a new lava version now expects some different format (either way it's a bug). Second, callers do not have a way to rectify such errors, they just treat it as "not found", making it useless.

### DIFF
--- a/common/fixation_entry.go
+++ b/common/fixation_entry.go
@@ -117,7 +117,12 @@ func (fs *FixationStore) setEntry(ctx sdk.Context, entry types.Entry) {
 }
 
 // AppendEntry adds a new entry to the store
-func (fs *FixationStore) AppendEntry(ctx sdk.Context, index string, block uint64, entryData codec.ProtoMarshaler) error {
+func (fs *FixationStore) AppendEntry(
+	ctx sdk.Context,
+	index string,
+	block uint64,
+	entryData codec.ProtoMarshaler,
+) error {
 	safeIndex, err := sanitizeIndex(index)
 	if err != nil {
 		details := map[string]string{"index": index}
@@ -133,8 +138,8 @@ func (fs *FixationStore) AppendEntry(ctx sdk.Context, index string, block uint64
 		// make sure the new entry's block is not smaller than the latest entry's block
 		if block < latestEntry.GetBlock() {
 			details := map[string]string{
-				"latestBlock": strconv.FormatUint(latestEntry.GetBlock(), 10),
-				"block":       strconv.FormatUint(block, 10),
+				"latestBlock": strconv.Itoa(int(latestEntry.GetBlock())),
+				"block":       strconv.Itoa(int(block)),
 				"index":       index,
 				"fs.prefix":   fs.prefix,
 			}
@@ -217,7 +222,7 @@ func (fs *FixationStore) ModifyEntry(ctx sdk.Context, index string, block uint64
 		details := map[string]string{
 			"fs.prefix": fs.prefix,
 			"index":     index,
-			"block":     strconv.FormatUint(block, 10),
+			"block":     strconv.Itoa(int(block)),
 		}
 		return utils.LavaError(ctx, ctx.Logger(), "SetEntry_cant_find_entry", details, "entry does not exist")
 	}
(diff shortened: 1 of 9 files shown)


common/fixation_entry.go
```diff
@@ -262,82 +267,61 @@ func (fs *FixationStore) getUnmarshaledEntryForBlock(ctx sdk.Context, safeIndex
...
 	fs.setEntry(ctx, entry)
-
-	return nil, true
+	return true
 }
 
-// get entry with index and block with ref decrease
-func (fs *FixationStore) PutEntry(ctx sdk.Context, index string, block uint64, entryData codec.ProtoMarshaler) (error, bool) {
+// PutEntry finds the entry by index and block and decrements the refcount
+func (fs *FixationStore) PutEntry(ctx sdk.Context, index string, block uint64) {
```
oren-lava, line 312: Not sure I follow on why the user must have the entry before using PutEntry. Isn't there a scenario where the user would want to retrieve the required entry and decrease the refCount automatically?
Also, if we keep this change (that PutEntry doesn't return anything), I think the function should be renamed, because PutEntry is not very indicative on what it's doing.

oren-lava (dismissed):

common/fixation_entry.go
orenl-lava: "Put" must always match a previous "Get", therefore:
- it is reasonable to require that the holder (of the reference) knows which entry version to put. 
- it needs only the entry version, not the actual entry - because it is assumed the entry is there (held by holder).
- it must not fail because the entry must have been "Get"ed before (entry not found is a bad bug).
The name "Put" is a common convention as the opposite of "Get": we "Get()" a reference and then we "Put()" it back when done.

common/fixation_entry.go
```diff
@@ -133,8 +138,8 @@ func (fs *FixationStore) AppendEntry(ctx sdk.Context, index string, block uint64
 		// make sure the new entry's block is not smaller than the latest entry's block
 		if block < latestEntry.GetBlock() {
 			details := map[string]string{
-				"latestBlock": strconv.FormatUint(latestEntry.GetBlock(), 10),
-				"block":       strconv.FormatUint(block, 10),
+				"latestBlock": strconv.Itoa(int(latestEntry.GetBlock())),
```
Yaroms, line 141: why is it better? we are casting uint to int just to use ITOA, is the convert more safe?

common/fixation_entry.go
```diff
@@ -262,82 +267,61 @@ func (fs *FixationStore) getUnmarshaledEntryForBlock(ctx sdk.Context, safeIndex
...
 	entry, found := fs.getUnmarshaledEntryForBlock(ctx, safeIndex, block)
 	if !found {
-		return types.ErrEntryNotFound, false
-	}
-
-	// unmarshal the entry's data
-	err = fs.cdc.Unmarshal(entry.GetData(), entryData)
-	if err != nil {
-		return utils.LavaError(ctx, ctx.Logger(), "GetEntry_cant_unmarshal", map[string]string{"err": err.Error()}, "can't unmarshal entry data"), false
+		panic("PutEntry with unknown index: " + index)
```
Yaroms, line 320: lets not panic and return an error

common/fixation_entry.go
```diff
@@ -262,82 +267,61 @@ func (fs *FixationStore) getUnmarshaledEntryForBlock(ctx sdk.Context, safeIndex
...
 	}
 
 	if entry.Refcount == 0 {
-		details := map[string]string{
-			"index":    index,
-			"refcount": strconv.FormatUint(entry.Refcount, 10),
-		}
-		return utils.LavaError(ctx, ctx.Logger(), "PutEntry_zero_count", details, "refcount already reached zero"), false
+		panic("PutEntry with refcount zero index: " + index)
```
Yaroms, line 324: lets not panic and return an error

x/plans/keeper/plan.go
```diff
@@ -16,37 +16,33 @@ func (k Keeper) AddPlan(ctx sdk.Context, planToAdd types.Plan) error {
...
 }
 
-// PutPlan gets a plan from the KVStore. It decreases the plan's refCount by 1
-func (k Keeper) PutPlan(ctx sdk.Context, index string, block uint64) bool {
-	var plan types.Plan
-	_, found := k.plansFs.PutEntry(ctx, index, block, &plan)
-	return found
+// PutPlan finds a plan with nearest-smaller block and decrements its refcount
+func (k Keeper) PutPlan(ctx sdk.Context, index string, block uint64) {
+	k.plansFs.PutEntry(ctx, index, block)
```
Yaroms, line 45: handle error

Yaroms (changes requested):

common/fixation_entry.go
orenl-lava: just less verbose.
I can revert it.

common/fixation_entry.go
orenl-lava: `PutEntry()` is expected to be called by the holder of a reference. Therefore, it is a serious bug if the entry is not found.

common/fixation_entry.go
orenl-lava: Calling `PutEntry()` extraneously is indicative of a serious bug and/or corrupted state. It is not recoverable.

x/plans/keeper/plan.go
orenl-lava: `PutEntry()` cannot fail (though it may panic).
Likewise, `PutPlan()` is expected to be called by the holder of a reference. Therefore, it is a serious bug if the Plan is not found.

Yaroms (approved):

PR #404, opened 2023-04-08
CNS-368: introduce TimerStore to manage future timeouts

orenl-lava:
TimerStore manages timers to efficiently support future timeouts. Timeouts can be based on either block-height or block-timestamp. When a timeout occurs, a designated callback function is called with the details (ctx and data). 

Note: this PR depends on #403.

### DIFF
--- a/common/timer.go
+++ b/common/timer.go
@@ -13,7 +13,7 @@ import (
 // can be based on either block-height or block-timestamp. When a timeout occurs,
 // a designated callback function is called with the details (ctx and data).
 //
-// Once instantiated with NewFixationStore(), it offers the following methods:
+// Once instantiated with NewTimerStore(), it offers the following methods:
 //    - WithCallbackByBlockHeight(callback): sets the callback for block-height timers
 //    - WithCallbackByBlockTime(callback): sets the callback for block-time timers
 //    - AddTimerByBlockHeight(ctx, block, data): add a timer to expire at block height
--- a/common/timer.go
+++ b/common/timer.go
@@ -1,0 +1,206 @@
+package common
+
+import (
+	"math"
+
+	"github.com/cosmos/cosmos-sdk/codec"
+	"github.com/cosmos/cosmos-sdk/store/prefix"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/lavanet/lava/common/types"
+)
+
+// TimerStore manages timers to efficiently support future timeouts. Timeouts
+// can be based on either block-height or block-timestamp. When a timeout occurs,
+// a designated callback function is called with the details (ctx and data).
+//
+// Once instantiated with NewFixationStore(), it offers the following methods:
+//    - WithCallbackByBlockHeight(callback): sets the callback for block-height timers
+//    - WithCallbackByBlockTime(callback): sets the callback for block-time timers
+//    - AddTimerByBlockHeight(ctx, block, data): add a timer to expire at block height
+//    - AddTimerByBlockTime(ctx, timestamp, data): add timer to expire at block timestamp
+//    - Tick(ctx): advance the timer to the ctx's block (height and timestamp)
+//
+// How does it work? The explanation below illustrates how the data is stored, assuming
+// the user is the module "package":
+//
+// 1. When instantiated, TimerStore gets a `prefix string` - used as a namespace to
+// separate between instances of TimerStore. For instance, module "packages" would
+// use its module name for prefix.
+//
+// 2. TimerStore keeps the timers with a special prefix; it uses the timeout value
+// (block height/timestamp) as the key, so that the standard iterator would yield them
+// in the desired chronological order. TimeSStore also keeps the next-timeout values
+// (of block height/timestamp) to efficiently determine if the iterator is needed.
+// For instance, module "packages" may have two (block height) timeouts with their data
+// set to "first" and "second" respectively:
+//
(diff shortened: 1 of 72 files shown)


common/timer.go
```diff
@@ -0,0 +1,206 @@
...
+//
+// 3. TimerStore tracks the next-timeout for both block-height/block-timestamp. On
+// every call to Tick(), it tests the current ctx's block height/timestamp against the
+// respective next-timeout:
+//
+// 4. If the next-timeout is reached/passed, then it will iterate through the timer
+// entries and invoke the (respective) callback for those entries; And finally it will
+// advance the (respective) next-timeout.
+
+// Thus, iterating on the prefix "packages_Entry_Index_" would yield all the package
```
oren-lava, line 50: From here to the example, I'm not sure how this is related to the Timer explanation. Also, you have two "4."

common/timer.go
```diff
@@ -0,0 +1,206 @@
...
+//     prefix: packages_Timer_Next_       key: BlockHeight      value: 150
+//     prefix: packages_Timer_Next_       key: BLockTimer       value: MaxInt64
+//     prefix: packages_Timer_Value_          key: 150          data: "first"
+//     prefix: packages_Entry_Value_          key: 180          data: "second"
+//
+// 3. TimerStore tracks the next-timeout for both block-height/block-timestamp. On
+// every call to Tick(), it tests the current ctx's block height/timestamp against the
+// respective next-timeout:
+//
+// 4. If the next-timeout is reached/passed, then it will iterate through the timer
```
oren-lava, line 46: A thought: isn't it problematic that the callbacks might be invoked long after they were supposed to be called?
For example, say I have a callback that is supposed to be called in block 100 and the user ticks the timer only on block 1000.
Maybe there should be a global timer that is incremented on every new block so there will be a single reference of the current time to all timers

oren-lava (dismissed):

common/timer.go
orenl-lava: The TimerStore is like a library, and users may use it as they see fit.
If exact timeouts are needed, the user can call AdvanceBlock() on every BeginBlock() of its own module.
If it is sufficient to have callbacks on epoch start/end, then the user can call AdvanceBlock() on epoch start/end (and save some CPU not checking blocks in between).
As long as the callback timing is deterministic, and the user controls the "AdvanceBlock()" calls, things are fine.

orenl-lava: Ideally, TimerStore will be its own module, and will have its own BeginBlock(), and won't depend on the user code calling AdvanceBlock(). There can also be a parameter to set the frequency of checking. Currently left for future work.

common/timer.go
orenl-lava: Sigh. Self copyright infringement :)
Will clean up.

common/timer.go
oren-lava: I think your comment about calling AdvanceBlock() on every BeginBlock() of the user's module for exact timeouts should be explicitly documented in the TimerStore documentation as a best/common practice (so a forgetful user, like me, won't forget to do it)

common/types/keys.go
```diff
@@ -6,13 +6,45 @@ func KeyPrefix(p string) []byte {
...
+)
+
+var NextTimerKey = []string{
+	"NextBlock", // for BlockHeight
+	"NextDate",  // for BlockTime
+}
+
+var TimerTypePrefix = []string{
+	"Block_", // for BlockHeight
+	"Date_",  // for BlockTime
```
Yaroms, line 37: lets stick with time instead of date?

Yaroms (changes requested):

omerlavanet (changes requested):
very nice work! high quality code

common/timer.go
```diff
@@ -0,0 +1,203 @@
...
+
+func (tstore *TimerStore) setNextTimeout(ctx sdk.Context, which types.TimerType, value uint64) {
+	store := tstore.getStore(ctx, types.NextTimerPrefix)
+	b := types.EncodeKey(value)
+	store.Set([]byte(types.NextTimerKey[which]), b)
+}
+
+func (tstore *TimerStore) addTimer(ctx sdk.Context, which types.TimerType, value uint64, data string) {
+	store := tstore.getStoreTimer(ctx, which)
+	store.Set(types.EncodeKey(value), []byte(data))
```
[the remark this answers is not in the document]
orenl-lava: See the "How does it work" section (bullet 2) at the beginning of the file.

common/timer_test.go
```diff
@@ -0,0 +1,164 @@
...
+		{op: "addheight", name: "add timer no 3", value: 140, data: "no-3."},
+		{op: "addheight", name: "add timer no 4", value: 150, data: "no-4."},
+		{op: "tickheight", name: "tick before all", value: 110, fire: 0},
+		{op: "tickheight", name: "tick between no-2, no-3", value: 135, fire: 2, data: "no-1.no-2."},
+		{op: "nextheight", name: "next timeout no-3", value: 140},
+		{op: "tickheight", name: "tick after all", value: 155, fire: 2, data: "no-3.no-4."},
+	}
+
+	testWithTimerTemplate(t, playbook, 1)
+}
```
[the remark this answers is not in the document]
orenl-lava: that's guaranteed by cosmos-sdk API KVStorePrefixIterator.
(see https://github.com/cosmos/cosmos-sdk/blob/main/store/types/utils.go)

Yaroms (approved):

PR #405, opened 2023-04-08
CNS-368: fixation store fix refcount and deletions

orenl-lava:
Fix refcount handling: FixationStore entry refcount was half-baked: the latest entries had refcount 0, and the code had special handling to avoid ignoring/removing them.

Fix garbage collection logic for stale entries: (1) When entries get StaleAt field set, set a timer to fire on that block; (2) Whenever such timer fires, scan all versions of the respective index to remove eligible stale versions ("eligible" stale version is one whose preceding version is stale; the oldest version is always "eligible").

### DIFF
--- a/common/fixation_entry.go
+++ b/common/fixation_entry.go
@@ -26,8 +26,9 @@ import (
 //    - GetEntry(index, *entry): get a copy (and reference) of the latest version of an entry
 //    - PutEntry(index, block): drop a reference of a version of an entry
 //    - [TBD] RemoveEntry(index): mark an entry as unavailable for new GetEntry() calls
-//    - GetAllEntryIndex(): get all the entries indices (without versions)
-//    - AdvanceBlock(): notify of block height progress for garbage collection
+//    - GetAllEntryIndices(): get all the entries indices (without versions)
+//    - GetAllEntryVersions(index): get all the versions of an entry (for testing)
+//    - AdvanceBlock(): notify of block progress (e.g. BeginBlock) for garbage collection
 //
 // Entry names (index) must contain only visible ascii characters (ascii values 32-125).
 // The ascii 'DEL' invisible character is used internally to terminate the index values
--- a/common/fixation_entry.go
+++ b/common/fixation_entry.go
@@ -151,13 +151,7 @@ func (fs *FixationStore) AppendEntry(
 			return fs.ModifyEntry(ctx, index, block, entryData)
 		}
 
-		// if the old latest entry has refcount of 0, then update its "stale_at" time
-		// TODO: remove this when the latest entry gets its own refcount.
-		if latestEntry.Refcount == 0 {
-			// never overflows because because ctx.BlockHeight is int64
-			latestEntry.StaleAt = uint64(ctx.BlockHeight()) + uint64(types.STALE_ENTRY_TIME)
-			fs.setEntry(ctx, latestEntry)
-		}
+		fs.putEntry(ctx, latestEntry)
 	}
 
 	// marshal the new entry's data
@@ -169,7 +163,7 @@ func (fs *FixationStore) AppendEntry(
 		Block:    block,
 		StaleAt:  math.MaxUint64,
 		Data:     marshaledEntryData,
-		Refcount: 0,
+		Refcount: 1,
 	}
 
 	fs.setEntry(ctx, entry)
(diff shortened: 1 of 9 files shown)


common/fixation_entry.go
```diff
@@ -169,42 +166,66 @@ func (fs *FixationStore) AppendEntry(
...
 	iterator := sdk.KVStorePrefixIterator(store, []byte{})
 	defer iterator.Close()
 
-	for iterator.Valid() {
-		// umarshal the old entry version
-		var oldEntry types.Entry
-		fs.cdc.MustUnmarshal(iterator.Value(), &oldEntry)
+	// remove "eligible" stale entry versions, where "eligible" is either:
+	//   an entry version that follows a stale entry version, -OR-
+	//   the older entry version
```
oren-lava, line 185: oldest?

common/fixation_entry.go
```diff
@@ -169,42 +166,66 @@ func (fs *FixationStore) AppendEntry(
...
+	// rationale: we keep certain stale entry versions in order to mark -
+	// for future findEntry - that blocks from that stale entry onward are
+	// stale (otherwise, a lookup might resolve successfully with an older
+	// non-stale entry version). For this, one - the oldest - marker is
```
oren-lava, line 189: why would we need the markers to mark which entry is stale? the refCount + STALE_TIME mechanism does just that. Also, what if we want to keep the oldest version of the entry? In other words, why is the oldest entry automatically stale?

common/fixation_entry.go
```diff
@@ -256,7 +277,7 @@ func (fs *FixationStore) getUnmarshaledEntryForBlock(ctx sdk.Context, safeIndex
 		var entry types.Entry
 		fs.cdc.MustUnmarshal(iterator.Value(), &entry)
 
-		if entry.GetBlock() <= block {
+		if entry.Block <= block {
```
oren-lava, line 280: Question: why did you change from ".GetBlock()" to ".Block"? I always use getters to avoid nil dereference. Isn't it a better practice to always use getters?

common/fixation_entry.go
```diff
@@ -308,6 +329,23 @@ func (fs *FixationStore) GetEntry(ctx sdk.Context, index string, entryData codec
...
+// putEntry decrements the refcount of an entry and marks for staleness if needed
+func (fs *FixationStore) putEntry(ctx sdk.Context, entry types.Entry) {
+	if entry.Refcount == 0 {
+		panic("Fixation: prefix " + fs.prefix + ": negative refcount safeIndex: " + entry.Index)
+	}
+
+	entry.Refcount -= 1
+
+	if entry.Refcount == 0 {
+		// never overflows because because ctx.BlockHeight is int64
```
oren-lava, line 341: nit: double because

common/fixation_entry.go
```diff
@@ -320,18 +358,12 @@ func (fs *FixationStore) PutEntry(ctx sdk.Context, index string, block uint64) {
...
-	if entry.Refcount == 0 {
-		// never overflows because because ctx.BlockHeight is int64
-		entry.StaleAt = uint64(ctx.BlockHeight()) + uint64(types.STALE_ENTRY_TIME) - 1
+	if entry.Block != block {
+		panic("PutEntry with block mismatch index: " + index +
+			" got " + strconv.Itoa(int(entry.Block)) + " expected " + strconv.Itoa(int(block)))
 	}
 
-	fs.setEntry(ctx, entry)
+	fs.putEntry(ctx, entry)
```
oren-lava, line 366: why do we need an additional putEntry? why not put all the code in one function?

oren-lava (changes requested):

common/fixation_entry.go
```diff
@@ -69,6 +71,7 @@ type FixationStore struct {
 	storeKey sdk.StoreKey
 	cdc      codec.BinaryCodec
 	prefix   string
+	tstore   TimerStore
```
Yaroms, line 74: lets call it timerStore for clearance

common/fixation_entry.go
```diff
@@ -26,7 +26,9 @@ import (
 //    - GetEntry(index, *entry): get a copy (and reference) of the latest version of an entry
 //    - PutEntry(index, block): drop a reference of a version of an entry
 //    - [TBD] RemoveEntry(index): mark an entry as unavailable for new GetEntry() calls
-//    - GetAllEntryIndex(): get all the entries indices (without versions)
+//    - GetAllEntryIndices(): get all the entries indices (without versions)
```
Yaroms, line 29: do we need a module migration here?

common/fixation_entry.go
```diff
@@ -344,8 +382,15 @@ func (fs *FixationStore) createStoreKey(index string) string {
...
+func (fs *FixationStore) AdvanceBlock(ctx sdk.Context) {
+	fs.tstore.Tick(ctx)
+}
+
 // NewFixationStore returns a new FixationStore object
 func NewFixationStore(storeKey sdk.StoreKey, cdc codec.BinaryCodec, prefix string) *FixationStore {
 	fs := FixationStore{storeKey: storeKey, cdc: cdc, prefix: prefix}
+	callback := func(ctx sdk.Context, data string) { fs.deleteStaleEntries(ctx, data) }
+	tstore := NewTimerStore(storeKey, cdc, prefix).WithCallbackByBlockHeight(callback)
+	fs.tstore = *tstore
```
Yaroms, line 394: what happens if we set only one type of callback?

Yaroms (changes requested):

common/fixation_entry.go
orenl-lava: not needed - just another API that leverages the existing data format.

Yaroms (approved):

omerlavanet (approved):
very very nice

PR #415, opened 2023-04-13
CNS-376 fixation versioning and migration

orenl-lava:
This PR adds:

1. Versioning and migration for FixationStore. Migrations are expected to be called for each instance explicitly by the users.

2. Migration from v1 to v2 that fixes the refcount of head entries (i.e. latest versions of entries): previously they were initialized with refcount 0, but they need to have refcount 1. It also sets timers for entries that have StaleAt field set: previously it didn't use timers, so now explicitly initialized them for existing entries.                                                                               

3.Module migrations for "plan" and "project" that invoke the fixation migration (v1 to v2) on their respective instances.


common/fixation_migrations.go
```diff
@@ -0,0 +1,77 @@
...
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/lavanet/lava/common/types"
+)
+
+func prefixForErrors(from uint64) string {
+	return fmt.Sprintf("FixationStore: migration from version %d", from)
+}
+
+func (fs *FixationStore) MigrateVersion(ctx sdk.Context) (err error) {
+	from := fs.getVersion(ctx)
```
Yaroms, line 16: the version was never set, how are we handling this?

common/fixation_migrations.go
```diff
@@ -0,0 +1,77 @@
...
+	to := types.FixationVersion()
+
+	for from < to {
+		switch from {
+		case 1:
```
Yaroms, line 21: why is the current version 1?

common/fixation_migrations.go
```diff
@@ -0,0 +1,77 @@
...
+
+	fs.setVersion(ctx, to)
+	return nil
+}
+
+// migrate1to2: fix refcounts
+//   - correct refcount of head (most recent) by adding one (because new
+//     entries used to begin with refcount 0 instead of refcount 1)
+//   - correct negative refcounts if found any (for extra care)
+func (fs *FixationStore) migrate1to2(ctx sdk.Context) error {
```
Yaroms, line 40: why dont just use the GetEntry on all of the indexes instead of doing it manually.
also refcounts cant be zero since we have protection for it

Yaroms (changes requested):

common/fixation_migrations.go
orenl-lava: getVersion() returns 1 if no version is set.

common/fixation_migrations.go
orenl-lava: > why dont just use the GetEntry on all of the indexes instead of doing it manually.

For the head entries (i.e. most recent version of entries), indeed GetEntry() would suffice.
But we do three additional things:
- for entries with refcount zero and StaleAt not set, need to set the StaleAt (sanity check, shouldn't happen)
- for entries with StaleAt set: need to start a timer to ensure proper cleanup. (And if the StaleAt has already passed, set a timer for the next block - need to add that to the logic).
- for all entries, check that refcount is not too big - as it would mean underflow occurred before (sanity check, shouldn't happen).
- 
And for these last three items we must do the work manually.

> also refcounts cant be zero since we have protection for it

refcounts can be zero - for entries that are in the middle (i.e. not oldest). That's part of what the fixes of refcount fix.

x/projects/keeper/migrations.go
```diff
@@ -0,0 +1,28 @@
+package keeper
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+type Migrator struct {
+	keeper Keeper
```
oren-lava, line 10: same as plans' migrator comment

oren-lava (changes requested):

x/plans/keeper/migrations.go
```diff
@@ -0,0 +1,25 @@
+package keeper
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+type Migrator struct {
```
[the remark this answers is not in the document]
orenl-lava: Not "should" but "may". Your example is for "bank" specific module migration. But other modules don't necessarily use it. Specifically, our modules do not have an `export/` subdir.

common/fixation_migrations.go
```diff
@@ -0,0 +1,77 @@
...
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/lavanet/lava/common/types"
+)
+
+func prefixForErrors(from uint64) string {
+	return fmt.Sprintf("FixationStore: migration from version %d", from)
+}
+
+func (fs *FixationStore) MigrateVersion(ctx sdk.Context) (err error) {
```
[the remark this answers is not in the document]
orenl-lava: As we discussed, for FixationStore (and later TimerStore) we keep it as is, because (1) they are not independent modules, (2) to avoid exposing internal API (or copying existing code). This decision can be revisited later.

common/fixation_migrations_test.go
```diff
@@ -0,0 +1,91 @@
+package common
```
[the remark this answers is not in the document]
orenl-lava: see comment above (https://github.com/lavanet/lava/pull/415#discussion_r1170080662)

x/projects/keeper/migrations.go
orenl-lava: same as plans' migrator reply comment

common/fixation_migrations.go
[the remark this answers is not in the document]
orenl-lava: As for "x/plans" and "x/projects", their upgrade involves just calling FixationStore's MigrateVersion(), not necessary to create a separate directory + function for that. See updated PR.

Yaroms (dismissed):

Yaroms (approved):

Yaroms (approved):

PR #417, opened 2023-04-13
CNS-382 fixation update prefix

orenl-lava:
This PR creates a new migration v2 to v3 of FixationStore to fix the use of prefix.
It also creates an upgrade to modules "plan" and "project" to run this migration.
Finally, it changes the prefix used by module "plan" to be consistent with our naming.

### DIFF
--- a/common/fixation_entry.go
+++ b/common/fixation_entry.go
@@ -418,8 +418,8 @@ func (fs *FixationStore) GetAllEntryVersions(ctx sdk.Context, index string, stal
 	return fs.getEntryVersionsFilter(ctx, index, 0, filter)
 }
 
-func (fs *FixationStore) createStoreKey(index string) string {
-	return types.EntryPrefix + fs.prefix + index
+func (fs *FixationStore) createEntryStoreKey(index string) string {
+	return fs.prefix + types.EntryPrefix + index
 }
 
 func (fs *FixationStore) AdvanceBlock(ctx sdk.Context) {
--- a/common/fixation_entry.go
+++ b/common/fixation_entry.go
@@ -80,16 +80,16 @@ type FixationStore struct {
 	tstore   TimerStore
 }
 
-func (fs *FixationStore) getStore(ctx sdk.Context, index string) *prefix.Store {
+func (fs *FixationStore) getEntryStore(ctx sdk.Context, index string) *prefix.Store {
 	store := prefix.NewStore(
 		ctx.KVStore(fs.storeKey),
-		types.KeyPrefix(fs.createStoreKey(index)))
+		types.KeyPrefix(fs.createEntryStoreKey(index)))
 	return &store
 }
 
 // getEntry returns an existing entry in the store
 func (fs *FixationStore) getEntry(ctx sdk.Context, safeIndex string, block uint64) (entry types.Entry) {
-	store := fs.getStore(ctx, safeIndex)
+	store := fs.getEntryStore(ctx, safeIndex)
 	byteKey := types.EncodeKey(block)
 	b := store.Get(byteKey)
 	if b == nil {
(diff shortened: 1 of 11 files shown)


common/fixation_entry.go
```diff
@@ -379,7 +379,7 @@ func (fs *FixationStore) removeEntry(ctx sdk.Context, index string, block uint64
 }
 
 func (fs *FixationStore) createStoreKey(index string) string {
-	return types.EntryPrefix + fs.prefix + index
```
Yaroms, line 382: how does this help us?
i feel like changing the keys is a dangerous game

orenl-lava: The purpose of the prefix provider by the user (fs.prefix) is to allow namespacing controlled by the user. As such, it should come *first* in the prefix/key. Otherwise, we're effectively hijacking the prefix "Entry_..." used by the fixation (and which cannot be controlled by the user). It was a mistake to set the order as it is now - it should be reversed.
There migration logic will take care of updating the store properly.

x/plans/keeper/migrations.go
```diff
@@ -59,3 +60,21 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
...
+// - Replace the store prefix from module-name ("plan") to "plan-fs"
+func (m Migrator) Migrate3to4(ctx sdk.Context) error {
+	const V4_PlanFixationStorePrefix = "plan"
+
+	// Note: MigratePrefix() already calls MigrateVersion() and thus will perform
+	// FixationStore version migration if needed, using the old prefix first.
+	// (This is required because otherwise the usual fixation-store MigrateVersion
+	// would wrongly use the new prefix).
+
+	if err := m.keeper.plansFS.MigratePrefix(ctx, V4_PlanFixationStorePrefix); err != nil {
```
Yaroms, line 75: why do we explicit call MigratePrefix and not the MigrateVertion?

common/fixation_migrate.go
```diff
@@ -77,3 +113,26 @@ func fixationMigrate1to2(ctx sdk.Context, fs *FixationStore) error {
...
 	return nil
 }
+
+// fixationMigrate2to3: fix refcounts
+//   - instead of "EntryPrefix + fs.prefix" -> "fs.prefix + EntryPrefix"
+//   - instead of "EntryIndexPrefix + fs.prefix"(x2) -> "fs.prefix + EntryIndexPrefix"
+func fixationMigrate2to3(ctx sdk.Context, fs *FixationStore) error {
+	const (
+		// copy of old values
+		V2_EntryIndexPrefix string = "Entry_Index_"
```
Yaroms, line 123: old values and new values are the same?

common/fixation_migrate.go
```diff
@@ -38,6 +41,39 @@ func (fs *FixationStore) MigrateVersion(ctx sdk.Context) (err error) {
...
+	fs.prefix = oldPrefix
+	err = fs.MigrateVersion(ctx)
+	fs.prefix = newPrefix
+
+	if err != nil {
+		return err
+	}
+
+	// prefix migration
+	fs.doMigratePrefix(ctx, oldPrefix, newPrefix)
```
Yaroms, line 59: why do we need to migrate a second time?

x/plans/keeper/migrations.go
Yaroms: cant we migrate from the old prefix directly to the new prefix(with the key change)

common/fixation_migrate.go
orenl-lava: Usure what you mean?

`fs.MigratePrefix()` is an API for users of FixationStore that wants to change their prefix, i.e. a data migration requested by the module that uses it.  It calls `fs.doMigratePrefix()` to do the actual work. But first it calls `fs.MigrateVersion()` to apply any internal updates of `fs.MigratePrefix()` (i.e. a migration decided/needed by the FixationStore) - and it does so using the *old* prefix (otherwise the *new* prefix would be used before it's in effect).

common/fixation_migrate.go
orenl-lava: in case the current value change in a newer version.

x/plans/keeper/migrations.go
orenl-lava: I will rename "MigratePrefix()" to "MigrateVersionWithPrefix()" to make it much clearer.

Yaroms (approved):

PR #422, opened 2023-04-14
removed obsolete map

omerlavanet:


orenl-lava (approved):
Fix looks fine.
Although, you really need to refactor `Start()`.
And replace dummy functions created for the sake of defer - with real functions (refactor!).

PR #425, opened 2023-04-16
CNS-379: enable lint on test files and fix lint issues

oren-lava:


oren-lava:
Closed due to being old. Will re-open from current latest main commit.

PR #427, opened 2023-04-16
CNS-385: project propagate charge of CU onward

orenl-lava:
Charging CUs to a project always happens to a certain fixation version (by block). 
When doing so, the charge should propagate from that version onward: until the
most recent version, or until a re-charged version (i.e. after subscription month 
end) is reached, as marked by a new snapshot.


x/projects/keeper/project.go
```diff
@@ -62,9 +63,20 @@ func (k Keeper) AddKeysToProject(ctx sdk.Context, projectID string, adminKey str
...
-func (k Keeper) ChargeComputeUnitsToProject(ctx sdk.Context, project types.Project, cu uint64) (err error) {
-	project.UsedCu += cu
-	return k.projectsFS.ModifyEntry(ctx, project.Index, uint64(ctx.BlockHeight()), &project)
+func (k Keeper) ChargeComputeUnitsToProject(ctx sdk.Context, project types.Project, blockHeight uint64, cu uint64) (err error) {
+	blocks := k.projectsFS.GetEntryVersionsRange(ctx, project.Index, blockHeight, uint64(commontypes.STALE_ENTRY_TIME))
+
+	for _, block := range blocks {
+		var proj types.Project
+		k.projectsFS.ReadEntry(ctx, project.Index, block, &proj)
+		if proj.Snapshot != project.Snapshot {
```
Yaroms, line 72: what if you have snapshots of 1,2,3 and you need to charge snapshot 2?
this means you will break at one before you even reach 2

Yaroms (changes requested):

x/projects/keeper/project.go
orenl-lava: That cannot happen: snapshot happens once a month, and should not be kept - for the purpose of accepting payments - for as long.  But even if they stay around, this behavior is still correct: a payment to a project version should propagate forward until the next recharge (= snapshot) of that project, and not beyond/into the new period of accounting.

Yaroms (approved):

Yaroms (approved):

PR #433, opened 2023-04-20
GROW-640 Add origin on the relays

tabakuj:
Add possibility to ignore certain userAgents


ranlavanet:
@nirtayeb can we dismiss this Draft?

tabakuj:
> @nirtayeb can we dismiss this Draft?



> @nirtayeb can we dismiss this Draft?

we need this. since we need to update gateway analytics project also that is the reason why its in draft

PR #435, opened 2023-04-25
PRT-514 quorum select first provider

Aleksao998:
## Description
This PR implement first of many strategies we are going to have for sending relays. The strategy which is implemented in this PR Is LATENCY. 

## How it works:
When a user sends a request to the consumer. The consumer will pick a subset of providers (for now hardcoded to 2) and it will send a relay to all selected providers. Whoever returns the first non-error response it will be served to the client. If all providers return an error the last response is returned to the consumer

## Manual tests:
1. Start local network with 3 providers and set optimiser to return 2 providers
2. Execute queries multiple times, and for each request 2 providers needs to get return
3. Stop 1 provider and execute multiple requests, both active needs to get payment for each
4. Stop 1 more, now we only have 1 active which is less than optimiser is returning, that one should still serve relays and get payment


ranlavanet (changes requested):

ranlavanet (approved):

protocol/rpcconsumer/rpcconsumer_server.go
```diff
@@ -186,83 +188,166 @@ func (rpccs *RPCConsumerServer) sendRelayToProvider(
...
-			if backoff_ {
-				backOffDuration = lavasession.BACKOFF_TIME_ON_FAILURE
+			// Extract fields from the sessionInfo
+			singleConsumerSession := sessionInfo.Session
+			epoch := sessionInfo.Epoch
+			reportedProviders := sessionInfo.ReportedProviders
+
+			relayRequest, err := lavaprotocol.ConstructRelayRequest(goroutineCtx, privKey, lavaChainID, chainID, &localRelayRequestData, providerPublicAddress, singleConsumerSession, int64(epoch), reportedProviders)
+			if err != nil {
+				responses <- &relayResponse{
```
omerlavanet, line 226: this pattern repeats itself on every return, we can use defer to do it once instead, i owuld use uniquely named variables like errRet instead of err to make sure scoping isn't overwritten

protocol/lavasession/consumer_session_manager.go
```diff
@@ -20,6 +20,14 @@ const (
 	debug = false
 )
 
+type SessionInfo struct {
```
omerlavanet, line 23: move to consumer_types

protocol/lavasession/consumer_session_manager.go
```diff
@@ -224,127 +232,157 @@ func (csm *ConsumerSessionManager) validatePairingListNotEmpty() uint64 {
...
-		if pairingEpoch != sessionEpoch {
-			// pairingEpoch and SessionEpoch must be the same, we validate them here if they are different we raise an error and continue with pairingEpoch
-			utils.LavaFormatError("sessionEpoch and pairingEpoch mismatch", nil, utils.Attribute{Key: "sessionEpoch", Value: sessionEpoch}, utils.Attribute{Key: "pairingEpoch", Value: pairingEpoch})
-			sessionEpoch = pairingEpoch
+		// If we do not have enough fetch more
+		sessionWithProviderMap, err = csm.getValidConsumerSessionsWithProvider(tempIgnoredProviders, cuNeededForSession, requestedBlock)
+
+		// If error exists but we have sessions, return them
+		if err != nil && len(sessions) != 0 {
+			return sessions, err
```
omerlavanet, line 374: the caller doesn't check sessions if there is an error, so if we fail to get more we just bail out

protocol/lavasession/consumer_session_manager.go
```diff
@@ -224,127 +232,157 @@ func (csm *ConsumerSessionManager) validatePairingListNotEmpty() uint64 {
...
 			}
-			continue
 		}
 
-		if pairingEpoch != sessionEpoch {
-			// pairingEpoch and SessionEpoch must be the same, we validate them here if they are different we raise an error and continue with pairingEpoch
-			utils.LavaFormatError("sessionEpoch and pairingEpoch mismatch", nil, utils.Attribute{Key: "sessionEpoch", Value: sessionEpoch}, utils.Attribute{Key: "pairingEpoch", Value: pairingEpoch})
-			sessionEpoch = pairingEpoch
+		// If we do not have enough fetch more
+		sessionWithProviderMap, err = csm.getValidConsumerSessionsWithProvider(tempIgnoredProviders, cuNeededForSession, requestedBlock)
```
omerlavanet, line 370: we have discussed getValidConsumerSessionsWithProvider should get an optional argument that specifies how many exactly to expect, this should be a small change in the providerOptimizer code

protocol/lavasession/consumer_session_manager.go
```diff
@@ -356,32 +394,71 @@ func (csm *ConsumerSessionManager) getValidProviderAddress(ignoredProvidersList
 	}
 
 	providers := csm.providerOptimizer.ChooseProvider(csm.validAddresses, ignoredProvidersList, cu, requestedBlock, OptimizerPerturbation)
```
omerlavanet, line 396: this can return an empty list without an error, we have to check there is at least one or err

omerlavanet (commented):
missing change to provider optimizer wanted providers number from 1 to 5, so the entire thing wouldn't work, also let's make it via a flag/config using viper

protocol/lavasession/consumer_session_manager.go
```diff
@@ -356,32 +394,71 @@ func (csm *ConsumerSessionManager) getValidProviderAddress(ignoredProvidersList
...
+				return sessionWithProviderMap, nil
+			}
+		}
+
+		// If we do not have enough fetch more
+		providerAddresses, err = csm.getValidProviderAddresses(ignoredProviders.providers, cuNeededForSession, requestedBlock)
+
+		// If error exists but we have providers, return them
+		if err != nil && len(sessionWithProviderMap) != 0 {
+			return sessionWithProviderMap, err
```
omerlavanet, line 453: caller doesnt check this map is there is an error

omerlavanet (changes requested):
don't forget to change optimizer settings to make use of this, configured on 1 it will be returning onely one provider

protocol/lavasession/provider_session_manager.go
```diff
@@ -159,7 +159,7 @@ func (psm *ProviderSessionManager) GetDataReliabilitySession(address string, epo
 
 func (psm *ProviderSessionManager) GetSession(ctx context.Context, address string, epoch uint64, sessionId uint64, relayNumber uint64) (*SingleProviderSession, error) {
 	if !psm.IsValidEpoch(epoch) { // fast checking to see if epoch is even relevant
-		utils.LavaFormatError("GetSession", InvalidEpochError, utils.Attribute{Key: "RequestedEpoch", Value: epoch}, utils.Attribute{Key: "blockedEpochHeight", Value: psm.blockedEpochHeight}, utils.Attribute{Key: "blockDistanceForEpochValidity", Value: psm.blockDistanceForEpochValidity})
+		utils.LavaFormatError("GetSessions", InvalidEpochError, utils.Attribute{Key: "RequestedEpoch", Value: epoch}, utils.Attribute{Key: "blockedEpochHeight", Value: psm.blockedEpochHeight}, utils.Attribute{Key: "blockDistanceForEpochValidity", Value: psm.blockDistanceForEpochValidity})
```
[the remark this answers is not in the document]
ranlavanet: @Aleksao998 , using replace all without validating what the change does is risky, always use F2 and rename the method. 
this is an example why its important to use F2 instead of replace all.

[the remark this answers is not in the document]
Aleksao998: Great catch, I haven't used the replace all, but rename variable from IDE, no idea why it changed all these. You can see in the code that there are many other _GetSession_ which were not changed, like the name of this method above. 

Anyways, it was my mistake not to validate all the changed entries. Thank you both for finding this 🙏🙏

Fixed e79a1b274b095bb14f4ac3262751b308cbf936b9

protocol/lavasession/provider_session_manager.go
```diff
@@ -106,7 +106,7 @@ func (psm *ProviderSessionManager) getOrCreateDataReliabilitySessionWithConsumer
 func (psm *ProviderSessionManager) GetDataReliabilitySession(address string, epoch uint64, sessionId uint64, relayNumber uint64, pairedProviders int64) (*SingleProviderSession, error) {
 	// validate Epoch
 	if !psm.IsValidEpoch(epoch) { // fast checking to see if epoch is even relevant
-		utils.LavaFormatError("GetSession", InvalidEpochError, utils.Attribute{Key: "RequestedEpoch", Value: epoch})
+		utils.LavaFormatError("GetSessions", InvalidEpochError, utils.Attribute{Key: "RequestedEpoch", Value: epoch})
```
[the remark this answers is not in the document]
Aleksao998: Fixed e79a1b274b095bb14f4ac3262751b308cbf936b9

protocol/lavasession/provider_session_manager_test.go
```diff
@@ -649,7 +649,7 @@ func TestPSMUsageSync(t *testing.T) {
 						// this is not a first relay so we expect this to work
 						require.NoError(t, err, "sessionID %d relayNum %d storedRelayNum %d epoch %d, history %s", sessionStoreTest.sessionID, sessionStoreTest.relayNum+1, sessionStoreTest.session.RelayNum, sessionStoreTest.epoch, sessionStoreTest.history)
 						require.Same(t, session, sessionStoreTest.session)
-						sessionStoreTest.history = append(sessionStoreTest.history, ",GetSession")
+						sessionStoreTest.history = append(sessionStoreTest.history, ",GetSessions")
```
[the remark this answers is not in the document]
Aleksao998: Fixed e79a1b274b095bb14f4ac3262751b308cbf936b9

protocol/lavasession/provider_session_manager_test.go
```diff
@@ -664,7 +664,7 @@ func TestPSMUsageSync(t *testing.T) {
 							sessionStoreTest.history = append(sessionStoreTest.history, ",RegisterGet")
 						} else {
 							sessionStoreTest.session = session
-							sessionStoreTest.history = append(sessionStoreTest.history, ",GetSession")
+							sessionStoreTest.history = append(sessionStoreTest.history, ",GetSessions")
```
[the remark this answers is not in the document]
Aleksao998: Fixed e79a1b274b095bb14f4ac3262751b308cbf936b9

protocol/lavasession/provider_session_manager_test.go
```diff
@@ -726,7 +726,7 @@ func TestPSMUsageSync(t *testing.T) {
 		}
 	}
 	// .IsValidEpoch(uint64(request.RelaySession.Epoch))
-	// .GetSession(context.Background(),consumerAddressString, uint64(request.Epoch), request.SessionId, request.RelayNum)
+	// .GetSessions(context.Background(),consumerAddressString, uint64(request.Epoch), request.SessionId, request.RelayNum)
```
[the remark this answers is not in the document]
Aleksao998: Fixed e79a1b274b095bb14f4ac3262751b308cbf936b9

protocol/lavasession/consumer_session_manager.go
Aleksao998: Moved 6dfcd732f31c68124804711c5a62cfefacce56e7

protocol/lavasession/consumer_session_manager.go
Aleksao998: The logic of this method prevents calling ChooseProvider if it will return 0. Before calling we make sure that there are valid providers. 

```
	totalValidLength := validAddressesLength - ignoredProvidersListLength
	if totalValidLength <= 0 {
		utils.LavaFormatDebug("Pairing list empty", utils.Attribute{Key: "Provider list", Value: csm.validAddresses}, utils.Attribute{Key: "IgnoredProviderList", Value: ignoredProvidersList})
		err = PairingListEmptyError
		return
	}
```

But for the sake of certainty and the logic in the ChooseProvider (which can change), i agree we can add double check just to be sure:

Added: 3678366158ad49c457980c745d9fa2276d185bf5

protocol/rpcconsumer/rpcconsumer_server.go
```diff
@@ -20,7 +20,8 @@ import (
 )
 
 const (
-	MaxRelayRetries = 4
+	MaxRelayRetries  = 4
+	numberOfSessions = 2
```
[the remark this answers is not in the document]
Aleksao998: Great catch, thank you 🙏
Removed f570d451956eb68ae70773673a6e70910c1dea9e

protocol/lavasession/consumer_session_manager.go
Aleksao998: I think that it should check this map even if there is an error. 

This method:
```
// If we do not have enough fetch more
providerAddresses, err = csm.getValidProviderAddresses(ignoredProviders.providers, cuNeededForSession, requestedBlock)
```
can return either, a list of providers (len!=0) or an error. The only error it can return is **_PairingListEmptyError_**. 

The scenario which can happen is that we've requested for example **5** providers, but while validating we removed a few (let's say 2). This is done here https://github.com/lavanet/lava/blob/PRT-514-quorum-select-first-provider/protocol/lavasession/consumer_session_manager.go#L430.

That means that when we exited for loop we do not have the desired number of providers (which was 5), we only have 3 providers. So we would request additional ones https://github.com/lavanet/lava/blob/PRT-514-quorum-select-first-provider/protocol/lavasession/consumer_session_manager.go#L449. 

What can happen is that the second time when we call **getValidProviderAddresses** there are no valid providers. So we have 3 valid from the previous cycle and no more. WHich means that even if we now have an error **_PairingListEmptyError_**, we have some valid providers to return.

Aleksao998: But something else i see should be fixed here is that if we are returning the sessionWithProviderMap we should not return an error:
Fixed: daaf96276a04ec70c5af95970cea84456fc7ccd1

protocol/lavasession/consumer_session_manager.go
Aleksao998: Same explanation as for this question, we can continue the discussion on this comment https://github.com/lavanet/lava/pull/435#discussion_r1208646820

Also fixed returned error:
Fixed: 19543de075259d24177d7e4faed04ed6bb30c986

Aleksao998: Resolving this so we have conversation only on 1 place, because it is a same logic

protocol/lavasession/consumer_session_manager.go
Aleksao998: This also impacts this comment https://github.com/lavanet/lava/pull/435#discussion_r1208501673

protocol/rpcconsumer/rpcconsumer_server.go
```diff
@@ -186,83 +188,166 @@ func (rpccs *RPCConsumerServer) sendRelayToProvider(
...
 	go func() {
-		new_ctx := context.Background()
-		new_ctx, cancel := context.WithTimeout(new_ctx, common.DataReliabilityTimeoutIncrease)
-		defer cancel()
-		err2 := rpccs.cache.SetEntry(new_ctx, relayRequest, chainMessage.GetInterface().Interface, nil, chainID, dappID, relayResult.Reply, relayResult.Finalized) // caching in the portal doesn't care about hashes
-		if err2 != nil && !performance.NotInitialisedError.Is(err2) {
-			utils.LavaFormatWarning("error updating cache with new entry", err2)
+		responsesReceived := 0
+		relayReturned := false
+		for {
```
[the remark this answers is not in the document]
Aleksao998: Great catch, thank you 🙏 How about 10 seconds? Or should I use something much less? 

Added: c5de24e5d35539ca14a699125e92098409e5736f

protocol/rpcconsumer/rpcconsumer_server.go
```diff
@@ -186,83 +188,166 @@ func (rpccs *RPCConsumerServer) sendRelayToProvider(
...
-		relayResult.Reply = reply
-		lavaprotocol.UpdateRequestedBlock(relayResult.Request.RelayData, reply) // update relay request requestedBlock to the provided one in case it was arbitrary
-		err = rpccs.consumerSessionManager.OnSessionUnUsed(singleConsumerSession)
-		return relayResult, err
-	}
+	// Iterate over the sessions map
+	for providerPublicAddress, sessionInfo := range sessions {
+		// Launch a separate goroutine for each session
+		go func(providerPublicAddress string, sessionInfo *lavasession.SessionInfo) {
+			goroutineCtx, goroutineCtxCancel := context.WithCancel(context.Background())
```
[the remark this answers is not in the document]
Aleksao998: Fixed: 189da9435c854f98edbda497f0668ff5f5aebfea

Now each goroutine should have its own context with the properties inherited from the original context passed to the sendRelayToProvider method

protocol/rpcconsumer/rpcconsumer_server.go
Aleksao998: Sure
Fixed 2241c0f5da1055375ef4b4d418224fd7cd1270b1

protocol/lavasession/consumer_session_manager.go
```diff
@@ -356,32 +394,71 @@ func (csm *ConsumerSessionManager) getValidProviderAddress(ignoredProvidersList
...
+	currentEpoch := csm.atomicReadCurrentEpoch() // reading the epoch here while locked, to get the epoch of the pairing.
 	if ignoredProviders.currentEpoch < currentEpoch {
 		utils.LavaFormatDebug("ignoredProviders epoch is not the current epoch, resetting ignoredProviders", utils.Attribute{Key: "ignoredProvidersEpoch", Value: ignoredProviders.currentEpoch}, utils.Attribute{Key: "currentEpoch", Value: currentEpoch})
 		ignoredProviders.providers = make(map[string]struct{}) // reset the old providers as epochs changed so we have a new pairing list.
 		ignoredProviders.currentEpoch = currentEpoch
 	}
 
-	providerAddress, err = csm.getValidProviderAddress(ignoredProviders.providers, cuNeededForSession, requestedBlock)
+	// Fetch provider addresses
+	providerAddresses, err := csm.getValidProviderAddresses(ignoredProviders.providers, cuNeededForSession, requestedBlock)
```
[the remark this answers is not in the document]
Aleksao998: I stumbled on the issue with the approach with optional params, that's why I decided not to add it. Let me explain:

The biggest challenge with the current implementation (one we used so far) is that we are validating if the provider is valid in multiple places. 

So what do we have:
1. **getValidProviderAddresses** which returns an array of providers using optimizer (chooseProvider) or  PairingListEmptyError
2. **getValidConsumerSessionsWithProvider** which calls getValidProviderAddresses and before returning providers it validates providers https://github.com/lavanet/lava/blob/main/protocol/lavasession/consumer_session_manager.go#L382
3. **GetSessions** which calls getValidConsumerSessionsWithProvider and also before returning validates providers on many places (for example https://github.com/lavanet/lava/blob/main/protocol/lavasession/consumer_session_manager.go#L295) 

Our goal is to return as many providers as possible from what we initialized in the optimizer.

Let's focus on the **getValidConsumerSessionsWithProvider** method to analyze the problem:

The situation can be like this. Optimizer is initialized to return 5 providers. Whenever we call **getValidProviderAddresses** optimized will return either up to 5 providers or PairingListEmptyError. We call getValidProviderAddresses and it returns 5 but out of these 5, only 3 are valid. So we need to ping again and try to find the remaining 2. If we would use optional params we would ask again optimizer to return 2, but nothing is guaranteeing that these 2 will pass this validation https://github.com/lavanet/lava/blob/main/protocol/lavasession/consumer_session_manager.go#L382. With this implementation, if we see we are missing 2 providers out of the initial 5, we would ask the optimizer again to return providers, we would again iterate over the list and if at any moment we have the wanted number of providers we will return. https://github.com/lavanet/lava/blob/PRT-514-quorum-select-first-provider/protocol/lavasession/consumer_session_manager.go#L443. If again there no valid, we will ask the optimizer for more and so on till we either have wanted number or we have no more to check. 

I do not see how the optional param in **chooseProvider** method would simplify the logic, or how we can bypass the validations in both the **getValidConsumerSessionsWithProvider** and **GetSessions**

On the other hand, optional parameters in go are not implemented by design. Ofc there are ways to create optional param behavior but I am trying to avoid it as much as possible. Few which are familiar to me: 
1. **Variadic Functions**
```
func sum(test string, nums ...int) {

	fmt.Println(test)
	if len(nums) != 0 {
		fmt.Println(nums[0])
	}
}

func main() {
	sum("a", 1)
	sum("a")
}
```
The issue I see with them is that I am not enforcing how many arguments can exist. So someone could call sum("a", 1,2,4) and the function will still look good. Ofc we could handle all of this internally but compiler would not catch the issue

2. We could use structs
```
type Params struct {
  a, b, c int
}

func doIt(p Params) int {
  return p.a + p.b + p.c 
}

// you can call it without specifying all parameters
doIt(Params{a: 1, c: 9})
```

3. Functional Options
```
func NewConfig(options ...OptionFunc) *Config {
	config := &Config{}
	for _, option := range options {
		option(config)
	}
	return config
}
```

Can you please explain how we can utilize optional params to make this flow more optimal 🙏

protocol/lavasession/consumer_session_manager.go
Aleksao998: Same as for https://github.com/lavanet/lava/pull/435#discussion_r1208676326

Aleksao998:
> missing change to provider optimizer wanted providers number from 1 to 5, so the entire thing wouldn't work, also let's make it via a flag/config using viper

The scope of this task didn't include modifying the optimizer to support multiple adjustable numbers of providers. While testing I've tried hardcoding this value to different ones so we can test it. That's why i had this variable as leftover https://github.com/lavanet/lava/pull/435#discussion_r1208365782

Anyways if you want, I can push it to the PR as well 😃💪

omerlavanet (changes requested):
everything good except for the 10s timeout that might be too low in some cases

protocol/rpcconsumer/rpcconsumer_server.go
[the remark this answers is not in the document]
Aleksao998: Timeout fix not to be constant 203c9933e7aff01e5135c64c36a2f9264821dad0

Aleksao998:
> everything good except for the 10s timeout that might be too low in some cases

Here are the fixes:
203c9933e7aff01e5135c64c36a2f9264821dad0
9aecd1d789bdc0cf16b25477495e4656324c4b16

protocol/rpcconsumer/rpcconsumer_server.go
```diff
@@ -208,8 +214,18 @@ func (rpccs *RPCConsumerServer) sendRelayToProvider(
 	// Make a channel for all providers to send responses
 	responses := make(chan *relayResponse, len(sessions))
 
+	// Set response timeout to 0
+	responseTimeout := 0 * time.Second
+
 	// Iterate over the sessions map
 	for providerPublicAddress, sessionInfo := range sessions {
+		// Calculate relay timeout
+		relayTimeout := extraRelayTimeout + common.GetTimePerCu(sessionInfo.Session.LatestRelayCu) + common.AverageWorldLatency
```
omerlavanet, line 223: this is not necessary as all sessionInfo.Session.LatestRelayCu are equal to chainMessage.GetServiceApi().ComputeUnits, so there is no need to find the maximum, we can set the timeout value outside of this function and remove a few lines of code

omerlavanet (changes requested):

omerlavanet (approved):

Aleksao998:
fixe 56aeba7e96bc0d3f6a0c3337820813e76a292342

PR #437, opened 2023-04-25
CNS-407: Migrator for project and plan changes due to CNS-336

oren-lava:
The plan and project protobuf changed in CNS-336. In this PR, I've implemented a migrator for these changes


x/plans/keeper/migrations.go
```diff
@@ -22,3 +25,39 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
 	}
 	return nil
 }
+
+func (m Migrator) Migrate3to4(ctx sdk.Context) error {
```
orenl-lava, line 31: 1. The migration from 2 to 3 didn't happen yet, so let's combine this into the exiting/upcoming Migrate2to3().

x/plans/keeper/migrations.go
```diff
@@ -22,3 +25,39 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
...
+	planIndices := m.keeper.GetAllPlanIndices(ctx)
+
+	for _, planIndex := range planIndices {
```
orenl-lava, line 34: This would update only the latest version of each plan, and not older fixated versions that may still be in use. You need to use fixation's GetAllEntryVersions(ctx, entry, true) to get the list of block (versions) for each index, and (nested) loop on them.

x/plans/keeper/migrations.go
```diff
@@ -22,3 +25,39 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
...
+		var oldPlanStruct v3.Plan
+		if found := m.keeper.plansFS.FindEntry(ctx, planIndex, uint64(ctx.BlockHeight()), &oldPlanStruct); !found {
```
orenl-lava: With the list of blocks from the previous comment, you can call plansFS.ReadEntry() to get each specific version (it avoids the iterator lookup).

x/plans/keeper/migrations.go
orenl-lava, line 31: 2. We decided to put actual migration code in v3/... subdir if possible. In this case, following x/bank/... example, introduce `x/plans/migrations/v3/store.go:MigrateStore(...)` and call it from here. 
Because the migration code needs access to `keeper.plansFS`, maybe we can add an API in `x/plans/keeper/migrations.go` to expose it, with a clear comment stating the API should only be used in migrations>

x/plans/keeper/migrations.go
```diff
@@ -22,3 +25,39 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
 	}
 	return nil
 }
+
+func (m Migrator) Migrate3to4(ctx sdk.Context) error {
+	planIndices := m.keeper.GetAllPlanIndices(ctx)
+
+	for _, planIndex := range planIndices {
+		var oldPlanStruct v3.Plan
```
orenl-lava: instead of "old" and "new", use the version, e.g. `var plan_v3 v3.Plan`
(also below)

x/plans/keeper/migrations.go
```diff
@@ -22,3 +25,39 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
...
+			Price:                    oldPlanStruct.Price,
+			OveruseRate:              oldPlanStruct.OveruseRate,
+			AllowOveruse:             oldPlanStruct.AllowOveruse,
+			Description:              oldPlanStruct.Description,
+			Type:                     oldPlanStruct.Type,
+			AnnualDiscountPercentage: oldPlanStruct.AnnualDiscountPercentage,
+			PlanPolicy:               planPolicy,
+		}
+
+		err := m.keeper.plansFS.ModifyEntry(ctx, planIndex, uint64(ctx.BlockHeight()), &newPlanStruct)
```
orenl-lava: ModifyEntry() should be given the *specific* block of the entry, which we should now (see above).

x/plans/module.go
```diff
@@ -140,6 +140,11 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	if err := cfg.RegisterMigration(types.ModuleName, 2, migrator.Migrate2to3); err != nil {
 		panic(fmt.Errorf("%s: failed to register migration to v3: %w", types.ModuleName, err))
 	}
+
+	// register v3 -> v4 migration
```
orenl-lava, line 144: Not need when we keep v2 -> v3 migration only.

x/plans/module.go
```diff
@@ -164,7 +169,7 @@ func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.Raw
 }
 
 // ConsensusVersion implements ConsensusVersion.
-func (AppModule) ConsensusVersion() uint64 { return 3 }
+func (AppModule) ConsensusVersion() uint64 { return 4 }
```
orenl-lava, line 172: not needed (see first comment)

x/projects/keeper/migrations.go
```diff
@@ -25,3 +27,71 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
 	}
 	return nil
 }
+
+// Migrate3to4 implements protobuf migration from v3 to v4:
+// Now the project protobuf has admin and subscription policies
+func (m Migrator) Migrate3to4(ctx sdk.Context) error {
```
orenl-lava, line 33: Same comments as in `x/plans/`:
1. The migration from 2 to 3 didn't happen yet, so let's combine this into the exiting/upcoming Migrate2to3().
2. We decided to put actual migration code in v3/... subdir if possible. In this case, following x/bank/... example, introduce x/plans/migrations/v3/store.go:MigrateStore(...) and call it from here. Because the migration code needs access to keeper.plansFS, maybe we can add an API in x/plans/keeper/migrations.go to expose it, with a clear comment stating the API should only be used in migrations.

x/projects/keeper/migrations.go
```diff
@@ -25,3 +27,71 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
...
+	projectIndices := m.keeper.projectsFS.GetAllEntryIndices(ctx)
+	for _, projectIndex := range projectIndices {
+		blocks := m.keeper.projectsFS.GetAllEntryVersions(ctx, projectIndex, true)
+		for _, block := range blocks {
+			var oldProjectStruct v3.Project
```
orenl-lava: Replace "old" and "new" with explicit versions, e.g. `var project_v3 v3.Project` (everywhere)

x/projects/keeper/migrations.go
```diff
@@ -25,3 +27,71 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
...
+			if found := m.keeper.projectsFS.FindEntry(ctx, projectIndex, block, &oldProjectStruct); !found {
```
orenl-lava, line 39: Use `ReadEntry()` when we know the exact block version.

x/projects/keeper/migrations.go
```diff
@@ -25,3 +27,71 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
...
+			newProjectKeys := []types.ProjectKey{}
+			for _, oldProjectKey := range oldProjectStruct.ProjectKeys {
+				newProjectKey := types.ProjectKey{
+					Key:   oldProjectKey.Key,
+					Vrfpk: oldProjectKey.Vrfpk,
+				}
+
+				for _, oldprojectKeyType := range oldProjectKey.Types {
+					if oldprojectKeyType == v3.ProjectKey_ADMIN {
+						newProjectKey.Types = append(newProjectKey.Types, types.ProjectKey_ADMIN)
```
orenl-lava, line 53: Why the switch back from bitmap (of key types) to inefficient list?  (@Yaroms)

x/projects/module.go
```diff
@@ -141,6 +141,11 @@ func (am AppModule) RegisterServices(cfg module.Configurator) {
 	if err := cfg.RegisterMigration(types.ModuleName, 2, migrator.Migrate2to3); err != nil {
 		panic(fmt.Errorf("%s: failed to register migration to v3: %w", types.ModuleName, err))
 	}
+
+	// register v3 -> v4 migration
```
orenl-lava, line 145: No need if we combine into existing/upcoming v2 -> v3

x/projects/module.go
```diff
@@ -165,7 +170,7 @@ func (am AppModule) ExportGenesis(ctx sdk.Context, cdc codec.JSONCodec) json.Raw
 }
 
 // ConsensusVersion implements ConsensusVersion.
-func (AppModule) ConsensusVersion() uint64 { return 3 }
+func (AppModule) ConsensusVersion() uint64 { return 4 }
```
orenl-lava, line 173: Not needed if we combine into existing/upcoming v2 -> v3

x/plans/keeper/migrations.go
```diff
@@ -26,37 +26,44 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
...
 
 	for _, planIndex := range planIndices {
-		var oldPlanStruct v3.Plan
-		if found := m.keeper.plansFS.FindEntry(ctx, planIndex, uint64(ctx.BlockHeight()), &oldPlanStruct); !found {
-			return fmt.Errorf("could not find plan with index %s", planIndex)
-		}
+		blocks := m.keeper.plansFS.GetAllEntryVersions(ctx, planIndex, true)
+		for _, block := range blocks {
+			var oldPlanStruct v3.Plan
+			if found := m.keeper.plansFS.FindEntry(ctx, planIndex, block, &oldPlanStruct); !found {
```
orenl-lava, line 38: Use ReadEntry() when we know the specific block version

x/plans/keeper/migrations.go
```diff
@@ -34,7 +34,7 @@ func (m Migrator) Migrate3to4(ctx sdk.Context) error {
 	for _, planIndex := range planIndices {
 		blocks := m.keeper.plansFS.GetAllEntryVersions(ctx, planIndex, true)
 		for _, block := range blocks {
-			var oldPlanStruct v3.Plan
+			var oldPlanStruct v3.PlanV3
```
orenl-lava, line 37: replace "old" / "new" with explicit version number.

x/projects/keeper/migrations.go
```diff
@@ -93,5 +93,24 @@ func (m Migrator) Migrate3to4(ctx sdk.Context) error {
 			}
 		}
 	}
+
+	developerDataIndices := m.keeper.developerKeysFS.GetAllEntryIndices(ctx)
+	for _, developerDataIndex := range developerDataIndices {
+		blocks := m.keeper.developerKeysFS.GetAllEntryVersions(ctx, developerDataIndex, true)
+		for _, block := range blocks {
+			var oldDeveloperDataStruct v3.ProtoDeveloperDataV3
+			m.keeper.developerKeysFS.FindEntry(ctx, developerDataIndex, block, &oldDeveloperDataStruct)
```
orenl-lava, line 102: Use ReadEntry() when the specific block version is known.

x/plans/keeper/migrations.go
```diff
@@ -4,6 +4,9 @@ import (
 	"fmt"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	v3 "github.com/lavanet/lava/x/plans/migrations/v3"
```
orenl-lava, line 7: I think you meant "v4" here (the subdir names the target version); 
But we should combine this anyhow into v2 -> v3 migration, so this ends up being ok :)

x/projects/keeper/migrations.go
```diff
@@ -4,6 +4,8 @@ import (
 	"fmt"
 
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	v3 "github.com/lavanet/lava/x/projects/migrations/v3"
```
orenl-lava, line 7: I think you meant "v4" here (the subdir names the target version);
But we should combine this anyhow into v2 -> v3 migration, so this ends up being ok :)

orenl-lava (changes requested):

x/projects/keeper/migrations.go
Yaroms: bitmap was never implemented

x/plans/keeper/migrations.go
oren-lava: done

x/plans/keeper/migrations.go
oren-lava: done

x/plans/module.go
oren-lava: done

x/projects/keeper/migrations.go
oren-lava: done

x/projects/keeper/migrations.go
oren-lava: done

x/projects/module.go
oren-lava: done

x/projects/module.go
oren-lava: done

x/plans/keeper/migrations.go
oren-lava: done

x/plans/keeper/migrations.go
Yaroms: for this migration it is not really needed, also we might want to create such function

x/projects/keeper/migrations.go
oren-lava: done

x/projects/keeper/migrations.go
Yaroms: same

x/plans/keeper/migrations.go
oren-lava: I opened a subdir for **old** protobuf versions. So I changed this to v2

x/plans/keeper/migrations.go
orenl-lava: We decided to do that in the fixation migration code (which is a lib itself, not a module with keeper), and move the migration to separate subdirs if the effort is reasonable. Hence my suggestion to expose keeper.plansFS.

x/plans/keeper/migrations.go
orenl-lava: Already created in CNS-385, but not merged yet.
I will make this change part of CNS-385.

PR #461, opened 2023-05-02
PRT- masking provider address when context deadline hits

ranlavanet:


protocol/chainlib/jsonRPC.go
```diff
@@ -423,6 +424,10 @@ func (cp *JrpcChainProxy) SendNodeMsg(ctx context.Context, ch chan interface{},
 	// the error check here would only wrap errors not from the rpc
 	if err != nil {
 		utils.LavaFormatDebug("received an error from SendNodeMsg", utils.Attribute{Key: "GUID", Value: ctx}, utils.Attribute{Key: "error", Value: err})
+		if strings.Contains(err.Error(), context.DeadlineExceeded.Error()) {
```
omerlavanet, line 427: string contains is not the way to check if there was a context deadline you have connectCtx.Error():
ctx.Err() == context.DeadlineExceeded

protocol/chainlib/grpc.go
```diff
@@ -341,6 +341,10 @@ func (cp *GrpcChainProxy) SendNodeMsg(ctx context.Context, ch chan interface{},
 	response := msgFactory.NewMessage(methodDescriptor.GetOutputType())
 	err = conn.Invoke(connectCtx, "/"+nodeMessage.Path, msg, response)
 	if err != nil {
+		if strings.Contains(err.Error(), context.DeadlineExceeded.Error()) {
```
omerlavanet, line 344: string contains is not the way to check if there was a context deadline you have connectCtx.Error():
ctx.Err() == context.DeadlineExceeded

omerlavanet (changes requested):

omerlavanet (approved):

PR #463, opened 2023-05-03
PRT 639 - Add gRPC spec part for Canto, Axelar & Evmos

candostyavuz:
* Automation scripts are updated to work with inheritance mechanism in specs.
* Generated gRPC parts are added to Canto, Axelar & Evmos specs


cookbook/specs/spec_add_axelar.json
```diff
@@ -1104,6 +1104,846 @@
...
+                                    "stateful": 0
+                                },
+                                "interface": "grpc",
+                                "type": "",
+                                "extra_compute_units": "0"
+                            }
+                        ]
+                    },
+                    {
+                        "name": "cosmos.bank.v1beta1.Query/Balance",
```
ranlavanet, line 1805: something here seems wrong I've seen both cosmos and IBC queries while we have them in 
spec_add_ibc.json + spec_add_cosmossdk.json 
so need to see why and improve the automation

ranlavanet (changes requested):

cookbook/specs/spec_add_axelar.json
candostyavuz: fixed: f4a57b9110b1fe63cfc51e8c5bb431538d2f2404

ranlavanet (approved):

PR #466, opened 2023-05-04
CNS-66: Update CU's for EVM family 

oren-lava:
wrote a python script to update ETH-based chains' CU and updated the specs. Note that I've mistakenly did it on CNS-400 design task instead of the actual CNS-66 task @SeanZoR


oren-lava:
after the script was applied, it printed the attached report.
[update_specs_cu_report.txt.txt](https://github.com/lavanet/lava/files/11399157/update_specs_cu_report.txt.txt)

Yaroms (approved):

PR #469, opened 2023-05-07
CNS-261: enforce lowercase in spec names

oren-lava:
added check in spec proposal handler that spec name can only have lowercase chars or ' '.
After talking with @SeanZoR , existing specs should already have all lowercase chain names.


Yaroms (approved):

gilxgil:
should we also limit to readable ascii only?

PR #475, opened 2023-05-09
PRT-656 add provider metrics

omerlavanet:


protocol/metrics/metrics_provider_manager.go
```diff
@@ -0,0 +1,137 @@
...
+		consumerQoSMetric:         consumerQoSMetric,
+		blockMetric:               blockMetric,
+	}
+}
+
+func (pme *ProviderMetricsManager) getProviderMetric(specID string, apiInterface string) *ProviderMetrics {
+	return pme.providerMetrics[specID+apiInterface]
+}
+
+func (pme *ProviderMetricsManager) setProviderMetric(metrics *ProviderMetrics) {
```
ranlavanet, line 93: add a comment these methods must be called when locked

protocol/metrics/metrics_provider_manager.go
```diff
@@ -0,0 +1,137 @@
...
+		totalCUServicedMetric:     totalCUServicedMetric,
+		totalCUPaidMetric:         totalCUPaidMetric,
+		totalRelaysServicedMetric: totalRelaysServicedMetric,
+		totalErroredMetric:        totalErroredMetric,
+		consumerQoSMetric:         consumerQoSMetric,
+		blockMetric:               blockMetric,
+	}
+}
+
+func (pme *ProviderMetricsManager) getProviderMetric(specID string, apiInterface string) *ProviderMetrics {
```
ranlavanet, line 89: add a comment these methods must be called when locked

protocol/metrics/metrics_provider_manager.go
```diff
@@ -0,0 +1,137 @@
...
+		return
+	}
+	availableAPIInterface := []string{
+		spectypes.APIInterfaceJsonRPC,
+		spectypes.APIInterfaceTendermintRPC,
+		spectypes.APIInterfaceRest,
+		spectypes.APIInterfaceGrpc,
+	}
+	for _, apiInterface := range availableAPIInterface {
+		providerMetrics := pme.getProviderMetric(specID, apiInterface)
```
ranlavanet, line 132: not locked here.

protocol/metrics/metrics_provider_manager.go
ranlavanet: IMO best practice to lock inside

protocol/metrics/metrics_provider_manager.go
ranlavanet: IMO best practice to lock inside

ranlavanet (changes requested):

ranlavanet (approved):

PR #478, opened 2023-05-10
Prt-624: Validate client.toml is taken into account when running the rpc provider consumer

candostyavuz:
* Chain Id fetching hierarchy implemented
* clientCtx is created with configured chain id


protocol/rpcconsumer/rpcconsumer.go
```diff
@@ -247,10 +248,25 @@ rpcconsumer 127.0.0.1:3333 COS3 tendermintrpc 127.0.0.1:3334 COS3 rest <flags>`,
 			}
 			// handle flags, pass necessary fields
 			ctx := context.Background()
-			networkChainId, err := cmd.Flags().GetString(flags.FlagChainID)
+			var networkChainId string
+			cmdChainIdFlag := viper.GetString(flags.FlagChainID)
+			clientTomlConfig, err := config.ReadFromClientConfig(clientCtx)
 			if err != nil {
 				return err
```
ranlavanet, line 255: lets not kill the process if this method failed. just print an error and continue without it

ranlavanet: you can basically add the functionality inside 

if err == nil {
    functionality 
}

also we want to read config only if chainId was not specified so 

if cmdChainIdFlag == app.Name {
   clientTomlConfig, err := config.ReadFromClientConfig(clientCtx)
   if err == nil {
       functionality 
   }
}

protocol/rpcconsumer/rpcconsumer.go
```diff
@@ -247,10 +248,25 @@ rpcconsumer 127.0.0.1:3333 COS3 tendermintrpc 127.0.0.1:3334 COS3 rest <flags>`,
...
+				networkChainId = cmdChainIdFlag
+			} else {
+				if clientTomlConfig.ChainID != "" {
+					networkChainId = clientTomlConfig.ChainID
+				} else {
+					networkChainId = app.Name
+				}
+			}
+			if networkChainId != app.Name {
+				utils.LavaFormatInfo("ChainID is different from " + app.Name)
```
ranlavanet, line 267: lets remove it as our app.name is lava and testnet is lava-testnet-1 so it will push this always 
instead we can print the chain id that was set 

utils.LavaFormatInfo("Running with chain-id:" + chainId)

protocol/rpcconsumer/rpcconsumer.go
```diff
@@ -247,10 +248,25 @@ rpcconsumer 127.0.0.1:3333 COS3 tendermintrpc 127.0.0.1:3334 COS3 rest <flags>`,
...
+			}
+
 			logLevel, err := cmd.Flags().GetString(flags.FlagLogLevel)
 			if err != nil {
```
ranlavanet, line 271: Please apply all changes to rpc provider as well.

ranlavanet (changes requested):

protocol/rpcconsumer/rpcconsumer.go
candostyavuz: fixed: [17dc045](https://github.com/lavanet/lava/pull/478/commits/17dc0459de269a9e434c5f3411692966a7e9525a)

protocol/rpcconsumer/rpcconsumer.go
candostyavuz: fixed: [106ac4b](https://github.com/lavanet/lava/pull/478/commits/106ac4b912d522d204c87c21d03b94417c521858)

ranlavanet (approved):

PR #482, opened 2023-05-10
Adding spec updater and event tracker functionality

ranlavanet:


protocol/rpcconsumer/rpcconsumer.go
```diff
@@ -134,11 +128,21 @@ func (rpcc *RPCConsumer) Start(ctx context.Context, txFactory tx.Factory, client
...
 			}
+
+			// Register For Updates
 			consumerSessionManager := lavasession.NewConsumerSessionManager(rpcEndpoint, optimizer)
 			rpcc.consumerStateTracker.RegisterConsumerSessionManagerForPairingUpdates(ctx, consumerSessionManager)
 
 			finalizationConsensus := &lavaprotocol.FinalizationConsensus{}
 			consumerStateTracker.RegisterFinalizationConsensusForUpdates(ctx, finalizationConsensus)
+
+			err = rpcc.consumerStateTracker.RegisterForSpecUpdates(ctx, chainParser, chainID, rpcEndpoint.ApiInterface)
```
omerlavanet, line 139: we can't do this here, registering for spec updates needs to be before we send the chainParser as an argument to other classes like the optimizer, this will crash on null spec

protocol/rpcconsumer/rpcconsumer.go
```diff
@@ -99,12 +99,6 @@ func (rpcc *RPCConsumer) Start(ctx context.Context, txFactory tx.Factory, client
 				return err
 			}
 			chainID := rpcEndpoint.ChainID
-			err = consumerStateTracker.RegisterChainParserForSpecUpdates(ctx, chainParser, chainID)
```
omerlavanet, line 102: why did you move this, this is bad, it has to come before  using the class

protocol/statetracker/spec_updater.go
```diff
@@ -0,0 +1,65 @@
...
+	CallbackKeyForSpecUpdate = "spec-update"
+)
+
+type SpecGetter interface {
+	GetSpec(ctx context.Context, chainID string) (*spectypes.Spec, error)
+}
+
+type SpecUpdater struct {
+	eventTracker     *EventTracker
+	chainParser      chainlib.ChainParser
```
omerlavanet, line 21: remove chainParser reference, use an interface (you only need SetSpec i think), interface can be called SpecUpdatable, and then we can call it once for all api interfaces, the key and arguments shouldn't get apiInterface

protocol/statetracker/spec_updater.go
```diff
@@ -0,0 +1,65 @@
...
+
+func NewSpecUpdater(chainId string, chainParser chainlib.ChainParser, specGetter SpecGetter, eventTracker *EventTracker, apiInterface string) *SpecUpdater {
+	return &SpecUpdater{chainId: chainId, chainParser: chainParser, specGetter: specGetter, eventTracker: eventTracker, apiInterface: apiInterface}
+}
+
+func (su *SpecUpdater) UpdaterKey() string {
+	utils.LavaFormatDebug("UpdaterKey: " + CallbackKeyForSpecUpdate + su.chainId + su.apiInterface)
+	return CallbackKeyForSpecUpdate + su.chainId + su.apiInterface
+}
+
```
omerlavanet, line 36: missing RegisterForSpecUpdates, if you support more than one, you need this
also you need locks for it like in the other updaters in the fix PR

protocol/statetracker/spec_updater.go
```diff
@@ -0,0 +1,65 @@
...
+func (su *SpecUpdater) Update(latestBlock int64) {
+	specUpdated := su.eventTracker.getLatestSpecModifyEvents()
```
omerlavanet, line 38: use the lock as read here

protocol/statetracker/spec_updater.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+}
+
+func (su *SpecUpdater) setSpec(spec *spectypes.Spec) {
+	// set blockLastUpdated
+	su.blockLastUpdated = spec.BlockLastUpdated
+	su.chainParser.SetSpec(*spec)
+}
+
+func (su *SpecUpdater) InitSpec(ctx context.Context) error {
```
omerlavanet, line 58: we don't need this function, when we use RegisterForSpecUpdates we can set the initial spec

protocol/statetracker/provider_state_tracker.go
```diff
@@ -44,17 +44,19 @@ func (pst *ProviderStateTracker) RegisterForEpochUpdates(ctx context.Context, ep
...
+	// register for spec updates when a spec has been modified
+	specUpdater := NewSpecUpdater(chainId, chainParser, pst.stateQuery, pst.eventTracker, apiInterface)
+	specUpdaterRaw := pst.StateTracker.RegisterForUpdates(ctx, specUpdater)
+	specUpdater, ok := specUpdaterRaw.(*SpecUpdater)
+	if !ok {
+		utils.LavaFormatFatal("invalid updater type returned from RegisterForUpdates", nil, utils.Attribute{Key: "updater", Value: specUpdaterRaw})
 	}
-	chainParser.SetSpec(*spec)
-	return nil
+	return specUpdater.InitSpec(ctx)
```
omerlavanet, line 55: use same format as other updaters like: voteUpdater.RegisterVoteUpdatable(ctx, &voteUpdatable, endpoint)

protocol/statetracker/provider_state_tracker.go
```diff
@@ -44,17 +44,19 @@ func (pst *ProviderStateTracker) RegisterForEpochUpdates(ctx context.Context, ep
 	epochUpdater.RegisterEpochUpdatable(ctx, epochUpdatable)
 }
 
-func (pst *ProviderStateTracker) RegisterChainParserForSpecUpdates(ctx context.Context, chainParser chainlib.ChainParser, chainID string) error {
-	spec, err := pst.stateQuery.GetSpec(ctx, chainID)
-	if err != nil {
-		return err
+func (pst *ProviderStateTracker) RegisterForSpecUpdates(ctx context.Context, chainParser chainlib.ChainParser, chainId string, apiInterface string) error {
+	// register for spec updates when a spec has been modified
+	specUpdater := NewSpecUpdater(chainId, chainParser, pst.stateQuery, pst.eventTracker, apiInterface)
```
omerlavanet, line 49: after consolidating all api interfaces to one specUpdater remove the apiInterface

protocol/statetracker/payment_updater.go
```diff
@@ -17,10 +17,11 @@ type PaymentUpdatable interface {
 type PaymentUpdater struct {
 	paymentUpdatable map[string]*PaymentUpdatable
 	stateQuery       *ProviderStateQuery
```
omerlavanet, line 19: so we can remove stateQuery

protocol/statetracker/consumer_state_tracker.go
```diff
@@ -48,6 +48,17 @@ func (cst *ConsumerStateTracker) RegisterConsumerSessionManagerForPairingUpdates
...
 
+func (cst *ConsumerStateTracker) RegisterForSpecUpdates(ctx context.Context, chainParser chainlib.ChainParser, chainId string, apiInterface string) error {
+	// register for spec updates when a spec has been modified
+	specUpdater := NewSpecUpdater(chainId, chainParser, cst.stateQuery, cst.eventTracker, apiInterface)
+	specUpdaterRaw := cst.StateTracker.RegisterForUpdates(ctx, specUpdater)
+	specUpdater, ok := specUpdaterRaw.(*SpecUpdater)
+	if !ok {
+		utils.LavaFormatFatal("invalid updater type returned from RegisterForUpdates", nil, utils.Attribute{Key: "updater", Value: specUpdaterRaw})
+	}
+	return specUpdater.InitSpec(ctx)
```
omerlavanet, line 59: since this is the same code in the provider state tracker can we move this to the base state tracker?

protocol/statetracker/event_tracker.go
omerlavanet, line 1: we need to add a task to make this class an interface for others, getting callbacks and move the functionality of parsing the objects into the classes that use them, no reason that each event we add has the full function here

protocol/statetracker/event_tracker.go
```diff
@@ -0,0 +1,114 @@
...
+	et.latestUpdatedBlock = latestBlock
+	et.blockResults = blockResults
+	return nil
+}
+
+func (et *EventTracker) getLatestPaymentEvents() (payments []*rewardserver.PaymentRequest, err error) {
+	et.lock.RLock()
+	defer et.lock.RUnlock()
+	transactionResults := et.blockResults.TxsResults
+	for _, tx := range transactionResults {
```
omerlavanet, line 42: what i would like to see is one loop on transaction results, have a map of event.Type to callback, and call the callback and that's it, on initialization when the updaters get the eventTracker, they call setCallbackForEvent and that's it, more efficient and cleaner class structure

omerlavanet (changes requested):

protocol/rpcconsumer/rpcconsumer.go
ranlavanet: Nice

omerlavanet (approved):

PR #486, opened 2023-05-14
CNS-428 fixation store to support DelEntry()

orenl-lava:
An entry can be deleted using DelEntry(), after which is will not possible to GetEntry(), and calls to FindEntry() for a block beyond that at time of deletion (or later) would fail. Until the data has been fully cleaned up (i.e. no refcount and beyond all stale-periods), calls to AppendEntry() with that entry's index will fail.


common/fixation_entry.go
```diff
@@ -269,10 +330,9 @@ func (fs *FixationStore) getUnmarshaledEntryForBlock(ctx sdk.Context, safeIndex
 		fs.cdc.MustUnmarshal(iterator.Value(), &entry)
 
 		if entry.Block <= block {
-			if entry.IsStale(ctx) {
+			if entry.IsStale(ctx) && !entry.Deleted {
```
Yaroms, line 333: can you explain why?
this means : dont show the entry if it is stale and not deleted
why we want to show it if it is stale and deleted?

common/fixation_entry.go
```diff
@@ -348,14 +408,113 @@ func (fs *FixationStore) PutEntry(ctx sdk.Context, index string, block uint64) {
...
+	block := uint64(ctx.BlockHeight()) + 1
+	if entry.Block >= block {
+		return sdkerrors.ErrInvalidHeight.Wrapf("unable to delete entry with future block: %d", block)
+	}
+
+	// add a marker version that is both deleted and stale to indicate this
+	// entry is deleted: "deleted" is needed to reject future GetEntry() calls;
+	// and "stale" is needed to reject future FindEntry() calls.
+
+	deleted := types.Entry{
```
Yaroms, line 434: why not just mark the existing entry with deleted instead of adding a new entry?

common/fixation_entry.go
orenl-lava: because AppendEntry() expects to always dig the latest entry. and if that latest entry is stale (because it's was deleted enough blocks ago), then it may get a non-latest entry and its logic will break (e.g. it assume not-found means no entry versions for that index exists, which may not be the case).

common/fixation_entry.go
```diff
@@ -348,6 +408,47 @@ func (fs *FixationStore) PutEntry(ctx sdk.Context, index string, block uint64) {
 	fs.putEntry(ctx, entry)
 }
 
+// DelEntry deletes marks the entry for deletion so that future GetEntry will
+// not find it anymore. (Cleanup of the EntryIndex will be done when the last
+// reference is removed).
+func (fs *FixationStore) DelEntry(ctx sdk.Context, index string) error {
+	safeIndex, err := types.SanitizeIndex(index)
+	if err != nil {
+		return sdkerrors.ErrNotFound.Wrapf("invalid non-ascii index")
```
oren-lava, line 417: provide additional details to the error? like the index or a more detailed description (specifically which function failed)

oren-lava (changes requested):

common/fixation_entry.go
orenl-lava: this is a library function. the error clearly indicates the failure: its (single) argument was invalid.
the caller may/should add details as it sees fit.

common/types/fixationEntry.go
```diff
@@ -1,20 +1,41 @@
...
+		if entry.StaleAt <= block {
 			return true
 		}
 	}
 	return false
 }
+
+// IsDeletedBy tests whether an entry is deleted, with respect to a given
+// block, i.e. has entry.DeletAt smaller or equal to that that block.
+func (entry Entry) IsDeletedBy(block uint64) bool {
```
Yaroms, line 22: confusing name, why not IsDeletedAt

common/types/fixationEntry.go
```diff
@@ -1,20 +1,41 @@
...
+// IsDeleted tests whether an entry is currently deleted, i.e. has its
+// entry.DeleteAt below the current ctx block height.
+func (entry Entry) IsDeleted(ctx sdk.Context) bool {
+	return entry.IsDeletedBy(uint64(ctx.BlockHeight()))
+}
+
+// IsMarkedDeleted tests whether an entry is marked for deletion, i.e. has
+// entry.Deleted != math.MaxUint64.
+func (entry Entry) HasDeleteAt() bool {
+	return entry.DeleteAt < math.MaxUint64
```
Yaroms, line 35: we should use a constant here

common/types/fixationEntry.go
```diff
@@ -1,20 +1,41 @@
...
+}
+
+// IsEntryIndexLive tests whether an entry-index is live
+func IsEntryIndexLive(value []byte) bool {
+	return value[0] == EntryIndexLive[0]
```
Yaroms, line 40: index 0? const

common/types/fixationEntry.go
```diff
@@ -1,20 +1,41 @@
...
 )
 
-func FixationVersion() uint64 {
-	return 3
-}
-
 // IsStale tests whether an entry is stale, i.e. has refcount zero _and_
 // has passed its stale_at time (more than STALE_ENTRY_TIME since deletion).
-func (entry Entry) IsStale(ctx sdk.Context) bool {
+func (entry Entry) IsStaleBy(block uint64) bool {
```
Yaroms, line 11: by who?
call it IsStaleAt?

common/fixation_entry_index.go
```diff
@@ -26,11 +30,14 @@ func (fs *FixationStore) getEntryIndexStore(ctx sdk.Context) *prefix.Store {
...
 // setEntryIndex stores an Entry index in the store
-func (fs FixationStore) setEntryIndex(ctx sdk.Context, safeIndex string) {
+func (fs FixationStore) setEntryIndex(ctx sdk.Context, safeIndex string, live bool) {
 	types.AssertSanitizedIndex(safeIndex, fs.prefix)
 	store := fs.getEntryIndexStore(ctx)
-	appendedValue := []byte(safeIndex) // convert the index value to a byte array
-	store.Set(types.KeyPrefix(safeIndex), appendedValue)
+	value := types.EntryIndexLive
+	if !live {
+		value = types.EntryIndexDead
```
Yaroms, line 38: why do we need to save dead entry? i dont see any use for it, we always fetch the live ones

common/fixation_entry.go
```diff
@@ -115,24 +201,37 @@ func (fs *FixationStore) AppendEntry(
...
+	blockHeight := uint64(ctx.BlockHeight())
+
+	if block < blockHeight {
+		panic(fmt.Sprintf("AppendEntry for block %d < current ctx block %d", block, blockHeight))
+	}
+
+	// find latest entry, including possible future entries
 	latestEntry, found := fs.getUnmarshaledEntryForBlock(ctx, safeIndex, block)
 
+	var deleteAt uint64 = math.MaxUint64
```
Yaroms, line 218: lets use a constant

common/fixation_entry.go
```diff
@@ -142,38 +241,149 @@ func (fs *FixationStore) AppendEntry(
...
+		// DeleteAt to the new entry, drop the old timer, and start a new timer. (note:
+		// deletion, if any, cannot be for the current block, because it would have been
+		// processed at the beginning of the block (and AppendEntry would fail earlier).
+
+		if block == blockHeight {
+			deleteAt = latestEntry.DeleteAt
+			latestEntry.DeleteAt = math.MaxUint64
+			fs.putEntry(ctx, latestEntry)
+		} else {
+			key := encodeForTimer(safeIndex, block, timerFutureEntry)
```
Yaroms, line 257: why do we need a timer for block > blockheight?

common/fixation_entry.go
```diff
@@ -142,38 +241,149 @@ func (fs *FixationStore) AppendEntry(
...
+		fs.deleteStaleEntries(ctx, safeIndex, block)
+	}
+}
+
+func (fs *FixationStore) updateFutureEntry(ctx sdk.Context, safeIndex string, block uint64) {
+	if block != uint64(ctx.BlockHeight()) {
+		panic(fmt.Sprintf("Future entry: future block %d != current block %d", block, ctx.BlockHeight()))
+	}
+
+	latestEntry, found := fs.getUnmarshaledEntryForBlock(ctx, safeIndex, block-1)
```
Yaroms, line 303: why is it for block-1?

common/fixation_entry.go
```diff
@@ -142,38 +241,149 @@ func (fs *FixationStore) AppendEntry(
...
+		// latest entry had extra refcount for being the latest; so drop that refcount
+		// because from now on it is no longer so.
+
+		fs.putEntry(ctx, latestEntry)
+	}
+}
+
+func (fs *FixationStore) deleteMarkedEntry(ctx sdk.Context, safeIndex string, block uint64) {
+	entry := fs.getEntry(ctx, safeIndex, block)
+	blockHeight := uint64(ctx.BlockHeight())
```
Yaroms, line 338: we need better names, block and blockheight are very confusing

common/fixation_entry.go
```diff
@@ -142,38 +241,149 @@ func (fs *FixationStore) AppendEntry(
...
+	for _, entry := range entriesToRemove {
+		key := encodeForTimer(entry.Index, entry.Block, timerFutureEntry)
+		fs.tstore.DelTimerByBlockHeight(ctx, entry.Block, key)
+		fs.removeEntry(ctx, entry.Index, entry.Block)
+	}
+}
+
+func (fs *FixationStore) deleteStaleEntries(ctx sdk.Context, safeIndex string, _ uint64) {
 	store := fs.getEntryStore(ctx, safeIndex)
+	block := uint64(ctx.BlockHeight())
```
Yaroms, line 376: current blockheight member name is different for every methodm lets call it currentBlockHeight in all the methods

common/fixation_entry.go
```diff
@@ -268,10 +494,18 @@ func (fs *FixationStore) getUnmarshaledEntryForBlock(ctx sdk.Context, safeIndex
...
+			// stale entries work as markers of the extent to where the preceding
+			// (non stale) entry is valid. So if we meet one, we bail.
+			// however if that stale entry is also deleted (which means, it is the
+			// latest too) then we need to not bail, and do return that entry, to
+			// comply with the expecations of AppendEntry(). thus, we may return an
+			// entry that is both stale and deleted.
+			// for this reason, we explicitly test for a deleted entry in EntryGet()
+			// and AppendEntry(), and for stale entry in EntryFind(). so EntryGet()
+			// and EntryFind() would return not-found, and AppendEntry would fail.
+			if entry.IsStaleBy(blockHeight) && !entry.IsDeletedBy(blockHeight) {
```
Yaroms, line 506: does this mean we can't use FindEntry to fetch a stale entry?
we want to be able to find stale entries, that is the point of keeping them in the state

Yaroms (changes requested):
nice :)

common/fixation_entry_index.go
orenl-lava: it is a marker that there are still referenced (or about to be stale) entries of this index.
it's useful e.g. for migrations, to find all the entry versions (regardless of state) in the store.
this is also useful for testing and debugging.

common/types/fixationEntry.go
orenl-lava: see previous comment/reply

common/types/fixationEntry.go
orenl-lava: the value here is 1 byte.  but the KVstore deals with []byte, so the code uses the first one. always.
by the way, for existing entries - that is valid ascii, so cannot be 0, or 1, that's why we don't need migration. over time these entries will disappear...

common/fixation_entry.go
orenl-lava: `math.MaxUint64` is a/the constant.

Yaroms (approved):

omerlavanet (approved):

PR #499, opened 2023-05-18
CNS-428: timer-store support different entries with same block

orenl-lava:
Previously timers keys kept only the block number, so one could not set more than one timer for the same block. For example, when FixationStore set a second timer for a given block (for a different index), it would overwrite the first one.
     
Fix this by extending the keys to also include an index (string) given by the caller. For example, FixationStore uses the entry index.


common/timer.go
```diff
@@ -152,74 +185,114 @@ func (tstore *TimerStore) setNextTimeout(ctx sdk.Context, which types.TimerType,
...
+	store.Delete(timerKey)
 }
 
-func (tstore *TimerStore) AddTimerByBlockHeight(ctx sdk.Context, block uint64, data string) {
-	tstore.addTimer(ctx, types.BlockHeight, block, data)
+// AddTimerByBlockHeight adds a new timer to expire on a given block height.
+// If a timer for that <block, key> tuple exists, it will be overridden.
+func (tstore *TimerStore) AddTimerByBlockHeight(ctx sdk.Context, block uint64, key []byte, data []byte) {
+	if block <= uint64(ctx.BlockHeight()) {
+		panic(fmt.Sprintf("timer expiry block %d smaller than ctx block %d",
```
oren-lava, line 212: Should we panic? why not returning an error?
I'm asking because in my badges PR, the expiry block I put for the timer is influenced by the badge's epoch field. Thus, the value of the timer's block can be dependent on user input (I don't think we should panic as a result of user input)

common/timer.go
```diff
@@ -152,74 +185,114 @@ func (tstore *TimerStore) setNextTimeout(ctx sdk.Context, which types.TimerType,
...
+	tstore.addTimer(ctx, types.BlockHeight, block, key, data)
 }
 
-func (tstore *TimerStore) AddTimerByBlockTime(ctx sdk.Context, timestamp uint64, data string) {
-	tstore.addTimer(ctx, types.BlockTime, timestamp, data)
+// AddTimerByBlockTime adds a new timer to expire on a future block with the given timestamp.
+// If a timer for that <timestamp, key> tuple exists, it will be overridden.
+func (tstore *TimerStore) AddTimerByBlockTime(ctx sdk.Context, timestamp uint64, key []byte, data []byte) {
+	if timestamp <= uint64(ctx.BlockTime().UTC().Unix()) {
+		panic(fmt.Sprintf("timer expiry time %d smaller than ctx time %d",
```
oren-lava, line 222: same comment as the last one

common/timer.go
```diff
@@ -152,74 +185,114 @@ func (tstore *TimerStore) setNextTimeout(ctx sdk.Context, which types.TimerType,
...
+			timestamp, uint64(ctx.BlockTime().UTC().Unix())))
+	}
+	tstore.addTimer(ctx, types.BlockTime, timestamp, key, data)
 }
 
-type timerTuple struct {
-	value uint64
-	data  string
+// DelTimerByBlockHeight removes an eximsting timer for the <block, key> tuple.
```
oren-lava, line 228: eximsting -> existing

common/timer.go
```diff
@@ -152,74 +185,114 @@ func (tstore *TimerStore) setNextTimeout(ctx sdk.Context, which types.TimerType,
...
+func (tstore *TimerStore) DelTimerByBlockHeight(ctx sdk.Context, block uint64, key []byte) {
+	tstore.delTimer(ctx, types.BlockHeight, block, key)
 }
 
-func (tstore *TimerStore) tickValue(ctx sdk.Context, which types.TimerType, tickValue uint64) {
-	nextValue := tstore.getNextTimeout(ctx, which)
-	if tickValue < nextValue {
-		return
+// DelTimerByBlockHeight removes an eximsting timer for the <timetamp, key> tuple.
```
oren-lava, line 233: same comment as above

common/timer.go
```diff
@@ -152,74 +185,114 @@ func (tstore *TimerStore) setNextTimeout(ctx sdk.Context, which types.TimerType,
...
+func (tstore *TimerStore) DelTimerByBlockTime(ctx sdk.Context, timestamp uint64, key []byte) {
+	tstore.delTimer(ctx, types.BlockTime, timestamp, key)
+}
+
+func (tstore *TimerStore) DumpTimers(ctx sdk.Context, which types.TimerType) string {
```
oren-lava, line 238: DumpTimers is a pretty good name, but maybe change it to PrintTimers (which is more clear)?

common/timer.go
```diff
@@ -152,74 +185,114 @@ func (tstore *TimerStore) setNextTimeout(ctx sdk.Context, which types.TimerType,
...
+
+	for ; iterator.Valid(); iterator.Next() {
+		value, key := types.DecodeBlockAndKey(iterator.Key())
+		b.WriteString(fmt.Sprintf("block: %d key %v\n", value, key))
 	}
 
+	return b.String()
+}
+
+func (tstore *TimerStore) getFrontTimer(ctx sdk.Context, which types.TimerType) (uint64, []byte, []byte) {
```
oren-lava, line 254: According to your example from the documentation above, an object can have multiple timers of different types. In your example, a package module's object has a blockHeight timer that expires on block 150 and three additional timers that expire on block 180. Also, it has a blockTime timer that never expires (maxUint64).

What is the hierarchy between them? According to getFrontTimer's implementation, you get the latest timer of a certain type. But since all the blockHeight timers will expire before the BlockTime timer expire, is it not correct returning them first? why do you differentiate between the timer types? you can convert the blockTime timers expiry field to blocks and have one list of timers.

oren-lava (changes requested):

common/timer.go
orenl-lava: It doesn't make sense to ask for a timeout in the past - it will never happen. Your code prune this case and not ask to add such timer. (note that the function does not return an error - it shouldn't).

common/timer.go
orenl-lava: see reply there.

common/timer.go
oren-lava: yeah, I check it before creating a timer

common/timer.go
Yaroms: i conversion of time to block and vise versa is a problematic area, i think we want to avoid that

Yaroms, line 254: lets name the returned valuse, it is not clear what they are

common/timer_migrate.go
```diff
@@ -35,3 +36,25 @@ func (tstore *TimerStore) MigrateVersion(ctx sdk.Context) (err error) {
...
+
+		iterator := sdk.KVStorePrefixIterator(store, []byte{})
+		defer iterator.Close()
+
+		for ; iterator.Valid(); iterator.Next() {
+			value, key := types.DecodeBlockAndKey(iterator.Key())
+			key_v1 := types.EncodeKey(value)
+			key_v2 := types.EncodeBlockAndKey(value, key)
+			store.Set(key_v2, iterator.Value())
+			store.Delete(key_v1)
```
Yaroms, line 55: why do we need key1? key and key1 are the same

Yaroms (changes requested):
good work :), did we test the migration?

common/timer.go
orenl-lava: it's a common convention (like core-dump). 
but can rename, sure.

common/timer.go
orenl-lava: > According to your example from the documentation above, an object can have multiple timers of different types. 

Actually, modules have timers, usually related to objects they manage ("objects" themselves do not have timers). Also, note: "multiple types" --> two types: either "block-height" or "block-time".

> In your example, a package module's object has a blockHeight timer that expires on block 150 and three additional timers that expire on block 180. Also, it has a blockTime timer that never expires (maxUint64).

There is no timer of type BlockTime, hence the "next" timeout there is infinity.
There are three timers of type BlockHeight: "first", "second", "third" (both keys and data). They differ in their key - each key usually "represents" a distinct object, as decided by the module. If the same key (and expiry block/time) is used for two different objects, then one's timer will overwrite the other's. 

> According to getFrontTimer's implementation, you get the latest timer of a certain type. 

Not the latest - but the earliest, i.e. the one coming up next.
Usually you can get it by looking at the record for the "next" timeout, but during the loop in `tickValue()` we can't reliably find the new "next" timeout (see comment there), so we need this function.

> But since all the blockHeight timers will expire before the BlockTime timer expire, is it not correct returning them first? 

The front - as in earliest - is returned. See the comment in the function.
There are no timers or type BlockTime in the example.
Anyway, the two types of timers are managed separately.

> why do you differentiate between the timer types? you can convert the blockTime timers expiry field to blocks and have one list of timers.

Not possible to convert the BlockTime expiry to blocks: BlockTime based expiry means "expire upon the first block whose timestamp is not earlier than the expiry time". We can't predict exactly how many blocks in the future it will happen.

Splitting into two types simplifies the code and allows to reuse most of the logic except for the calculation of expiry conditions.

orenl-lava: > i conversion of time to block and vise versa is a problematic area, i think we want to avoid that

why is it problematic?
the following definition is simple and deterministic: "time-based expiry means to expire upon the first block whose timestamp is not earlier than the expiry time".
we already do it with subscriptions.

common/timer_migrate.go
orenl-lava: `key` and `key_v1` are not the same.  instead, `key_v1` and `iterator.Key()` are the same.
the assignment in line 52 can use `iterator.Key()` instead of re-encoding; but I think the latter is clearer to show the contrast with `key_v2`. I'll add a comment.

Yaroms (approved):

orenl-lava:
turns out migration not needed here - so I reverted it.
it moved to #486

Yaroms (approved):

PR #505, opened 2023-05-20
Update start_node.sh

NodesLLS:
Added double quotes around variables to prevent issues with whitespace and special characters.
    Used source instead of . to load the setup_config.sh file.
    Used lowercase variable names for consistency.
    Improved readability with consistent indentation and spacing.
    Removed unnecessary debug function.
    Added -o flag in the unzip command to overwrite existing files.
    Improved comments for better clarity.
    Moved the shebang line to #!/bin/bash to ensure compatibility with Bash features.

### DIFF
--- a/docker/start_node.sh
+++ b/docker/start_node.sh
@@ -1,4 +1,4 @@
-#!/bin/sh
+#!/bin/bash
 # vim:sw=4:ts=4:et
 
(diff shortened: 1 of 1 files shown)


Yaroms:
not relevant anymore, thanks for the help!

PR #512, opened 2023-05-24
CNS-450 implemented getting subscription's projects

oren-lava:
This query gets a subscription address and returns all the subscription's projects

### DIFF
--- a/proto/subscription/query.proto
+++ b/proto/subscription/query.proto
@@ -21,6 +21,11 @@ service Query {
 		option (google.api.http).get = "/lavanet/lava/subscription/current/{consumer}";
 	}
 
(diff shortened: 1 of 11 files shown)


proto/subscription/query.proto
```diff
@@ -21,6 +21,11 @@ service Query {
 		option (google.api.http).get = "/lavanet/lava/subscription/current/{consumer}";
 	}
 
+// Queries a list of ListProjects items.
+	rpc ListProjects(QueryListProjectsRequest) returns (QueryListProjectsResponse) {
```
Yaroms, line 25: can we call it just projects? 
"lavad q subscription projects addr"

Yaroms (approved):

PR #530, opened 2023-06-01
CNS - spec add query info

omerlavanet:


proto/spec/query.proto
```diff
@@ -87,6 +87,7 @@ message ShowAllChainsInfoStruct {
 	string chainName = 1;
 	string chainID = 2;
 	repeated string enabledApiInterfaces = 3;
+  uint64 api_count = 4;
```
Yaroms, line 90: tab :)

x/spec/keeper/grpc_query_show_all_chains.go
```diff
@@ -26,14 +26,26 @@ func (k Keeper) ShowAllChains(goCtx context.Context, req *types.QueryShowAllChai
 		chainName := spec.GetName()
 		chainId := spec.GetIndex()
 
+		expectedInterfaces := make(map[string]bool)
```
Yaroms, line 29: why do we need the make here?

omerlavanet: Because the internal function expects a map and updates it

Yaroms: but it receives an empty map and than you override the existing one' why not just create the map inside the method?

Yaroms (approved):

PR #548, opened 2023-06-08
PRT-701-add-header-allow-listing

omerlavanet:


omerlavanet:
already included in a more advanced PR

PR #551, opened 2023-06-12
PRT-649 chain ID - sanity check on rpcprovider boot

Aleksao998:
## Description
The provider initialization process requires us to specify the chain IDs alongside the desired RPC endpoint for proper functioning. However, a potential problem arises when someone claims to be serving Polygon while utilizing an RPC endpoint from the Ethereum network. To address this issue, this pull request (PR) introduces a solution by implementing chain ID verification during the Bootstrap phase. By leveraging the RPC endpoint, we can retrieve the actual chain ID and compare it with the specified chain ID, ensuring accurate network configuration.

## How it works:
In each specification, a new parsing method is added to validate the chain ID. Here is an example of how it works
```
   {
      "function_template": "{\"jsonrpc\":\"2.0\",\"method\":\"eth_chainId\",\"params\":[],\"id\":1}",
      "function_tag": "getChainId",
      "result_parsing": {
          "parser_arg": [
              "0"
          ],
          "parser_func": "PARSE_BY_ARG",
          "encoding": "hex"
      },
      "wanted_result": "0x1",
      "api_name": "eth_chainId"
  }
```
In this example, we define the method that will retrieve the chain ID from the RPC, which is eth_chainId in this case. We also specify the expected result value, which is "0x1". 

If a chain does not have this specific parsing method defined, the rule will be ignored, and the user will receive a warning indicating the absence of validation for the chain ID.

## Manual tests:
1. I've manually tested all rpc endpoints with all api interfaces to make sure it works

## Additional info:
1. We can not verify Lava chain id, because we have different chain ids locally/staging/testnet
2. I couldn't find the chain id of Solana


protocol/parser/parser.go
```diff
@@ -107,6 +107,27 @@ func ParseBlockFromParams(rpcInput RPCInput, blockParser spectypes.BlockParser)
 	return rpcInput.ParseBlock(resString)
 }
 
+func ParseSpecIDFromReply(rpcInput RPCInput, blockParser spectypes.BlockParser) (string, error) {
```
ranlavanet, line 110: duplicated code in ParseBlockFromReply, you can call ParseSpecIDFromReply -> ParseFromReply and use it inside ParseBlockFromReply as they are basically the same except for the return statement

proto/spec/api_collection.proto
```diff
@@ -48,6 +48,7 @@ message Parsing {
   string function_template = 2;
   BlockParser result_parsing = 3 [(gogoproto.nullable) = false];
   string api_name = 4;
+  string wanted_result = 5;
```
ranlavanet, line 51: change wanted_result to expected_result 

2. fix merge conflict with function_tag, as it was changed to Enum in omer's branch

x/spec/types/types.go
```diff
@@ -35,10 +35,11 @@ const (
 const (
 	GET_BLOCKNUM                = "getBlockNumber"
 	GET_BLOCK_BY_NUM            = "getBlockByNumber"
+	GET_CHAIN_ID                = "getChainId"
```
ranlavanet, line 38: this changed to an enum in the latest version make sure to make adjustments with the 701 PR

ranlavanet (changes requested):

ranlavanet:
@Yaroms , FYI on this PR and the spec changes

proto/spec/api_collection.proto
Aleksao998: Fixed 4dc611b08e0ecf359f4bf505ce65013511829283

protocol/parser/parser.go
```diff
@@ -107,25 +107,34 @@ func ParseBlockFromParams(rpcInput RPCInput, blockParser spectypes.BlockParser)
...
 	if err != nil || result == nil {
-		return spectypes.NOT_APPLICABLE, err
+		return "", err
 	}
 
-	blockstr, ok := result[0].(string)
+	response, ok := result[0].(string)
 	if !ok {
-		return spectypes.NOT_APPLICABLE, errors.New("block number is not string parseable")
+		return "", errors.New("chain ID is not string parseable")
```
omerlavanet, line 118: this error is wrong since this parser is generic for blocks too

proto/spec/api_collection.proto
```diff
@@ -49,6 +49,7 @@ message ParseDirective {
   string function_template = 2;
   BlockParser result_parsing = 3 [(gogoproto.nullable) = false];
   string api_name = 4;
+  string expected_result = 5;
```
omerlavanet, line 52: let's use existing field default_value just because we can

omerlavanet (changes requested):
left changes as comments

protocol/parser/parser.go
Aleksao998: fixed 365de6df9cbd0eda56a97a1ecde6878189fea590

proto/spec/api_collection.proto
Aleksao998: Fixed 3b283cc485803420e643319f50bead02960b67c3

ranlavanet (approved):

PR #559, opened 2023-06-19
CNS-434 subscription to use fixation

orenl-lava:


x/pairing/keeper/msg_server_relay_payment.go
```diff
@@ -365,7 +365,7 @@ func (k Keeper) chargeComputeUnitsToProjectAndSubscription(ctx sdk.Context, clie
 		return fmt.Errorf("failed to add CU to the project")
 	}
 
-	err = k.subscriptionKeeper.ChargeComputeUnitsToSubscription(ctx, project.GetSubscription(), relay.CuSum)
+	err = k.subscriptionKeeper.ChargeComputeUnitsToSubscription(ctx, project.GetSubscription(), uint64(ctx.BlockHeight()), relay.CuSum)
```
Yaroms, line 368: should be the block of the relay, relay.block

x/subscription/keeper/subscription.go
```diff
@@ -217,13 +156,6 @@ func (k Keeper) CreateSubscription(
 	sub.DurationTotal = duration
```
Yaroms, line 156: doesnt this needs to be += duration?

x/subscription/keeper/subscription.go
```diff
@@ -256,20 +188,91 @@ func (k Keeper) CreateSubscription(
...
+	return err
+}
+
+func (k Keeper) advanceMonth(ctx sdk.Context, subkey []byte) {
+	date := ctx.BlockTime()
+	block := uint64(ctx.BlockHeight())
+	consumer := string(subkey)
+
+	var sub types.Subscription
+	_ = k.subsFS.FindEntry(ctx, consumer, block, &sub)
```
Yaroms, line 208: handle the return value

x/subscription/keeper/subscription.go
```diff
@@ -256,20 +188,91 @@ func (k Keeper) CreateSubscription(
...
+
+	if sub.DurationLeft == 0 {
+		utils.LavaFormatError("critical: negative duration for subscription, extend by 1 month",
+			fmt.Errorf("negative duration in AdvanceMonth()"),
+			utils.Attribute{Key: "consumer", Value: consumer},
+			utils.Attribute{Key: "plan", Value: sub.PlanIndex},
+			utils.Attribute{Key: "block", Value: block},
+		)
+		// normally would panic! but can "recover" by auto-extending by 1 month
+		expiry := NextMonth(date).UTC().Unix()
```
Yaroms, line 218: i this case we should delete the subscription? or do nothing
adding a month wont really help

x/subscription/keeper/subscription.go
```diff
@@ -256,20 +188,91 @@ func (k Keeper) CreateSubscription(
...
+			// as is with same remaining duration (but not renewed CU)
+			utils.LavaFormatError("critical: failed to recharge subscription", err,
+				utils.Attribute{Key: "consumer", Value: consumer},
+				utils.Attribute{Key: "plan", Value: sub.PlanIndex},
+				utils.Attribute{Key: "block", Value: block},
+			)
+		}
+	} else {
+		// delete all projects before deleting
+		allProjectsIDs := k.projectsKeeper.GetAllProjectsForSubscription(ctx, consumer)
```
Yaroms, line 248: lets create a function, SubscriptionCleanup

x/subscription/keeper/subscription.go
orenl-lava: `DurationTotal` is the total duration **last requested**. 
It is/will be useful for auto-renewal, so whoever does auto-renewal (be it the chain or a 3rd part) will be able to repeat the last subscription duration. For instance, a user that buys a 1-year subscription with discount, usually expects it to auto-renew by 1-year each time.

x/subscription/keeper/subscription.go
orenl-lava: Right, I meant to put a panic here. We come here only from the timer so the entry must exist, and if it doesn't we really should not to continue, as something bad happened internally.

x/subscription/keeper/subscription_test.go
```diff
@@ -581,14 +528,66 @@ func TestExpiryTime(t *testing.T) {
...
+	err = keeper.ChargeComputeUnitsToSubscription(ts.ctx, account.String(), block, 10)
+	require.Nil(t, err)
+
+	// fast-forward one month
+	ts.advanceMonths(timestamp, 1)
+	_, found = keeper.GetSubscription(ts.ctx, creator)
+	require.False(t, found)
+
+	// subscription no longer searchable, but can still charge for previous usage
+	_, found = keeper.GetSubscription(ts.ctx, account.String())
```
oren-lava, line 578: you perform this exact check 3 lines above

oren-lava (dismissed):

x/subscription/keeper/migrations.go
```diff
@@ -0,0 +1,72 @@
...
+		}
+
+		if sub_V3.DurationLeft > 0 {
+			expiry := sub_V2.MonthExpiryTime
+			if expiry >= uint64(ctx.BlockTime().UTC().Unix()) {
+				return fmt.Errorf("epiry time passed for subscription %s", err, sub_V3.Consumer)
+			}
+			keeper.subsTS.AddTimerByBlockTime(ctx, expiry, []byte(sub_V3.Consumer), []byte{})
+		} else {
+			keeper.subsFS.DelEntry(ctx, sub_V3.Consumer, sub_V2.PrevExpiryBlock)
```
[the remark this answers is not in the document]
orenl-lava: If the subscription in the the V2 store - alive or during stale-period - then it must also be in the V3 fixation store - so append is mandatory. If it also happens to be in stale-period, i.e. expired but not yet fully removed - then it has to also be deleted (to reproduce the same state; note the block used for deletion).

x/pairing/keeper/msg_server_relay_payment.go
orenl-lava: a007fabe

x/subscription/keeper/subscription.go
orenl-lava: fixed: 7ee029b6

x/subscription/keeper/subscription.go
orenl-lava: fixed: 7ee029b6

x/subscription/keeper/migrations.go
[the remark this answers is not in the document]
orenl-lava: added comments: 557e2dda

x/subscription/keeper/subscription_test.go
orenl-lava: fixed: 45467498

x/subscription/keeper/subscription.go
```diff
@@ -205,7 +205,12 @@ func (k Keeper) advanceMonth(ctx sdk.Context, subkey []byte) {
 	consumer := string(subkey)
 
 	var sub types.Subscription
-	_ = k.subsFS.FindEntry(ctx, consumer, block, &sub)
+	if found := k.subsFS.FindEntry(ctx, consumer, block, &sub); !found {
+		// subscription (monthly) timer has expired for an unknown subscription:
+		// either the timer was set wrongly, or the subscription was incorrectly
+		// removed; and we cannot even return an error about it.
+		panic("subscription month expiry with unknown key " + consumer)
```
Yaroms, line 212: lets not panic, just bail with an error print

x/subscription/keeper/subscription.go
```diff
@@ -256,20 +188,91 @@ func (k Keeper) CreateSubscription(
...
+		)
+		// normally would panic! but can "recover" by auto-extending by 1 month
+		expiry := NextMonth(date).UTC().Unix()
+		k.subsTS.AddTimerByBlockTime(ctx, uint64(expiry), []byte(consumer), []byte{})
+		return
+	}
+
+	sub.DurationLeft -= 1
+
+	if sub.DurationLeft > 0 {
```
Yaroms, line 230: also this is a uint, if it is not 0 than it must be > 0

x/subscription/keeper/subscription.go
Yaroms: i agree with the error but i dont thing adding a month is doing nothing, especially if the subscription has reached 0, this means that the problem will happen next month again? why not just delete it and move on?

orenl-lava: Something went wrong, I prefer to leave things they are for forensics and recovery.
There is no harm in moving the expiry one month forward - no additional CUs.

x/subscription/keeper/subscription.go
orenl-lava: fixed: e88390d

Yaroms (approved):

PR #562, opened 2023-06-20
[PRT-702] Enable add fix lavasession tests

Aleksao998:
## Description
In our codebase, we had a problem where lavasession unit tests were not running because we make a mistake while initializing them.
To create grpc server before every test we've used
```
func TestMain(m *testing.M){
  createGRPCServer()
}
```
but inside of it, we haven't called any tests. This PR adds a test run for every test case and also fixes tests which were failing

## Manual tests:
This PR enables lavassesion tests on every PR, but if anyone wants to run them individually, you can use
```
go test ./protocol/lavasession/...
```

## Additional info:
This PR doesn't refactor existing tests, it just enables them and makes sure that they are passing

### DIFF
--- a/protocol/lavasession/consumer_session_manager_test.go
+++ b/protocol/lavasession/consumer_session_manager_test.go
@@ -38,7 +38,7 @@ const (
 )
 
 // This variable will hold grpc server address
-var grpcListener = "localhost:28353"
+var grpcListener = "localhost:0"
 
 func CreateConsumerSessionManager() *ConsumerSessionManager {
 	rand.Seed(time.Now().UnixNano())
--- a/protocol/lavasession/consumer_session_manager_test.go
+++ b/protocol/lavasession/consumer_session_manager_test.go
@@ -4,8 +4,10 @@ import (
 	"context"
 	"encoding/json"
 	"fmt"
+	"log"
 	"math/rand"
 	"net"
+	"os"
 	"strconv"
 	"testing"
 	"time"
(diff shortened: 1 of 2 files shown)


protocol/lavasession/consumer_session_manager_test.go
```diff
@@ -28,45 +30,75 @@ const (
...
-	grpcListener                       = "localhost:28353"
 	servicedBlockNumber                = int64(30)
 	relayNumberAfterFirstCall          = uint64(1)
 	relayNumberAfterFirstFail          = uint64(1)
 	latestRelayCuAfterDone             = uint64(0)
 	cuSumOnFailure                     = uint64(0)
 )
 
+// This variable will hold grpc server address
+var grpcListener = "localhost:28353"
```
omerlavanet, line 41: dont set a specific port, use 0 to make it dynamic so it doesn't fail

protocol/lavasession/consumer_session_manager_test.go
```diff
@@ -28,45 +30,75 @@ const (
...
+
 func CreateConsumerSessionManager() *ConsumerSessionManager {
-	rand.Seed(time.Now().UnixNano())
```
omerlavanet, line 40: removing the rand seed makes session id selection deterministic

omerlavanet (changes requested):

protocol/lavasession/consumer_session_manager_test.go
Aleksao998: Returned c5b912d7298e0942b6ffffcbc62ab424ec181639

protocol/lavasession/consumer_session_manager_test.go
Aleksao998: Fixed f702f884e44c4a1fd4b45f720f3a1675d8c9c158

PR #576, opened 2023-06-26
CNS-467 panic - prune or document

orenl-lava:
Review all panic() in x/* and either -
- replace with critical error `utils.LavaFormatError("critical: ...", ...)`
- approve, by adding `// panic:ok: ...` comment
- remove, if irrelevant

(recommended to review commit-by-commit)


utils/lavalog.go
```diff
@@ -152,26 +163,30 @@ func LavaFormatLog(description string, err error, attributes []Attribute, severi
...
 
+func LavaFormatPanic(description string, err error, attributes ...Attribute) {
+	attributes = append(attributes, Attribute{Key: "StackTrace", Value: debug.Stack()})
+	LavaFormatLog(description, err, attributes, LAVA_LOG_PANIC)
+}
+
 func LavaFormatFatal(description string, err error, attributes ...Attribute) {
 	attributes = append(attributes, Attribute{Key: "StackTrace", Value: debug.Stack()})
-	LavaFormatLog(description, err, attributes, 4)
-	os.Exit(1)
```
Yaroms, line 158: is no one using this? someone in the protocol might rely on it killing the process?

utils/lavalog.go
```diff
@@ -71,20 +80,22 @@ func LavaFormatLog(description string, err error, attributes []Attribute, severi
 
 	var logEvent *zerolog.Event
 	switch severity {
-	case 4:
+	case LAVA_LOG_PANIC:
+		// prefix = "Panic:"
+		logEvent = zerologlog.Panic()
```
Yaroms, line 85: does panic actually panics?

utils/lavalog.go
orenl-lava: They don't - and they never did - because level 4 (Fatal) implies exit(1) by the logging library...

utils/lavalog.go
orenl-lava: > does panic actually panics?

time for some existential thoughts...

`zerologlog.Panic()` returns a logEvent, it doesn't panic.
the subsequent logging of the resulting logEvent implies panic() by the logging library.

x/projects/keeper/creation.go
```diff
@@ -260,9 +261,16 @@ func (k Keeper) unregisterKey(ctx sdk.Context, key types.ProjectKey, project *ty
 		if err != nil {
 			return err
 		}
+
 		found = project.DeleteKey(types.ProjectDeveloperKey(key.Key))
 		if !found {
-			panic("unregisterKey: developer key not found")
+			// the developer key must be found, because it was already found in
+			// the developerKeysFS for this project; return error to avoid panic.
+			return utils.LavaFormatWarning("critical: develper key missing in project", sdkerrors.ErrNotFound,
```
oren-lava, line 269: is this supposed to be LavaFormatWarning (can a warning be critical)?
also develper -> developer

oren-lava (changes requested):

utils/lavalog.go
oren-lava: not sure I entirely understand. The logger implies panic() but it doesn't really happen?

orenl-lava: calling `utils.LavaFormatPanic(msg, err, ...)` will spit an error to the log and then `panic(msg)`.

x/projects/keeper/creation.go
orenl-lava: nice catch. will fix.

x/pairing/keeper/keeper.go
```diff
@@ -30,6 +30,16 @@ type (
 	}
 )
 
+// sanity checks at start time
+func init() {
```
Yaroms, line 34: where is this called?

x/pairing/keeper/unresponsive_provider.go
```diff
@@ -11,10 +11,11 @@ import (
 )
 
 // Function that returns a map that links between a provider that should be punished and its providerCuCounterForUnreponsiveness
-func (k Keeper) UnstakeUnresponsiveProviders(ctx sdk.Context, epochsNumToCheckCUForUnresponsiveProvider uint64, epochsNumToCheckCUForComplainers uint64) error {
+func (k Keeper) UnstakeUnresponsiveProviders(ctx sdk.Context, epochsNumToCheckCUForUnresponsiveProvider uint64, epochsNumToCheckCUForComplainers uint64) {
 	// check the epochsNum consts
 	if epochsNumToCheckCUForComplainers <= 0 || epochsNumToCheckCUForUnresponsiveProvider <= 0 {
-		return utils.LavaFormatError("epochsNumToCheckCUForUnresponsiveProvider or epochsNumToCheckCUForComplainers are smaller or equal than zero", fmt.Errorf("invalid unresponsive provider consts"),
+		utils.LavaFormatError("epoch to check CU for unresponsive provider or for complainer is zero",
```
Yaroms, line 17: if an error was found lets return here since the logic afterwards will be broken

x/pairing/keeper/unstaking.go
```diff
@@ -62,51 +62,61 @@ func (k Keeper) UnstakeEntry(ctx sdk.Context, chainID string, creator string, un
 	return k.epochStorageKeeper.AppendUnstakeEntry(ctx, existingEntry, unstakeHoldBlocks)
 }
 
-func (k Keeper) CheckUnstakingForCommit(ctx sdk.Context) error {
+func (k Keeper) CheckUnstakingForCommit(ctx sdk.Context) {
 	// this pops all the entries that had their deadline pass
 	unstakingEntriesToCredit := k.epochStorageKeeper.PopUnstakeEntries(ctx, uint64(ctx.BlockHeight()))
 
 	if unstakingEntriesToCredit != nil {
-		err := k.creditUnstakingEntries(ctx, unstakingEntriesToCredit) // true for providers
```
Yaroms, line 70: lets not ignore completely, log the error

Yaroms: nvm, this is happening inside

x/pairing/keeper/unstaking.go
```diff
@@ -62,51 +62,61 @@ func (k Keeper) UnstakeEntry(ctx sdk.Context, chainID string, creator string, un
...
+		return fmt.Errorf("insufficient balance to unstake %s (current balance: %s)", neededAmount, moduleBalance)
+	}
+	err := k.bankKeeper.SendCoinsFromModuleToAccount(ctx, types.ModuleName, addr, []sdk.Coin{neededAmount})
+	if err != nil {
+		return fmt.Errorf("failed to send coins from module to %s: %w", addr, err)
+	}
 	return nil
 }
 
 func (k Keeper) creditUnstakingEntries(ctx sdk.Context, entriesToUnstake []epochstoragetypes.StakeEntry) error {
```
Yaroms, line 86: if we always return nil this func should not have a return value

x/pairing/keeper/keeper.go
orenl-lava: `init()` function is a feature of Go (https://go.dev/ref/spec#Package_initialization)
If it exists (per package) then it is called at package "startup" before `main().`

x/pairing/keeper/unresponsive_provider.go
orenl-lava: this is called from EpochStart(), in turn called from BeginBlock() - there is nothing to do anything with this error except log it (which is what the original code does). In other words - if there is concern about breaking things in the future then we need to panic here.

x/pairing/keeper/unstaking.go
orenl-lava: good catch; forgot to change this.

common/fixation_entry.go
```diff
@@ -233,7 +237,14 @@ func (fs *FixationStore) AppendEntry(
 		fs.setEntryIndex(ctx, safeIndex, true)
 	} else {
 		if block < latestEntry.Block {
-			panic(fmt.Sprintf("AppendEntry for block %d < latest entry block %d", block, latestEntry.Block))
+			// how come getUnmarshaledEntryForBlock lied to us?!
+			utils.LavaFormatError("critical: AppendEntry block smaller than latest",
```
oren-lava, line 241: why are we not returning? is it safe to continue without returning?

oren-lava (changes requested):

common/fixation_entry.go
Yaroms: should we keep this panic?
or return an error

x/epochstorage/keeper/stake_storage.go
```diff
@@ -119,17 +130,19 @@ func (k Keeper) removeAllEntriesPriorToBlockNumber(ctx sdk.Context, block uint64
...
-				}
 				storageBlock := entry.Index[:(len(entry.Index) - len(chainId))]
 				blockHeight, err := strconv.ParseUint(storageBlock, 10, 64)
 				if err != nil {
 					if storageBlock == "" {
 						// if storageBlock is empty its stake entry current. so we dont remove it.
 						continue
 					}
-					panic("failed to convert storage block to int: " + storageBlock)
+					utils.LavaFormatError("critical: failed to decode storage block", err,
```
Yaroms, line 140: we said this panic can stay since this function is only used for testing

common/types/fixationEntry.go
```diff
@@ -6,7 +6,7 @@ import (
 	sdk "github.com/cosmos/cosmos-sdk/types"
 )
 
-// IsStale tests whether an entry is stale, i.e. has refcount zero _and_
+// IsStaleBy tests whether an entry is stale, i.e. has refcount zero _and_
 // has passed its stale_at time (more than STALE_ENTRY_TIME since deletion).
 func (entry Entry) IsStaleBy(block uint64) bool {
```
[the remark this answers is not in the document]
orenl-lava: `block` is used in line 48 (old line 13)
`IsStale()` s like `IsStaleBy(ctx.BlockHeight())`

x/epochstorage/keeper/stake_storage.go
```diff
@@ -137,10 +137,11 @@ func (k Keeper) removeAllEntriesPriorToBlockNumber(ctx sdk.Context, block uint64
...
-					// panic:ok: should never happen; avoid further storage corruption
-					utils.LavaFormatPanic("critical: failed to decode storage block", err,
+					utils.LavaFormatError("critical: failed to decode storage block", err,
+						utils.LogAttr("chainID", chainId),
 						utils.LogAttr("index", entry.Index),
 						utils.LogAttr("storageBlock", storageBlock),
+						utils.LogAttr("block", block),
 					)
 				}
 				if blockHeight < block {
```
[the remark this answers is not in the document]
orenl-lava: the parsing code panics if it is asked to deal with unknown type (.e. anything not uint64), but returns 0 (and an error) if the parsing itself fails.  @Yaroms is correct to note that we decided to let it be because it's only called from testing code.

x/pairing/keeper/epoch_payments.go
```diff
@@ -142,12 +142,14 @@ func (k Keeper) RemoveAllEpochPaymentsForBlock(ctx sdk.Context, blockForDelete u
 
 			// validate its an old entry, for sanity
 			if uniquePaymentStorage.Block > blockForDelete {
-				// panic:ok: all payments retrieves must match the requested block (for delete)
-				utils.LavaFormatPanic("critical: failed to delete epoch payment",
+				// this should not happen; to avoid panic we simply skip this one (thus
+				// freeze the situation so it can be investigated and orderly resolved).
```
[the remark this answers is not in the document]
orenl-lava: > when you're saying this (here and before), what do you mean? does "orderly resolved" mean that we'll look at the logs, see the critical error and manually delete troublesome entries?

yes. 

> shouldn't this be automatic in some way? maybe a special handler that checks for troublesome entries and deletes them?

I intentionally avoid automatic clean up for two reasons: first, something really bad happened and making changes might cause more damage (instead, by skipping such entries now we hopefully make them harmless); second, freezing the state as early as possible makes forensics and debugging is easier.

common/fixation_entry.go
orenl-lava: good catch - should return.
fixed d71c4644

oren-lava: you didn't return?
In the commit, it's still the same

x/epochstorage/keeper/stake_storage.go
orenl-lava: right ... restore the panic and moved the function to testing code.
fixed 4b66dc3c

common/fixation_entry.go
orenl-lava: I cited the wrong commit - now it's the right commit reference.

Yaroms (approved):

Yaroms:
thanks for the hard work :)
can you fix the conflicts, ill merge after that

Yaroms (approved):

Issue #431, opened 2023-04-19
Contribution - Provider rewards persistency on shutdown


omerlavanet:
# Reward Server

# **Product Requirements Document (PRD)**

Title: Load Saved Payment Proofs from Disk at Start
Date: April 16, 2023
Author: Ran

## **Problem Statement**

The current reward server saves payment proofs, but it loses all proofs when there is a reset or panic. This problem needs to be solved to ensure that the reward server can continue its operation without losing essential information.

## **Objective**

The main objective is to enable the reward server to load saved payment proofs from disk at startup to avoid losing any data during a reset or panic.

## **Requirements**

1. Implement a mechanism to save payment proofs to disk periodically and securely.
2. Enable the reward server to load saved payment proofs from disk at startup.
3. Ensure that the saved data remains consistent and uncorrupted during the saving and loading process.
4. Provide appropriate error handling and recovery mechanisms to deal with possible disk read/write issues.

## **Success Criteria**

The success of this PRD will be measured by the ability of the reward server to load saved payment proofs from disk at startup without losing any data during resets or panics.

## **Dependencies**

- Access to the existing reward server code.

## **Assumptions**

- The provided reward server code is functional and does not contain any bugs or issues that would affect the implementation of the new functionality.

## **Constraints**

- The changes to the reward server code should not introduce any new vulnerabilities or negatively impact the performance of the system.

## **Out of Scope**

- Changes to the core functionality of the reward server that are unrelated to the saving and loading of payment proofs.

## **Milestones**

1. Design the mechanism for saving and loading payment proofs to/from disk.
2. Implement the designed mechanism in the reward server code.
3. Test the implemented mechanism to ensure it meets the success criteria.
4. Deploy the updated reward server code to the production environment.

## **Risks**

1. Disk read/write errors could lead to corrupted or lost data.
2. Changes to the reward server code could introduce bugs or vulnerabilities if not carefully implemented and tested.

## **Mitigation Strategies**

1. Implement robust error handling and recovery mechanisms to minimize the impact of disk read/write issues.
2. Thoroughly test the implemented mechanism to ensure it works as expected and does not introduce new bugs or vulnerabilities.

## **Approval**

@ranlavanet
@omerlavanet

## **Reward Server Documentation**

The Reward Server package in Golang is used for managing consumer rewards and payment handling. It is designed to interact with the Lava Network.

### **Main Components**

1. **PaymentRequest**: A struct representing the payment request information.
[...]

PR #583, opened 2023-06-28
431 persist provider rewards

alexandru-calinoiu:
Added persistence for reward server using badger

Here are the benchmark results for in memory vs local storage

BenchmarkSendNewProofInMemory-16    	  178911	      6418 ns/op
BenchmarkSendNewProofInMemory-16    	  159352	      7123 ns/op
BenchmarkSendNewProofInMemory-16    	  164092	      6725 ns/op

BenchmarkSendNewProofLocal-16    	  122911	      8856 ns/op
BenchmarkSendNewProofLocal-16    	  136767	      8672 ns/op
BenchmarkSendNewProofLocal-16    	  129583	      8709 ns/op

Solves https://github.com/lavanet/lava/issues/431

## TODO
- Most likely the storage path should be configured
- Maybe add a config for provider to choose between local vs in-memory
- Experimenting with moving save / delete to goroutines since badger is threat safe


protocol/rpcprovider/rewardserver/rewards_store.go
```diff
@@ -0,0 +1,169 @@
...
+	return strings.Join(keyParts, keySeparator)
+}
+
+func keyParts(key string) (epoch uint64, consumerAddr string, sessionId uint64, consumerKey string) {
+	parts := strings.Split(key, keySeparator)
+
+	epoch, _ = strconv.ParseUint(parts[0], 10, 64)
+	consumerAddr = parts[1]
+	sessionId, _ = strconv.ParseUint(parts[2], 10, 64)
+	consumerKey = parts[3]
```
ranlavanet, line 166: can get out of range here if the key doesn't have the consumer key.  (but also no longer need this method so we can remove it)

protocol/rpcprovider/rewardserver/rewards_store.go
```diff
@@ -0,0 +1,169 @@
...
+
+func (rs *RewardStore) FindAll(ctx context.Context) (map[uint64]*EpochRewards, error) {
+	rawRewards, err := rs.db.FindAll(ctx)
+	if err != nil {
+		return nil, err
+	}
+
+	result := make(map[uint64]*EpochRewards)
+	for key, rewards := range rawRewards {
+		epoch, consumerAddr, sessionId, consumerKey := keyParts(key)
```
ranlavanet, line 81: we dont need to extract all this information from the key, as we already unmarshal the information from the rewards object and get the rewardsEntity, which already contains the information we need. (epoch , session id, consumerkey , consumer address..

protocol/rpcprovider/rewardserver/rewards_store.go
```diff
@@ -0,0 +1,169 @@
...
+	return nil
+}
+
+func NewRewardStore(db DB) *RewardStore {
+	return &RewardStore{
+		db: db,
+	}
+}
+
+func assembleKey(epoch uint64, consumerAddr string, sessionId uint64, consumerKey string) string {
```
ranlavanet, line 146: Because we no longer need to disassmble keys (see previous comment) we now don't need a seperator and can simplify this method a bit 

` return strconv.FormatUint(epoch, 10) + consumerAddr + strconv.FormatUint(sessionId, 10) + consumerKey`

protocol/rpcprovider/rewardserver/rewards_store_test.go
```diff
@@ -0,0 +1,68 @@
...
+	require.NoError(t, err)
+
+	rewards, err := rs.FindAll(ts.ctx)
+	require.NoError(t, err)
+
+	require.Equal(t, 1, len(rewards))
+}
+
+func TestFindOne(t *testing.T) {
+	ts := setup(t)
```
ranlavanet, line 31: remember to remove setup from all tests in the protocol side :)

protocol/rpcprovider/rpcprovider.go
```diff
@@ -87,7 +87,7 @@ func (rpcp *RPCProvider) Start(ctx context.Context, txFactory tx.Factory, client
 	rpcp.providerStateTracker = providerStateTracker
 	providerStateTracker.RegisterForUpdates(ctx, statetracker.NewMetricsUpdater(providerMetricsManager))
 	// single reward server
-	rewardServer := rewardserver.NewRewardServer(providerStateTracker, providerMetricsManager)
+	rewardServer := rewardserver.NewRewardServer(providerStateTracker, providerMetricsManager, rewardserver.NewRewardStore(rewardserver.NewLocalDB(".storage/rewardserver/badger")))
```
ranlavanet, line 90: Let's make this the default value but add a viper option to the config file when setting up the rpcprovider so this path could be modifiable, you can see similar examples on rpcprovider.go (i.e viper.GetString(metrics.MetricsListenFlagName))

ranlavanet (changes requested):

protocol/rpcprovider/rewardserver/rewards_store.go
bunea: The reason for the separator is to be a bit more human friendly. In case there is ever a need to debug the entries in the database it is easier to visually parse the key. I have no strong preference either way,  so I can change it to your suggestion if performance/complexity is an issue.

ranlavanet: readability is important you can keep it that way. we don't need to disassemble also so its mostly storing :)

ranlavanet (approved):

PR #586, opened 2023-07-02
PRT-734 Adding information regarding missing cu

ranlavanet:


protocol/lavasession/provider_types.go
```diff
@@ -182,25 +182,26 @@ func (pswc *ProviderSessionsWithConsumer) atomicWriteMissingComputeUnits(cu uint
...
 	if currentMissingCU+missing > uint64(float64(max)*allowedThreshold) {
-		return false
+		return false, totalMissingCu
 	}
 	// do not allow having more missing than already used
 	if currentMissingCU+missing > used {
-		return false
+		return false, totalMissingCu
 	}
 	// TODO: use compare and swap for race avoidance
```
omerlavanet, line 202: can you do TODO? it might be a race here that triggers this

omerlavanet (changes requested):

protocol/lavasession/provider_types.go
ranlavanet: done [fa53d82](https://github.com/lavanet/lava/pull/586/commits/fa53d82e8f100b1ed50bd52dcee6fa1e5635f5e5)

omerlavanet (approved):

PR #590, opened 2023-07-04
PRT-693: Protocol Upgradeability - add version validation to both rpc consumer / provider

candostyavuz:
<h2> Description </h2>
This pull request implements a version controller for both rpcconsumer and rpcprovider that checks if the min & target versions are compatible with lavad.

<h2> Testing Cases </h2>

- Target version mismatch logs a warning message:
Screenshot 2023-07-04 at 09 34 58

- Min version mismatch throws a fatal error and logs the error message:
Screenshot 2023-07-04 at 09 36 20


protocol/rpcprovider/rpcprovider.go
```diff
@@ -270,6 +285,43 @@ func ParseEndpoints(viper_endpoints *viper.Viper, geolocation uint64) (endpoints
 	return
 }
 
+func fetchVersionFromConsensus() (string, string, error) {
```
ranlavanet, line 288: not at all how we communicate with the blockchain through cmds... 

you were supposed to use the state_tracker and the event_tracker to implement the updater 
in NewStateTracker you can validate the current version and initialize see how we query other information there 

first initializing the current version in the NewStateTracker method and then listening to changes while implementing an updater (i.e spec_updater does the same for spec updates) 

it's like a service that waits for a specific action, also when identifying mismatching versions, (within the allowed range) instead of a warning print an error and do this for every block update (so its a bit spamming so the user will notice)

ranlavanet (changes requested):
some redesign needed

protocol/upgrade/protocol_version.go
```diff
@@ -0,0 +1,128 @@
...
+
+func BuildVersionFromParamChangeEvent(event terderminttypes.Event) bool {
+	foundVersionParam := false
+	var versionValue []byte
+
+	for _, attribute := range event.Attributes {
+		key := string(attribute.Key)
+		value := string(attribute.Value)
+
+		if key == "param" {
```
Yaroms, line 54: you can simplify this like so:

if key == "param" && value != "version" 
   return false

if key == "value" 
   versionValue = attribute.Value

candostyavuz: thanks for the suggestion. fixed here: 030a24144b1653ed0663e5c43f345b075624d543

x/protocol/types/params.go
```diff
@@ -14,10 +14,10 @@ var _ paramtypes.ParamSet = (*Params)(nil)
...
 	KeyVersion     = []byte("Version")
 	DefaultVersion = Version{
-		ProviderTarget: "0.0.1",
-		ProviderMin:    "0.0.0",
-		ConsumerTarget: "0.0.1",
-		ConsumerMin:    "0.0.0",
+		ProviderTarget: "0.16.0",
+		ProviderMin:    "0.16.0",
+		ConsumerTarget: "0.16.0",
+		ConsumerMin:    "0.16.0",
```
ranlavanet, line 20: we need to change it to v0.20.2 
also change the protocol version

protocol/upgrade/protocol_version.go
```diff
@@ -0,0 +1,122 @@
...
+
+type ProtocolVersion struct {
+	ProviderTarget string
+	ProviderMin    string
+	ConsumerTarget string
+	ConsumerMin    string
+}
+
+var LavaProtocolVersion = ProtocolVersion{
+	ProviderTarget: "0.16.0",
```
ranlavanet, line 22: we need to change it to v0.20.2

ranlavanet (approved):

PR #591, opened 2023-07-04
CNS-479 add project key effective now

orenl-lava:
Changes are non-trivial the handle the following two cases:
    
1. delete a key handles admin key immediately, but developer key deferred to beginning of next epoch.
2. add a key after a previous delete key in the same epoch (which created a future entry); not necessarily the same key.

### DIFF
--- a/common/fixation_entry.go
+++ b/common/fixation_entry.go
@@ -28,6 +28,7 @@ import (
 //    - ModifyEntry(index, block, *entry): modify an existing entry with "index" and exact "block" (*)
 //    - ReadEntry(index, block, *entry): copy an existing entry with "index" and exact "block" (*)
 //    - FindEntry(index, block, *entry): get a copy (no reference) of a version of an entry (**)
+//    - FindEntry2(index, block, *entry): get a copy (no reference) of a version of an entry (**)
 //    - GetEntry(index, *entry): get a copy (and reference) of the latest version of an entry
 //    - PutEntry(index, block): drop reference to an existing entry with "index" and exact "block" (*)
 //    - DelEntry(index, block): mark an entry as unavailable for new GetEntry() calls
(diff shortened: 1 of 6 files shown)


common/fixation_entry.go
```diff
@@ -511,14 +512,14 @@ func (fs *FixationStore) getUnmarshaledEntryForBlock(ctx sdk.Context, safeIndex
 	return types.Entry{}, false
 }
 
-// FindEntry returns the entry by index and block without changing the refcount
-func (fs *FixationStore) FindEntry(ctx sdk.Context, index string, block uint64, entryData codec.ProtoMarshaler) bool {
+// FindEntry2 returns the entry by index and block without changing the refcount
+func (fs *FixationStore) FindEntry2(ctx sdk.Context, index string, block uint64, entryData codec.ProtoMarshaler) (uint64, bool) {
```
Yaroms, line 516: we need a more concrete name for this

x/projects/keeper/project.go
```diff
@@ -51,115 +62,204 @@ func (k Keeper) GetProjectForDeveloper(ctx sdk.Context, developerKey string, blo
...
+		err := k.registerKey(ctx, projectKey, &project, ctxBlock)
+		if err != nil {
+			return utils.LavaFormatError("failed to register key to project", err,
+				utils.Attribute{Key: "project", Value: projectID},
+				utils.Attribute{Key: "block", Value: ctxBlock},
+			)
+		}
+	}
+
+	err = k.projectsFS.AppendEntry(ctx, projectID, ctxBlock, &project)
```
Yaroms, line 91: we need to make sure the keys are registered in the beggining of the epoch, instead of append use modify entry to edit the project

Yaroms: actually we can use append but make sure to use the block of the project and not ctxBlock

x/projects/keeper/project.go
```diff
@@ -51,115 +62,204 @@ func (k Keeper) GetProjectForDeveloper(ctx sdk.Context, developerKey string, blo
...
+	// futureBlock is the version found nearest to (not-larger than) next epoch; If it
+	// is smaller than nextEpoch (specifically, equal to ctxBlock) then we know there
+	// isn't a future entry there. But if it is euqal to nextEpoch then we know there
+	// exists a future version at next epoch different than the current version (must
+	// have been created due to an delete key earlier in this epoch) - so add the key
+	// to that future version too.
+
+	if futureBlock == nextEpoch {
+		for _, projectKey := range projectKeys {
+			err := k.registerKey(ctx, projectKey, &projectNextEpoch, nextEpoch)
```
Yaroms, line 132: in this case souldnt we just add the keys to the keys list in the project?

x/projects/keeper/project.go
```diff
@@ -51,115 +62,204 @@ func (k Keeper) GetProjectForDeveloper(ctx sdk.Context, developerKey string, blo
...
 			utils.Attribute{Key: "project", Value: projectID},
 			utils.Attribute{Key: "block", Value: ctxBlock},
 		)
 	}
 
-	if !project.IsAdminKey(adminKey) {
-		return utils.LavaFormatWarning("failed to add keys",
-			fmt.Errorf("requesting key must be admin key"),
+	for _, projectKey := range projectKeys {
+		err := k.registerKey(ctx, projectKey, &project, ctxBlock)
```
Yaroms, line 82: should always be an epoch

oren-lava (approved):

common/fixation_entry.go
orenl-lava: It's the FindEntry with 2 return values :)

Yaroms (approved):

Yaroms (approved):

PR #592, opened 2023-07-04
CNS-484: implement generic pairing scoring mechanism

oren-lava:
This mechanism allows choosing provider for pairing according to a generic scoring mechanism.
Currently, we allocate pairing slots, group slots with identical requirements and pick providers to fill each group.

Important notes:
1. the only supported requirement is stake, so this PR only changes the choice mechanism in the back, but doesn't add new features.
2. In the original design we wanted to sort the groups so adjacent group would have minimal changes in their requirements. The sort was not implemented to allow fast deployment
3. In the next PR, we'll implement the GeoRequirement, which will add geolocation considerations in the choice mechanism


oren-lava:
closing due to merging main in

PR #594, opened 2023-07-05
CNS-484: implement generic pairing scoring mechanism

oren-lava:
This mechanism allows choosing provider for pairing according to a generic scoring mechanism.
Currently, we allocate pairing slots, group slots with identical requirements and pick providers to fill each group.

Important notes:

the only supported requirement is stake, so this PR only changes the choice mechanism in the back, but doesn't add new features.
In the original design we wanted to sort the groups so adjacent group would have minimal changes in their requirements. The sort was not implemented to allow fast deployment
In the next PR, we'll implement the GeoRequirement, which will add geolocation considerations in the choice mechanism


x/pairing/keeper/scores/score_req.go
```diff
@@ -0,0 +1,159 @@
...
+	planstypes "github.com/lavanet/lava/x/plans/types"
+	tendermintcrypto "github.com/tendermint/tendermint/crypto"
+)
+
+var uniformStrategy scorestypes.ScoreStrategy
+
+// TODO: currently we'll use weight=1 for all reqs. In the future, we'll get it from policy
+func init() {
+	uniformStrategy = make(scorestypes.ScoreStrategy)
+	uniformStrategy[scorestypes.STAKE_REQ_NAME] = 1
```
Yaroms, line 21: lets use the type instead of defining a string

x/pairing/keeper/scores/score_req.go
```diff
@@ -0,0 +1,159 @@
...
+	for i := range slots {
+		slots[i] = scorestypes.NewPairingSlot(stakeReqMap)
+	}
+
+	return slots
+}
+
+// group the slots
+func GroupSlots(slots []*scorestypes.PairingSlot) []scorestypes.PairingSlotGroup {
+	slotGroups := []scorestypes.PairingSlotGroup{}
```
Yaroms, line 39: add some comments to explain what is going on through the method

x/pairing/keeper/scores/score_req.go
```diff
@@ -0,0 +1,159 @@
...
+		reqName := req.GetName()
+		weight, ok := strategy[reqName]
+		if !ok {
+			return utils.LavaFormatError("req not in strategy", sdkerrors.ErrKeyNotFound,
+				utils.Attribute{Key: "req_bitmap_value", Value: reqName})
+		}
+
+		for _, score := range scores {
+			// normalize stake so we won't overflow the score result (uint64)
+			providerWithNormalizedStake := score.Provider
```
Yaroms, line 84: providerWithNormalizedStake is a pointer, meaning if we have mote than 1 req it will normalize it again, should be done in the stakescore.score

x/pairing/types/scores/stake_req.go
```diff
@@ -0,0 +1,24 @@
...
+	epochstoragetypes "github.com/lavanet/lava/x/epochstorage/types"
+)
+
+// stake requirement that implements the ScoreReq interface
+type StakeReq struct {
+	stake sdk.Coin
+}
+
+const (
+	STAKE_REQ_NAME = "stake-req"
```
Yaroms, line 14: lets use the type instead of defining a string

x/pairing/types/scores/stake_req.go
```diff
@@ -0,0 +1,24 @@
...
+)
+
+// calculates the stake score of a provider (which is simply the normalized stake)
+func (sr StakeReq) Score(provider epochstoragetypes.StakeEntry, weight uint64) uint64 {
+	return provider.Stake.Amount.ToDec().Power(weight).BigInt().Uint64()
```
Yaroms, line 19: add here the normalizing of the stake

x/pairing/types/scores/score_req.go
```diff
@@ -0,0 +1,76 @@
+package score
```
Yaroms, line 1: this file is more than the ScoreReq, either change its name or move the other structs to another file

Yaroms (commented):
good job :) very clean code

x/pairing/keeper/scores/score_req.go
orenshva: done. see https://github.com/lavanet/lava/pull/594/commits/b0c34d4222b4a24e26081ea469300481dafa0f5d

x/pairing/keeper/scores/score_req.go
orenshva: done. see https://github.com/lavanet/lava/commit/434a5174aa62741e54089e5d392dbe39d9c58d50

x/pairing/keeper/scores/score_req.go
oren-lava: done. see https://github.com/lavanet/lava/commit/b861439bb85b2ef959e9e3594ed7ed25f0946bc3

x/pairing/types/scores/score_req.go
oren-lava: done. see https://github.com/lavanet/lava/pull/594/commits/c1af94249ab68d6734ff129f2d937f389db2fba2

oren-lava: disregard last comment (moving files again). Will update commit soon

oren-lava: done. see https://github.com/lavanet/lava/pull/594/commits/0372ac47a9f29e94684b16557db8de6e693ffc9c

Yaroms (approved):

x/pairing/keeper/scores/pairing_score.go
```diff
@@ -0,0 +1,31 @@
...
+
+	epochstoragetypes "github.com/lavanet/lava/x/epochstorage/types"
+)
+
+const (
+	// TODO: temp strategy weight until we implement taking it out of the policy
+	UNIFORM_WEIGHT uint64 = 1
+)
+
+// object to hold a provider's score and the requirements that construct the score (map[scoreReqName] -> score from that req)
```
orenl-lava, line 12: Let's stick to Go's doc style going forward (see: https://go.dev/blog/godoc)
For instance, here could be:
```
// PairingScore holds a provider's score with respect to a set of requirements (ScoreReq).
```

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+	tendermintcrypto "github.com/tendermint/tendermint/crypto"
+)
+
+var (
+	uniformStrategy ScoreStrategy
+	allReqNames     []string
+)
+
+// TODO: currently we'll use weight=1 for all reqs. In the future, we'll get it from policy
+func init() {
```
orenl-lava, line 83: you can initialize these directly and drop the `init()` function. e.g.
```
var allReqNames = []string{
    stake_req_ name,
}

var uniformStrategy = ScoreStrategy{ 
    stake_req_name: 1,
}
```

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+	// init strategy
+	uniformStrategy = make(ScoreStrategy)
+	for _, reqName := range allReqNames {
+		uniformStrategy[reqName] = 1
+	}
+}
+
+// get the overall requirements from the policy and assign slots that'll fulfil them
+// TODO: this function should be changed in the future since it only supports stake reqs
+func CalcSlots(policy planstypes.Policy, minStake sdk.Int) []*PairingSlot {
```
orenl-lava, line 96: should `minStake` be part of the policy?
it's weird that it is passed as another arg to `CalcSlots()` - what other extra args would it take later?

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+	slotReqs := map[string]ScoreReq{stakeReq.GetName(): stakeReq}
+	for i := range slots {
+		slots[i] = NewPairingSlot(slotReqs)
+	}
+
+	return slots
+}
+
+// group the slots
+func GroupSlots(slots []*PairingSlot) []PairingSlotGroup {
```
orenl-lava, line 111: Grouping can be done on `PairingSlot` directly (add `Count` field)

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+	slotGroups := []PairingSlotGroup{}
+	if len(slots) == 0 {
+		utils.LavaFormatError("no slots", sdkerrors.ErrLogic)
+		return slotGroups
+	}
+
+	for k := range slots {
+		foundGroup := false
+		for i := range slotGroups {
+			diff := slots[k].Diff(slotGroups[i].Slot)
```
orenl-lava, line 121: if this intends to catch equality, then this `Diff` is insufficient.
consider `slot[k] = {a, b}`, and `slotGroups[i] = {a, b, c}` for example.
replace "Diff" with "Equal".

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+			if len(diff.Reqs) == 0 {
+				slotGroups[i].Count += 1
+				foundGroup = true
+				break
+			}
+		}
+
+		if !foundGroup {
```
orenl-lava, line 129: this test is unneeded because in line 125 breaks out of the inner loop when `foundGroup = true`

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// TODO: currently we'll use weight=1 for all reqs. In the future, we'll get it from policy
+func GetStrategy() ScoreStrategy {
+	return uniformStrategy
+}
+
+// calculates the final pairing score for all slot groups (with strategy)
+// we calculate only the diff between the current and previous slot groups
+func CalcPairingScore(scores []*PairingScore, strategy ScoreStrategy, diffSlot *PairingSlot, minStake sdk.Int) error {
```
orenl-lava, line 145: same comment about the arg `minStake` here as in `CalcSlots()`

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+	// calculate the score for each req for each provider
+	for _, req := range diffSlot.Reqs {
+		reqName := req.GetName()
+		weight, ok := strategy[reqName]
+		if !ok {
+			return utils.LavaFormatError("req not in strategy", sdkerrors.ErrKeyNotFound,
```
orenl-lava, line 151: This is critical error (a la panic).
In fact, you can confirm at start time (e.g. with `init()` function) that the `len(stragey) == len(allReqNames)` and panic if not.

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+				utils.Attribute{Key: "req_name", Value: reqName})
+		}
+
+		for _, score := range scores {
+			// normalize stake so we won't overflow the score result (uint64)
+			newScoreComp := req.Score(*score.Provider, weight)
```
orenl-lava, line 157: Shouldn't `weight` be applied on the `newScoreComp` (rather than passed to `req.Score`)?

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+
+			// divide by previous score component (if exists) and multiply by new score
```
orenl-lava, line 159: why is this?

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+			prevReqScoreComp, ok := score.ScoreComponents[reqName]
+			if ok {
+				if prevReqScoreComp == 0 {
+					return utils.LavaFormatError("previous req score is zero", fmt.Errorf("invalid req score"),
```
orenl-lava, line 163: should this be a critical error?  I'd think  that this should never happen no matter what.

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+						utils.Attribute{Key: "req_name", Value: reqName})
```
orenl-lava, line 164: also say something about the provider/stake-entry, for post-mortem analysis.

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+
+	return slotGroups
+}
+
+// TODO: currently we'll use weight=1 for all reqs. In the future, we'll get it from policy
+func GetStrategy() ScoreStrategy {
+	return uniformStrategy
+}
+
+// calculates the final pairing score for all slot groups (with strategy)
```
orenl-lava, line 143: this comment looks inaccurate: `CalcPairingScore()` handles a single slot (group).
and also unsure if it calculates diff between current and previous slot groups.

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+}
+
+// given a list of scores, pick a <group-count> providers with a pseudo-random weighted choice
+func PickProviders(ctx sdk.Context, projectIndex string, scores []*PairingScore, groupCount uint64, block uint64, chainID string, epochHash []byte, indexToSkipPtr *map[int]bool) (returnedProviders []epochstoragetypes.StakeEntry) {
+	scoreSum := sdk.NewUint(0)
+	hashData := make([]byte, 0)
+	for _, providerScore := range scores {
+		scoreSum = scoreSum.Add(sdk.NewUint(providerScore.Score))
+	}
+	if scoreSum.IsZero() {
```
orenl-lava, line 185: if this intends to check for empty list - then better test directly (and before the above loop).
otherwise, is it possible to get a zero score sum?  (if not, treat it like the unexpected score zero in line 162)

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+
+			// update the score component map
+			score.ScoreComponents[reqName] = newScoreComp
+		}
+	}
+
+	return nil
+}
+
+// given a list of scores, pick a <group-count> providers with a pseudo-random weighted choice
```
orenl-lava, line 178: this function merits more info (if only by the number of args)

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+		// list is empty
+		return returnedProviders
+	}
+
+	// add the session start block hash to the function to make it as unpredictable as we can
+	hashData = append(hashData, epochHash...)
+	hashData = append(hashData, chainID...)      // to make this pairing unique per chainID
+	hashData = append(hashData, projectIndex...) // to make this pairing unique per consumer
+
+	indexToSkip := *indexToSkipPtr
```
orenl-lava, line 195: I see how it's used, but unclear why it comes from the caller.
(why is the arg a pointer to begin with?)

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+	for it := 0; it < int(groupCount); it++ {
+		hash := tendermintcrypto.Sha256(hashData) // TODO: we use cheaper algo for speed
+		bigIntNum := new(big.Int).SetBytes(hash)
+		hashAsNumber := sdk.NewUintFromBigInt(bigIntNum)
+		modRes := hashAsNumber.Mod(scoreSum)
+
+		newScoreSum := sdk.NewUint(0)
+		// we loop the servicers list form the end because the list is sorted, biggest is last,
```
orenl-lava, line 203: who sorts it?
(also fix typo)

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,230 @@
...
+			// update the score component map
+			score.ScoreComponents[reqName] = newScoreComp
+		}
+	}
+
+	return nil
+}
+
+// given a list of scores, pick a <group-count> providers with a pseudo-random weighted choice
+func PickProviders(ctx sdk.Context, projectIndex string, scores []*PairingScore, groupCount uint64, block uint64, chainID string, epochHash []byte, indexToSkipPtr *map[int]bool) (returnedProviders []epochstoragetypes.StakeEntry) {
```
orenl-lava, line 179: arg `block` is unused below.
how about one `hash` arg, and the caller responsible to generate that hash?  not only less args, but also it's the caller that knows what these args means and why use them to generate the seed).

x/pairing/keeper/pairing.go
```diff
@@ -110,24 +109,28 @@ func (k Keeper) GetPairingForClient(ctx sdk.Context, chainID string, clientAddre
...
+		return nil, 0, "", fmt.Errorf("did not find providers for pairing: epoch:%d, chainID: %s", block, chainID)
+	}
+
 	project, err := k.GetProjectData(ctx, clientAddress, chainID, block)
 	if err != nil {
 		return nil, 0, "", err
 	}
 
-	strictestPolicy, allowedCU, err = k.getProjectStrictestPolicy(ctx, project, chainID)
+	strictestPolicy, allowedCU, minStake, err := k.getProjectStrictestPolicy(ctx, project, chainID)
```
orenl-lava, line 127: (with reference to other comments on `minStake` - maybe store it inside `strictedPolicy` (and then rid the extra arg to `CalcSlots()` ?)

x/pairing/keeper/pairing.go
```diff
@@ -137,16 +140,42 @@ func (k Keeper) getPairingForClient(ctx sdk.Context, chainID string, clientAddre
...
+	for i := range possibleProviders {
+		providerScore := pairingscores.NewPairingScore(&possibleProviders[i])
+		providerScores = append(providerScores, providerScore)
+	}
+
+	// calculate score (always on the diff in score components of consecutive groups) and pick providers
+	prevGroupSlot := pairingscores.NewPairingSlot(map[string]pairingscores.ScoreReq{}) // init dummy slot to compare to
+	indexToSkipMapPtr := make(map[int]bool)                                            // keep the indices of chosen providers to we won't pick the same providers twice (for different groups)
+	for _, group := range slotGroups {
+		diffSlot := group.Slot.Diff(prevGroupSlot)
```
orenl-lava, line 160: What is the reason/purpose of using the diff here?

x/pairing/keeper/pairing.go
```diff
@@ -137,16 +140,42 @@ func (k Keeper) getPairingForClient(ctx sdk.Context, chainID string, clientAddre
...
+		err := pairingscores.CalcPairingScore(providerScores, pairingscores.GetStrategy(), diffSlot, minStake)
+		if err != nil {
+			return nil, 0, "", err
+		}
+
+		pickedProviders := pairingscores.PickProviders(ctx, project.Index, providerScores, group.Count, block, chainID, epochHash, &indexToSkipMapPtr)
+		providers = append(providers, pickedProviders...)
```
orenl-lava, line 167: it looked like `PickProviders()` could pick more providers than asked. 
if that's the case, should you trim it before appending?

orenl-lava (changes requested):
Missing is how weights are used (I know they are all 1 now, but still unused).

I wonder about the weights - given a slot, should they be normalized in some way based on the total sum (or product) of the relevant weights (for that slot)?

For example, if slot A has requirements {a, b, c, d}, and slot B has requirements {a, e}, weights - should the weights remain as is in both contexts?

orenl-lava:
Missing is how weights are used (I know they are all 1 now, but still unused).

I wonder about the weights - given a slot, should they be normalized in some way based on the total sum (or product) of the relevant weights (for that slot)?

For example, if slot A has requirements {a, b, c, d}, and slot B has requirements {a, e}, weights - should the weights remain as is in both contexts?

x/plans/types/plan.pb.go
```diff
@@ -56,8 +56,9 @@ func (SELECTED_PROVIDERS_MODE) EnumDescriptor() ([]byte, []int) {
 	return fileDescriptor_e5909a10cd0e3497, []int{0}
 }
 
-// The geolocations' values are powers of 2 that act as a bit map.
```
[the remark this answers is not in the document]
oren-lava: outdated (not showing there now)

x/pairing/types/scores/stake_req.go
```diff
@@ -0,0 +1,16 @@
...
+	epochstoragetypes "github.com/lavanet/lava/x/epochstorage/types"
+)
+
+// stake requirement that implements the ScoreReq interface
+type StakeReq struct {
+	stake sdk.Coin
+}
+
+// calculates the stake score of a provider (which is simply the normalized stake)
+func (sr StakeReq) Score(provider epochstoragetypes.StakeEntry, weight uint64) uint64 {
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/6905360993861e914ed47e23a32267749b939c4c

x/pairing/types/scores/score_req.go
```diff
@@ -0,0 +1,77 @@
...
+
+const (
+	// TODO: temp strategy weight until we implement taking it out of the policy
+	UNIFORM_WEIGHT uint64 = 1
+)
+
+// Score() calculates a provider's score according to the requirement
+// GetName() gets the ScoreReq's name
+type ScoreReq interface {
+	Score(provider epochstoragetypes.StakeEntry, weight uint64) uint64
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/6905360993861e914ed47e23a32267749b939c4c

x/pairing/keeper/scores/score_req.go
```diff
@@ -0,0 +1,14 @@
...
+import (
+	epochstoragetypes "github.com/lavanet/lava/x/epochstorage/types"
+)
+
+// when adding a new req, update CalcSlots()
+
+// Score() calculates a provider's score according to the requirement
+// GetName() gets the ScoreReq's name
+type ScoreReq interface {
+	Score(provider epochstoragetypes.StakeEntry, weight uint64) uint64
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/6905360993861e914ed47e23a32267749b939c4c

x/pairing/types/scores/stake_req.go
```diff
@@ -0,0 +1,16 @@
...
+	return provider.Stake.Amount.ToDec().Power(weight).BigInt().Uint64()
```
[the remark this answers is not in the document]
oren-lava: outdated. in the newer version, I normalize the stake to avoid overflow

x/pairing/types/scores/score_req.go
```diff
@@ -0,0 +1,77 @@
...
+
+// when adding a new req, update CalcSlots()
+
+const (
+	// TODO: temp strategy weight until we implement taking it out of the policy
+	UNIFORM_WEIGHT uint64 = 1
+)
+
+// Score() calculates a provider's score according to the requirement
+// GetName() gets the ScoreReq's name
```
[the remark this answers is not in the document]
oren-lava: outdated. There is such API now

x/pairing/keeper/scores/score_req.go
```diff
@@ -0,0 +1,14 @@
+package scores
+
+import (
+	epochstoragetypes "github.com/lavanet/lava/x/epochstorage/types"
+)
+
+// when adding a new req, update CalcSlots()
```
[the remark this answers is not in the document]
oren-lava: yes. if you won't update it with your new req, the slots won't get that new req. do you have any suggestion on how can I avoid it?

x/pairing/keeper/scores/score.go
oren-lava: yeah, but if the loop ends without breaking, foundGroup == false and we need to create a new group. This check is necessary

x/pairing/keeper/pairing.go
oren-lava: When I'm calculating the score for a slot, I can calculate only parts of the score if the previous and current slots have shared score components. For example, let's say one slot has stake+geo-eu+archive and the next one has stake+geo-eu. Instead of calculating the next slot's score from scratch I can just remove the archive score component from the previous score

x/pairing/keeper/scores/pairing_score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/13d50d5931010aab718c8591c5e649e16ad380c0

x/pairing/keeper/scores/stake_req.go
```diff
@@ -0,0 +1,23 @@
+package scores
+
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	epochstoragetypes "github.com/lavanet/lava/x/epochstorage/types"
+)
+
+const stake_req_name = "stake-req"
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/13d50d5931010aab718c8591c5e649e16ad380c0

x/pairing/keeper/scores/stake_req.go
```diff
@@ -0,0 +1,23 @@
...
+
+// stake requirement that implements the ScoreReq interface
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/13d50d5931010aab718c8591c5e649e16ad380c0

x/pairing/keeper/scores/stake_req.go
```diff
@@ -0,0 +1,23 @@
...
+type StakeReq struct {
+	MinStake sdk.Int
+}
+
+// calculates the stake score of a provider (which is simply the normalized stake)
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/13d50d5931010aab718c8591c5e649e16ad380c0

x/pairing/keeper/scores/score_req.go
```diff
@@ -0,0 +1,14 @@
...
+
+// Score() calculates a provider's score according to the requirement
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/13d50d5931010aab718c8591c5e649e16ad380c0

x/pairing/keeper/scores/pairing_slot.go
```diff
@@ -0,0 +1,39 @@
+package scores
+
+// object to hold the requirements for a specific slot (map of req names pointing to req object)
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/13d50d5931010aab718c8591c5e649e16ad380c0

x/pairing/keeper/scores/pairing_slot.go
```diff
@@ -0,0 +1,39 @@
...
+type PairingSlot struct {
+	Reqs map[string]ScoreReq
+}
+
+func NewPairingSlot(reqs map[string]ScoreReq) *PairingSlot {
+	slot := PairingSlot{
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/13d50d5931010aab718c8591c5e649e16ad380c0

x/pairing/keeper/scores/pairing_slot.go
```diff
@@ -0,0 +1,39 @@
...
+		Reqs: reqs,
+	}
+	return &slot
+}
+
+// generate a diff slot that contains the reqs that are in the slot receiver but not in the "other" slot
+func (s PairingSlot) Diff(other *PairingSlot) *PairingSlot {
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/13d50d5931010aab718c8591c5e649e16ad380c0

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/13d50d5931010aab718c8591c5e649e16ad380c0

x/pairing/keeper/scores/score.go
oren-lava: removed the check and put it in init(). see https://github.com/lavanet/lava/commit/c2ce6a219214545c35ff7e455b49f6f3a89152e0

x/pairing/keeper/scores/score.go
oren-lava: kept the init function so I can verify that len(allReqs) == len(strategy). see https://github.com/lavanet/lava/commit/c2ce6a219214545c35ff7e455b49f6f3a89152e0

x/pairing/keeper/scores/score.go
oren-lava: in my geolocation branch (CNS-359) I put an equal function there. So I won't fix it here

x/pairing/keeper/scores/pairing_slot.go
```diff
@@ -0,0 +1,39 @@
...
+	reqsDiff := make(map[string]ScoreReq)
+	for key := range s.Reqs {
+		if _, found := other.Reqs[key]; !found {
+			reqsDiff[key] = s.Reqs[key]
+		}
+	}
+
+	return NewPairingSlot(reqsDiff)
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/e42e3940323fd67a526ff4bc1b57c536d2c37d56

x/pairing/keeper/scores/pairing_slot.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// object to hold a slot and the number of times it's required
+type PairingSlotGroup struct {
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/e42e3940323fd67a526ff4bc1b57c536d2c37d56

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/e42e3940323fd67a526ff4bc1b57c536d2c37d56

x/pairing/keeper/scores/score.go
oren-lava: after consulting with @Yaroms, we decided to keep it as it is

x/pairing/keeper/pairing.go
oren-lava: after consulting with @Yaroms, we decided to keep it as it is

x/pairing/keeper/scores/score_req.go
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/67b0f52bd0d3af56d04eebbcb4981b8d1aebcd4f

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/67b0f52bd0d3af56d04eebbcb4981b8d1aebcd4f

x/pairing/keeper/scores/score.go
oren-lava: basically to avoid unnecessary calculation. I recalculate the score components of diffSlot so this is the way to update the provider's existing score according to specific components only

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/accb3416d8b3fdb6c70a372d44b04d27f2db36b3

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/accb3416d8b3fdb6c70a372d44b04d27f2db36b3

x/pairing/keeper/scores/score.go
oren-lava: I copied this function from the old pairing mechanism (originally written by Omer), and this is an unwanted relic. Should I just delete the comment or do you think it's necessary to perform sort?

x/pairing/keeper/scores/score.go
orenl-lava: sure; please share the reasoning?

x/pairing/keeper/scores/score_req.go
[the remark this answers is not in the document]
orenl-lava: I meant the comment -- it looks out of place, and and any how new score-req implementation don't get added in this file, it isn't that useful either.

x/pairing/types/scores/stake_req.go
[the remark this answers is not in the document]
orenl-lava: Nice.  I believe there are still similar leftovers e.g.
```    possibleProviders, found, epochHash := k.epochStorageKeeper.GetEpochStakeEntries(ctx, epoch, chainID)```
that need rename too.

File: x/pairing/keeper/scores/score.go (hunk 12 above)
Comment (orenl-lava) [reply x/pairing/keeper/scores/score.go]: right. FWIW, you can use `continue LABEL` (see https://relistan.com/continue-statement-with-labels-in-go), but I see this code was replaced altogether anyhow.

File: x/pairing/keeper/scores/score.go (hunk 23 above)
Comment (orenl-lava) [reply x/pairing/keeper/scores/score.go]: depends on the desired functionality. 
it sounds like sorting helps the goal (in terms of scoring strategy). 
it also may help determinism (but does not guarantee order of entries with same score).

File: x/pairing/keeper/pairing.go (hunk 26 above)
Comment (orenl-lava) [reply x/pairing/keeper/pairing.go]: it would require careful handling in that case.
but anyhow I see this code is gone.

File: x/pairing/keeper/scores/pairing_slot.go [hunk 44]
```diff
@@ -0,0 +1,38 @@
...
+	}
+
+	diffSlot := NewPairingSlot()
+	diffSlot.Reqs = reqsDiff
+	return diffSlot
+}
+
+// GetSlotKey generates a unique key of the slot based on its requirements
+func (s PairingSlot) GetSlotKey() string {
+	key := ""
```
orenl-lava, line 32: how about:
```
    keys := make([]string, 0, len(s.Reqs))
    for k := range s.Reqs {
        keys = append(keys, k)
    }
    return strings.Join(keys, "-")
```

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,214 @@
...
+	allReqNames = append(allReqNames, stakeReqName)
+
+	// init strategy
+	uniformStrategy = make(ScoreStrategy)
+	for _, reqName := range allReqNames {
+		uniformStrategy[reqName] = 1
+	}
+
+	if len(allReqNames) != len(uniformStrategy) {
+		panic("strategy does not contain all score reqs")
```
orenl-lava, line 69: "uniform strategy"

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,214 @@
...
+	}
+
+	return slots
+}
+
+// group the slots
+func GroupSlots(slots []*PairingSlot) []*PairingSlot {
+	uniqueSlots := make(map[string]*PairingSlot)
+	if len(slots) == 0 {
+		utils.LavaFormatError("no slots", sdkerrors.ErrLogic)
```
orenl-lava, line 94: if this should be impossible, then make it "critical".
otherwise what's the error?

x/pairing/keeper/scores/pairing_slot.go
```diff
@@ -0,0 +1,38 @@
...
+		}
+	}
+
+	diffSlot := NewPairingSlot()
+	diffSlot.Reqs = reqsDiff
+	return diffSlot
+}
+
+// GetSlotKey generates a unique key of the slot based on its requirements
+func (s PairingSlot) GetSlotKey() string {
```
orenl-lava, line 31: actually, in Go iteration over a map is not specified and not guarantee to be the same even from one iteration to another. if this key is being used for comparisons (of slots), then the list must be sorted alphabetically - or each ScoreReq implementation would tell not only it's name but also it's (pre-defined) order for the looping.

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,214 @@
...
+		return []*PairingSlot{}
+	}
+
+	for i := 0; i < len(slots); i++ {
+		slot := slots[i]
+		key := slot.GetSlotKey()
```
orenl-lava, line 100: GetSlotKey() must be deterministic - but it isn't. See comment there.

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,214 @@
...
+
+		for _, score := range scores {
+			newScoreComp := req.Score(*score.Provider)
+			newScoreComp = uint64(math.Pow(float64(newScoreComp), float64(weight)))
+
+			// divide by previous score component (if exists) and multiply by new score
+			prevReqScoreComp, ok := score.ScoreComponents[reqName]
+			if ok {
+				if prevReqScoreComp == 0 {
+					utils.LavaFormatFatal("previous req score is zero", fmt.Errorf("invalid req score"),
```
orenl-lava, line 142: replace with: `utils.LavaFormatError("critical: ...", ...)`
(the `LavAFormatFatal()` version would exit with an error)

x/pairing/keeper/scores/score.go
oren-lava: if we're changing it, there are two options:
1. putting MinStake inside policy protobuf, which will require a migrator
2. creating a struct that will hold the policy and the minStake

even if there are other methods of solving this, Yarom didn't want to waste time on it because it looks kinda weird

x/pairing/keeper/scores/score.go
oren-lava: to save time (as discussed in the daily), I'm keeping this code.

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/da6f0ad1b3b168a14adf906e971bbe1fc62678ad

x/pairing/keeper/scores/score_req.go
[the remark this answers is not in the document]
oren-lava: removed. see https://github.com/lavanet/lava/commit/ff4a09b8652aeb456bdb316710baf563103d764d

x/pairing/types/scores/stake_req.go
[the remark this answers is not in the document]
oren-lava: changed. see https://github.com/lavanet/lava/commit/ff4a09b8652aeb456bdb316710baf563103d764d

x/pairing/types/scores/stake_req.go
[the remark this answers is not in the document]
oren-lava: what do you mean?
the weight is already uint64 and the provider's stake is saved as a sdk.Coin object which I can't change (we use it because it's more secure than uint64)

x/pairing/keeper/scores/score.go
oren-lava: I sort the scores like expected in the comment. If there are equal scores, I determine their order by the provider's address (alphabetically) to keep determinism. see https://github.com/lavanet/lava/commit/81e1f47bebf82edec3e41942d4f9c428954b4cbd

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/d30134e7c735b1c043ac5b7455dca1dff4ad612d
note that I use a normal panic because my branch doesn't have LavaFormatPanic. I've opened a task to change the panic instances after everything is done

x/pairing/keeper/scores/score.go
oren-lava: ignore last comment. After talking with Yarom, I removed the sort. see https://github.com/lavanet/lava/commit/d30134e7c735b1c043ac5b7455dca1dff4ad612d

x/pairing/keeper/pairing.go
oren-lava: PickProviders pick group.Count providers. All the groups' count add up to the amount of slots. The number of slots is always MaxProvidersToPair. So it wouldn't pick more

x/pairing/keeper/scores/pairing_slot.go
oren-lava: I changed GroupSlots to be deterministic without relying on a slot key. see https://github.com/lavanet/lava/commit/be6c5597c344f0ad20e94028ffee8928abae3543

x/pairing/keeper/scores/pairing_slot.go
oren-lava: I changed the implementation of GroupSlots to not using a key. see https://github.com/lavanet/lava/commit/be6c5597c344f0ad20e94028ffee8928abae3543

x/pairing/keeper/scores/score.go
oren-lava: I changed the implementation of GroupSlots to not using a key. see https://github.com/lavanet/lava/commit/be6c5597c344f0ad20e94028ffee8928abae3543

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/c582dc8be1bcf91cd2f1fab5c9df67a994e7ec94

x/pairing/keeper/scores/score.go
oren-lava: done (using normal panic as stated before). see https://github.com/lavanet/lava/commit/c582dc8be1bcf91cd2f1fab5c9df67a994e7ec94

x/pairing/keeper/scores/stake_req.go
```diff
@@ -0,0 +1,29 @@
...
+	return stakeEntry.Stake.Amount.Quo(sr.MinStake).ToDec().BigInt().Uint64()
+}
+
+func (sr StakeReq) GetName() string {
+	return stakeReqName
+}
+
+// Equal() used to compare slots to determine slot groups.
+// Usually Equal() compares the internal values of the ScoreReq,
+// but StakeReq returns always true because the stake is the score itself
```
orenl-lava, line 26: I think this history is irrelevant going forward,  you can replace with:
``` // Equal always returns true (StakeReq score of an entry is the entry's stake)```

File: x/pairing/keeper/scores/score.go [hunk 51]
```diff
@@ -0,0 +1,213 @@
...
+	return hashData
+}
+
+// PickProviders pick a <group-count> providers set with a pseudo-random weighted choice (using the providers' score list and hashData)
+func PickProviders(ctx sdk.Context, scores []*PairingScore, groupCount uint64, hashData []byte, chosenProvidersIdx *map[int]bool) (returnedProviders []epochstoragetypes.StakeEntry) {
+	if len(scores) == 0 {
+		return returnedProviders
+	}
+
+	scoreSum := sdk.NewUint(0)
```
orenl-lava, line 171: it would be more efficient to sum the scores as uin64, and then convert with NewUint().
you can also add FindSum() generic (in common/types/utils.go) for that.

orenl-lava: actually, why even use `sdk.Int` instead of `uint64` all through here?

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,213 @@
...
+		bigIntNum := new(big.Int).SetBytes(hash)
+		hashAsNumber := sdk.NewUintFromBigInt(bigIntNum)
+		modRes := hashAsNumber.Mod(scoreSum)
+
+		newScoreSum := sdk.NewUint(0)
+
+		for idx := len(scores) - 1; idx >= 0; idx-- {
+			providerScore := scores[idx]
+			if indexToSkip[idx] {
+				// this is an index we added
```
orenl-lava, line 192: ``` // skip index of providers already selected ```

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,213 @@
...
+			newScoreSum = newScoreSum.Add(sdk.NewUint(providerScore.Score))
+			if modRes.LT(newScoreSum) {
+				// we hit our chosen provider
+				returnedProviders = append(returnedProviders, *providerScore.Provider)
+				scoreSum = scoreSum.Sub(sdk.NewUint(providerScore.Score)) // we remove this provider from the random pool, so the sum is lower now
+				indexToSkip[idx] = true
+				break
+			}
+		}
+		if uint64(len(returnedProviders)) >= groupCount {
```
orenl-lava, line 204: make `groupcount` type be int and save the trouble of conversions

oren-lava: done. see https://github.com/lavanet/lava/commit/2cabf1d176e04902c5eb6e41490706fe0783b514

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/2cabf1d176e04902c5eb6e41490706fe0783b514

x/pairing/keeper/scores/stake_req.go
oren-lava: done. see https://github.com/lavanet/lava/commit/2cabf1d176e04902c5eb6e41490706fe0783b514

x/pairing/keeper/scores/score.go
orenl-lava: and I think this test is redundant?

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,211 @@
...
+	scoreSum := sdk.NewUint(0)
+	for _, providerScore := range scores {
+		scoreSum = scoreSum.Add(sdk.NewUint(providerScore.Score))
+	}
+	if scoreSum.IsZero() {
+		err := fmt.Errorf("score sum is zero. Cannot pick providers for pairing")
+		panic(err)
+	}
+
+	indexToSkip := *chosenProvidersIdx
```
orenl-lava, line 180: what is the purpose of passing `chosenProviderIdx` as pointer, if it get dereferenced here?
as is, local changes will not affect the original map, so may as well pass it not as pointer.

x/pairing/keeper/scores/score.go
orenl-lava: Go's native mod operation is `%`

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,211 @@
...
+	for it := 0; it < groupCount; it++ {
+		hash := tendermintcrypto.Sha256(hashData) // TODO: we use cheaper algo for speed
+		bigIntNum := new(big.Int).SetBytes(hash)
+		hashAsNumber := sdk.NewUintFromBigInt(bigIntNum)
+		modRes := hashAsNumber.Mod(scoreSum)
+
+		newScoreSum := sdk.NewUint(0)
+
+		for idx := len(scores) - 1; idx >= 0; idx-- {
+			providerScore := scores[idx]
```
orenl-lava, line 190: move this line to after the test of `indexToSkip`

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/1f805f2d22fff3c48b645d1671576f71bf4ea389

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/1f805f2d22fff3c48b645d1671576f71bf4ea389

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,211 @@
...
+			if modRes.LT(newScoreSum) {
+				// we hit our chosen provider
+				returnedProviders = append(returnedProviders, *providerScore.Provider)
+				scoreSum = scoreSum.Sub(sdk.NewUint(providerScore.Score)) // we remove this provider from the random pool, so the sum is lower now
+				indexToSkip[idx] = true
+				break
+			}
+		}
+
+		if scoreSum.IsZero() {
```
orenl-lava, line 204: unsure I understand when does this happen?   could it somehow go below zero?
does it mean that we may return with less providers than the `groupcount`?

oren-lava: It's a situation that should never happen (hence the panic). The scoreSum won't be zero because we're always adding non-negative numbers (and it's defined as an sdk.Uint)

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/5bdce0ff0c83e5d6f133414306b05605d746e1be

x/pairing/keeper/scores/score.go
oren-lava: yeah sorry, I thought you talked about the scoreSum zero check at the start of the function (so yeah, it's a `break`). Is there any needed change? If not, please resolve

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,211 @@
...
+
+// PickProviders pick a <group-count> providers set with a pseudo-random weighted choice (using the providers' score list and hashData)
+func PickProviders(ctx sdk.Context, scores []*PairingScore, groupCount int, hashData []byte, chosenProvidersIdx map[int]bool) (returnedProviders []epochstoragetypes.StakeEntry) {
+	if len(scores) == 0 {
+		return returnedProviders
+	}
+
+	var scoreSum uint64
+	for _, providerScore := range scores {
+		scoreSum += providerScore.Score
```
orenl-lava, line 173: if `chosenProviderIdx` could be non-empty - i.e with certain providers already marked as selected (and therefore unavailable to be chosen now), then those should be skipped when computing `scoreSum`, no?

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,211 @@
...
+			if modRes < newScoreSum {
+				// we hit our chosen provider
+				returnedProviders = append(returnedProviders, *providerScore.Provider)
+				scoreSum -= providerScore.Score // we remove this provider from the random pool, so the sum is lower now
+				chosenProvidersIdx[idx] = true
+				break
+			}
+		}
+
+		if scoreSum == 0 {
```
orenl-lava, line 205: I still think this is redundant:  IIUC, `scoreSum` will be zero when groupCount is exactly the number of providers _not yet_ selected, and in that case the loop will terminate anyhow.

x/pairing/keeper/scores/score.go
orenl-lava: I still think it's redundant - see comment on the newer code.

x/pairing/keeper/scores/score.go
oren-lava: in the case there are less providers than maxProvidersToPair, this scoreSum can be zero before ending the loop

x/pairing/keeper/scores/score.go
oren-lava: right. see https://github.com/lavanet/lava/commit/8f1730c2d24bee33c050e29ad4d903685ca5ed64

x/pairing/keeper/scores/score.go
orenl-lava: oh, right. 
maybe this can be made explicit by iterating on the min(groupCount, len(providers) - len(chosenProviderIdx)) instead?  checking for zero is an indirect test and therefore subtle and more prone to errors should something change later.

x/pairing/keeper/pairing.go
```diff
@@ -110,43 +109,74 @@ func (k Keeper) GetPairingForClient(ctx sdk.Context, chainID string, clientAddre
...
+	prevGroupSlot := pairingscores.NewPairingSlot() // init dummy slot to compare to
+	prevGroupSlot.Reqs = map[string]pairingscores.ScoreReq{}
+	indexToSkip := make(map[int]bool) // keep the indices of chosen providers to we won't pick the same providers twice (for different groups)
+	for _, group := range slotGroups {
+		diffSlot := group.Subtract(prevGroupSlot)
+		err := pairingscores.CalcPairingScore(providerScores, pairingscores.GetStrategy(), diffSlot, minStake)
+		if err != nil {
+			return nil, 0, "", err
+		}
+		hashData := pairingscores.PrepareHashData(project.Index, chainID, epochHash)
```
Yaroms, line 166: why is this calculated everytime, it is the same for each iteration

x/pairing/keeper/scores/pairing_slot.go
```diff
@@ -0,0 +1,43 @@
...
+
+func NewPairingSlot() *PairingSlot {
+	return &PairingSlot{Count: 1}
+}
+
+// Subtract generates a diff slot that contains the reqs that are in the slot receiver but not in the "other" slot
+func (s PairingSlot) Subtract(other *PairingSlot) *PairingSlot {
+	reqsDiff := make(map[string]ScoreReq)
+	for key := range s.Reqs {
+		if _, found := other.Reqs[key]; !found {
```
Yaroms, line 18: fetch the req here instead of fetching it 2 lines below again

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,216 @@
...
+			continue
+		}
+		scoreSum += providerScore.Score
+	}
+	if scoreSum == 0 {
+		err := fmt.Errorf("score sum is zero. Cannot pick providers for pairing")
+		panic(err)
+	}
+
+	iterationNum := commontypes.FindMin([]int{groupCount, len(scores) - len(chosenProvidersIdx)})
```
Yaroms, line 187: instead of this you can do:
if groupCount >= len(scores) - len(chosenProvidersIdx) //this means we will want every provider so no need to calculate
    return all providers

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,216 @@
...
+
+var (
+	uniformStrategy ScoreStrategy
+	allReqNames     []string
+)
+
+// TODO: currently we'll use weight=1 for all reqs. In the future, we'll get it from policy
+func init() {
+	// gather all req names to a list
+	allReqNames = append(allReqNames, stakeReqName)
```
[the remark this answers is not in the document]
oren-lava: but I want this variable to be available outside of the init() function. So I declare it above and append. In your way, its scope will be init()

x/pairing/keeper/scores/score.go
oren-lava: I can do this, but I'll need to write another loop (because I've got the scores and not providers. So I'll need to create a new array and append providers from scores). What's wrong with the current code?

x/pairing/keeper/pairing.go
oren-lava: done. see https://github.com/lavanet/lava/commit/cd637d649bba13af1a7a698952aa92539cd49074

x/pairing/keeper/scores/pairing_slot.go
Yaroms: not what i meant,
otherValue , found := other.Reqs[key]
than you can use it in the else if

x/pairing/keeper/scores/score.go
Yaroms: you will do computations of scoring and choosing for nothing instead of returning all of them

x/pairing/keeper/scores/score.go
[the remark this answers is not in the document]
Yaroms: why would the scope be within init?
the variable is declared outside of it, just looks weird appending an empty list instead of initializing it with the values.
this is not a must, just my opinion

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/6ba55d5846fd3f875247146f357d52c5249e7d4a

x/pairing/keeper/scores/pairing_slot.go
oren-lava: done. see https://github.com/lavanet/lava/commit/6ba55d5846fd3f875247146f357d52c5249e7d4a

x/pairing/keeper/scores/score.go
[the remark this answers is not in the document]
oren-lava: I thought I saw allReqName := ... (with ":"). see https://github.com/lavanet/lava/commit/6ba55d5846fd3f875247146f357d52c5249e7d4a

Yaroms (approved):

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,222 @@
...
+
+var (
+	uniformStrategy ScoreStrategy
+	allReqNames     []string
+)
+
+// TODO: currently we'll use weight=1 for all reqs. In the future, we'll get it from policy
+func init() {
+	// gather all req names to a list
+	allReqNames = []string{stakeReqName}
```
omerlavanet, line 58: can we skip one of them? also we can define var with initial values like: uniformStrategy = ScoreStrategy{stakeReqName:1}

x/pairing/keeper/scores/score.go
omerlavanet: you can define them staticly and only check length, it's going to be ocmpiled this way so it's more memory efficient and less code to read

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,222 @@
...
+// TODO: this function should be changed in the future since it only supports stake reqs
+func CalcSlots(policy planstypes.Policy, minStake sdk.Int) []*PairingSlot {
+	// init slot array (should be as the number of providers to pair)
+	slots := make([]*PairingSlot, policy.MaxProvidersToPair)
+
+	// all slots should consider the stake, so we init them with stakeReq
+	stakeReq := StakeReq{MinStake: minStake}
+	stakeReqName := stakeReq.GetName()
+
+	for i := range slots {
```
omerlavanet, line 81: this code shouldnt be inside score, it should be inside stakeReq and be called with an interface same as we do with filter

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,222 @@
...
+
+// group the slots
+func GroupSlots(slots []*PairingSlot) []*PairingSlot {
+	uniqueSlots := []*PairingSlot{}
+
+	if len(slots) == 0 {
+		panic("no pairing slots available")
+	}
+
+	uniqueSlots = append(uniqueSlots, slots[0])
```
omerlavanet, line 99: why do we need the special case for 0? we can start with an empty uniqueSlots array

common/types/utils.go
```diff
@@ -63,3 +64,12 @@ func Intersection[T comparable](arrays ...[]T) []T {
 
 	return intersection
 }
+
+// SafePow implements a deterministic power function
+// Go's math.Pow() doesn't guarantee determinism when executed on different hardwares
+func SafePow(base uint64, exponent uint64) uint64 {
+	baseInt := sdk.NewIntFromUint64(base)
+	baseDec := sdk.NewDecFromInt(baseInt)
```
omerlavanet, line 72: why not use NewDecWithPrec? we have examples how to creade Dec from uint without creating an Int

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,222 @@
...
+
+// PickProviders pick a <group-count> providers set with a pseudo-random weighted choice (using the providers' score list and hashData)
+func PickProviders(ctx sdk.Context, scores []*PairingScore, groupCount int, hashData []byte, chosenProvidersIdx map[int]bool) (returnedProviders []epochstoragetypes.StakeEntry) {
+	if len(scores) == 0 {
+		return returnedProviders
+	}
+
+	var scoreSum uint64
+	for idx, providerScore := range scores {
+		if chosenProvidersIdx[idx] {
```
omerlavanet, line 176: this implementation uses extra overhead and keeps updating these providers scores even if they are no longer relevant
we can save this extra memory and overhead, by setting an empryScoreObject inside the slots of already chosen providers, with score 0, this way we will also not update them when updating score

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,222 @@
...
+	if groupCount >= len(scores)-len(chosenProvidersIdx) {
+		providers := []epochstoragetypes.StakeEntry{}
+		for _, score := range scores {
+			providers = append(providers, *score.Provider)
+		}
+		return providers
+	}
+
+	for it := 0; it < groupCount; it++ {
+		hash := tendermintcrypto.Sha256(hashData) // TODO: we use cheaper algo for speed
```
omerlavanet, line 196: bug: you are using the same hashData for different groups resulting in the same number, this is not pseudo random and not fail, you need a differentiator in hashData for each slot
in the previous algo we did everything in one loop and the loop iterator was the differentiator but since your differentiator is starting form 0 all the time it's a mistake
i want a test that verifies pairing is uniformly sitributed pseudo random by scores

x/pairing/keeper/scores/stake_req.go
```diff
@@ -0,0 +1,28 @@
...
+const stakeReqName = "stake-req"
+
+// StakeReq implements the ScoreReq interface for provider staking requirement(s)
+type StakeReq struct {
+	MinStake sdk.Int
+}
+
+// Score calculates the the provider score as the normalized stake
+func (sr StakeReq) Score(stakeEntry epochstoragetypes.StakeEntry) uint64 {
+	return stakeEntry.Stake.Amount.Quo(sr.MinStake).Uint64()
```
omerlavanet, line 17: we said to normalize by minStake/16 so we have a more refined quantization, currently:
5000000 and 999999 will have the same chance

x/pairing/keeper/pairing.go
```diff
@@ -165,19 +196,26 @@ func (k Keeper) getProjectStrictestPolicy(ctx sdk.Context, project projectstypes
...
-		return planstypes.Policy{}, 0, fmt.Errorf("chain ID not found in any of the policies")
+		return planstypes.Policy{}, 0, sdk.Int{}, fmt.Errorf("chain ID not found in any of the policies")
 	}
 
+	spec, found := k.specKeeper.GetSpec(ctx, chainID)
+	if !found {
+		return planstypes.Policy{}, 0, sdk.Int{}, fmt.Errorf("spec not found with chain ID")
+	}
+
+	minStake := spec.MinStakeProvider.Amount
```
omerlavanet, line 207: does not belong to this function, and other places use it, remove it form here to somewhere else, also you are reading the spec already in different places you can save on reading the spec (expensive), specFoundAndActive reads it already

x/pairing/keeper/pairing.go
```diff
@@ -116,43 +115,75 @@ func (k Keeper) GetPairingForClient(ctx sdk.Context, chainID string, clientAddre
...
+	for i := range stakeEntries {
+		providerScore := pairingscores.NewPairingScore(&stakeEntries[i])
+		providerScores = append(providerScores, providerScore)
+	}
+
+	// calculate score (always on the diff in score components of consecutive groups) and pick providers
+	prevGroupSlot := pairingscores.NewPairingSlot() // init dummy slot to compare to
+	prevGroupSlot.Reqs = map[string]pairingscores.ScoreReq{}
+	indexToSkip := make(map[int]bool) // keep the indices of chosen providers to we won't pick the same providers twice (for different groups)
+	hashData := pairingscores.PrepareHashData(project.Index, chainID, epochHash)
```
omerlavanet, line 166: same issue in hashData being deterministic accross groups, solvable by utilizing group_index[separator]it in the hash data

omerlavanet (dismissed):

x/pairing/keeper/scores/score.go
oren-lava: not sure what you mean by skipping one of them. A few lines below I panic if `len(allReqNames) != len(uniformStrategy)`. Do you want me to fix that?
Furthermore, I know I can do `uniformStrategy = ScoreStrategy{stakeReqName:1}`, but I used a loop because there will be many types of reqs in the future

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/157c0df6080ba5ebb599d47c6873aca506b82b24

common/types/utils.go
oren-lava: done. see https://github.com/lavanet/lava/commit/157c0df6080ba5ebb599d47c6873aca506b82b24

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/157c0df6080ba5ebb599d47c6873aca506b82b24

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,222 @@
...
+}
+
+// CalcPairingScore calculates the final pairing score for a pairing slot (with strategy)
+// For efficiency purposes, we calculate the score on a diff slot which represents the diff reqs of the current slot
+// and the previous slot
+func CalcPairingScore(scores []*PairingScore, strategy ScoreStrategy, diffSlot *PairingSlot, minStake sdk.Int) error {
+	// calculate the score for each req for each provider
+	for _, req := range diffSlot.Reqs {
+		reqName := req.GetName()
+		weight := strategy[reqName] // not checking if key is found because it's verified in init()
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/157c0df6080ba5ebb599d47c6873aca506b82b24

x/pairing/keeper/scores/stake_req.go
oren-lava: done. see https://github.com/lavanet/lava/commit/fa97b594033898263bad056fd6fde790cca859c3

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,222 @@
...
+	for idx, providerScore := range scores {
+		if chosenProvidersIdx[idx] {
+			// skip index of providers already selected
+			continue
+		}
+		scoreSum += providerScore.Score
+	}
+	if scoreSum == 0 {
+		err := fmt.Errorf("score sum is zero. Cannot pick providers for pairing")
+		panic(err)
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/fa97b594033898263bad056fd6fde790cca859c3

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,222 @@
...
+
+	return nil
+}
+
+// PrepareHashData prepares the hash needed in the pseudo-random choice of providers
+func PrepareHashData(projectIndex string, chainID string, epochHash []byte) []byte {
+	hashData := []byte{}
+
+	// add the session start block hash to the function to make it as unpredictable as we can
+	hashData = append(hashData, epochHash...)
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/fa97b594033898263bad056fd6fde790cca859c3

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,222 @@
...
+			if newScoreComp == 0 {
+				err := fmt.Errorf("score component is zero. score component name: %s, provider address: %s", reqName, score.Provider.Address)
+				panic(err)
+			}
+			newScoreComp = commontypes.SafePow(newScoreComp, weight)
+
+			// divide by previous score component (if exists) and multiply by new score
+			prevReqScoreComp, ok := score.ScoreComponents[reqName]
+			if ok {
+				score.Score /= prevReqScoreComp
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/fa97b594033898263bad056fd6fde790cca859c3

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,222 @@
...
+	// calculate the score for each req for each provider
+	for _, req := range diffSlot.Reqs {
+		reqName := req.GetName()
+		weight := strategy[reqName] // not checking if key is found because it's verified in init()
+
+		for _, score := range scores {
+			newScoreComp := req.Score(*score.Provider)
+			if newScoreComp == 0 {
+				err := fmt.Errorf("score component is zero. score component name: %s, provider address: %s", reqName, score.Provider.Address)
+				panic(err)
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/fa97b594033898263bad056fd6fde790cca859c3

common/types/utils.go
oren-lava: corrected it. see https://github.com/lavanet/lava/commit/9fdaee7f74bb1c371aebabd65334ee396226e6f9

x/pairing/keeper/scores/score.go
```diff
@@ -0,0 +1,222 @@
...
+func GetStrategy() ScoreStrategy {
+	return uniformStrategy
+}
+
+// CalcPairingScore calculates the final pairing score for a pairing slot (with strategy)
+// For efficiency purposes, we calculate the score on a diff slot which represents the diff reqs of the current slot
+// and the previous slot
+func CalcPairingScore(scores []*PairingScore, strategy ScoreStrategy, diffSlot *PairingSlot, minStake sdk.Int) error {
+	// calculate the score for each req for each provider
+	for _, req := range diffSlot.Reqs {
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/9fdaee7f74bb1c371aebabd65334ee396226e6f9

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/b1037f110483c4f36f6492d622158d4ef67b42d1

x/pairing/keeper/pairing.go
oren-lava: not relevant anymore

x/pairing/keeper/scores/score.go
oren-lava: fixed

x/pairing/keeper/scores/score.go
oren-lava: not relevant anymore

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/9fdaee7f74bb1c371aebabd65334ee396226e6f9

x/pairing/keeper/scores/score.go
oren-lava: done. see https://github.com/lavanet/lava/commit/0e1e5bc0039df5af58310188c100a4cdb2383d05

x/pairing/keeper/pairing.go
oren-lava: done. see https://github.com/lavanet/lava/commit/0e1e5bc0039df5af58310188c100a4cdb2383d05

x/pairing/keeper/scores/score.go
omerlavanet: to be fixed

oren-lava: done. see https://github.com/lavanet/lava/commit/6a09f8a2ab74ad252278dc19090569bb1cd4d889

Yaroms (approved):

PR #595, opened 2023-07-05
PRT-730 adding better node error handling 

ranlavanet:


protocol/chainlib/tendermintRPC.go
```diff
@@ -569,9 +569,10 @@ func (cp *tendermintRpcChainProxy) SendRPC(ctx context.Context, nodeMessage *rpc
 		defer cancel()
 		// perform the rpc call
 		rpcMessage, err = rpc.CallContext(connectCtx, nodeMessage.ID, nodeMessage.Method, nodeMessage.Params)
-		if err != nil && connectCtx.Err() == context.DeadlineExceeded {
-			// Not an rpc error, return provider error without disclosing the endpoint address
-			return nil, "", nil, utils.LavaFormatError("Provider Failed Sending Message", context.DeadlineExceeded)
+		if err != nil {
+			if handledError := cp.BaseChainProxy.ErrorHandler.HandleTendermintRPCErrors(connectCtx, err); handledError != nil {
```
omerlavanet, line 573: do an interface HandleErrors, and set the right object in thhe baseChainProxy then you call the same function in all api interfaces, the HandleXXXErrors is ugly

protocol/chainlib/node_error_handler.go
```diff
@@ -0,0 +1,72 @@
...
+		return utils.LavaFormatError("Provider Side Failed Sending Message, Reason: DNS resolution failed", nil)
+	} else if opErr, ok := err.(*net.OpError); ok {
+		if sysErr, ok := opErr.Err.(*os.SyscallError); ok && sysErr.Err == syscall.ECONNREFUSED {
+			return utils.LavaFormatError("Provider Side Failed Sending Message, Reason: Connection refused", nil)
+		}
+	}
+	return nil // Return original error if it doesn't match any specific cases
+}
+
+func (neh *NodeErrorHandler) HandleGenericErrors(ctx context.Context, nodeError error) error {
```
omerlavanet, line 34: create 4 structs implementing the same interface and use the same function

protocol/chainlib/rest.go
```diff
@@ -389,11 +389,9 @@ func (rcp *RestChainProxy) SendNodeMsg(ctx context.Context, ch chan interface{},
 
 	res, err := httpClient.Do(req)
 	if err != nil {
-		if err != nil && connectCtx.Err() == context.DeadlineExceeded {
-			// Not an rpc error, return provider error without disclosing the endpoint address
-			return nil, "", nil, utils.LavaFormatError("Provider Failed Sending Message", context.DeadlineExceeded)
+		if handledError := rcp.BaseChainProxy.ErrorHandler.HandleRestErrors(connectCtx, err); handledError != nil {
```
omerlavanet, line 392: the new code has a path where it doesn't return an error, the old code doesn't it returns an error from this if case in both cases, is this intentional?

protocol/chainlib/node_error_handler.go
```diff
@@ -0,0 +1,72 @@
...
+	} else if opErr, ok := err.(*net.OpError); ok && opErr.Timeout() {
+		return utils.LavaFormatError("Provider Side Failed Sending Message, Reason: Network operation timed out", nil)
+	} else if _, ok := err.(*net.DNSError); ok {
+		return utils.LavaFormatError("Provider Side Failed Sending Message, Reason: DNS resolution failed", nil)
+	} else if opErr, ok := err.(*net.OpError); ok {
+		if sysErr, ok := opErr.Err.(*os.SyscallError); ok && sysErr.Err == syscall.ECONNREFUSED {
+			return utils.LavaFormatError("Provider Side Failed Sending Message, Reason: Connection refused", nil)
+		}
+	}
+	return nil // Return original error if it doesn't match any specific cases
```
omerlavanet, line 31: this doesn't return the original error

omerlavanet (changes requested):
need a refactor,
also the return on error flow is different from expected, returning nil when there is an error

PR #603, opened 2023-07-09
CNS-443 propagate spec change to descendants

orenl-lava:
When a Spec gets updated, we set its BlockLastUpdated field to the current
block height (useful for producers to know whether they use the latest Spec
version).
    
If there are other Specs that inherit this Spec, then they effectively also
changed - so we hunt them and set their BlockLastUpdated field too.

### DIFF
--- a/proto/spec/api_collection.proto
+++ b/proto/spec/api_collection.proto
@@ -11,28 +11,28 @@ message ApiCollection {
   CollectionData collection_data =2 [(gogoproto.nullable) = false];
   repeated Api apis = 3;
   repeated Header headers = 4;
-	repeated CollectionData inheritance_apis = 5; // by collectionKey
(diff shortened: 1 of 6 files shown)


omerlavanet (commented):
are we aware this whole code is different in the fork branch? how about we rebase from there?

orenl-lava:
The changes are quite simple where the expansion/recursion occurs.
You can also back-port (or forward-port) to that branch :)

omerlavanet:
since we are not deploying another version before the fork, we can start doing it as part of this task

orenl-lava:
@omer, @Yaroms  - ok, done, and (force) pushed here.

Yaroms (approved):

Yaroms (approved):

PR #606, opened 2023-07-10
PRT-753-add-headers-for-provider-generated-queries

omerlavanet:


protocol/chainlib/chain_fetcher.go
```diff
@@ -44,6 +45,13 @@ func (cf *ChainFetcher) Validate(ctx context.Context) error {
 	return nil
 }
 
+func (cf *ChainFetcher) ChainFetcherMetadata() []pairingtypes.Metadata {
+	ret := []pairingtypes.Metadata{
```
ranlavanet, line 49: won't it be better if we store it in the struct and read if instead of every time allocating and accessing it?

ranlavanet (approved):

PR #612, opened 2023-07-12
GROW-942 Setup logging level on badges

tabakuj:


protocol/badgegenerator/badgeserver.go
```diff
@@ -99,7 +99,12 @@ func RunBadgeServer(cmd *cobra.Command, v *viper.Viper) {
 	if err != nil {
 		utils.LavaFormatFatal("Error in open listener", err)
 	}
-	// set up the grpc server
+
+	logLevel, err := cmd.Flags().GetString(flags.FlagLogLevel)
```
ranlavanet, line 103: you should set it in the RunE: command

ranlavanet: @tabakuj

ranlavanet (approved):

PR #626, opened 2023-07-17
CNS-514: Improve policy setting

oren-lava:
This PR includes two new features for policy setting (using a YAML file):
1. support for geo enum: you can write "USE" or "EU" instead of number in the YAML file
2. you can omit some of the fields in the policy file and it'll still be ok (to let users define policies with some of the fields, for convenience)


common/types/yamlUtils.go
```diff
@@ -19,25 +21,114 @@ func ReadYaml(filePath string, primaryKey string, content interface{}, hooks []m
...
+func findMissingFields(content interface{}) []string {
+	var missingFields []string
+
+	expectedFields := make(map[string]bool)
+	contentValue := reflect.ValueOf(content)
+	contentType := contentValue.Type()
+
+	// Ensure content is a struct or a pointer to a struct
+	if contentType.Kind() != reflect.Struct && (contentType.Kind() != reflect.Ptr || contentType.Elem().Kind() != reflect.Struct) {
+		return missingFields
```
orenl-lava, line 58: if calling `findMissingFields()` with non-struct is a serious bug, it needs to be communicated somehow - panic, or log, or return error.

common/types/yamlUtils.go
```diff
@@ -19,25 +21,114 @@ func ReadYaml(filePath string, primaryKey string, content interface{}, hooks []m
...
+	for i := 0; i < contentType.NumField(); i++ {
+		field := contentType.Field(i)
+		fieldName := field.Name
+		expectedFields[fieldName] = true
+	}
+
+	// Check if each expected field is present in the unmarshaled content
+	for fieldName := range expectedFields {
+		fieldValue := contentValue.FieldByName(fieldName)
+		if !fieldValue.IsValid() || fieldValue.IsZero() {
```
orenl-lava, line 77: This can be tricky: what if the fieldValue is legitimately zero, in the input?

x/plans/types/policy.go
```diff
@@ -4,13 +4,24 @@ import (
...
 
+var policyDefaultValues map[string]interface{}
+
+// init policy default values (for fields that their natural zero value is not good)
+// the values were chosen in a way that they will not influence the strictest policy calculation
+func init() {
+	policyDefaultValues = make(map[string]interface{})
+
+	policyDefaultValues["GeolocationProfile"] = uint64(Geolocation_value["GL"])
+	policyDefaultValues["MaxProvidersToPair"] = uint64(math.MaxUint64)
```
orenl-lava, line 22: you can define these without needing `init()` function:
```
    var policyDefaultValues = map[string]interface{}{
        "GeolocationProfile":   uint64(Geolocation_value["GL"]),
	"MaxProvidersToPair": uint64(math.MaxUint64),
    }

**Review by orenl-lava — CHANGES_REQUESTED** [review]

File: x/pairing/keeper/pairing.go [hunk 4]
```diff
@@ -246,6 +246,9 @@ func (k Keeper) CalculateEffectiveAllowedCuPerEpochFromPolicies(policies []*plan
 	var policyTotalCuLimit []uint64
 	for _, policy := range policies {
 		if policy != nil {
+			if policy.EpochCuLimit == 0 || policy.TotalCuLimit == 0 {
```
Yaroms, line 249: should be 2 if cases, one for each limit

common/types/yamlUtils.go
oren-lava: I'll open a tech debt ticket

x/pairing/keeper/pairing.go
oren-lava: done. see https://github.com/lavanet/lava/commit/a85725d29f3c84baa38732c19018a9970ae7bca3

common/types/yamlUtils.go
orenl-lava: specifically, the problem is that this can't distinguish an implicit zero value (due to missing field) from an explicit zero value. Therefore, if a specific field:
* _may_ have the zero value, -and-
* has its default value set to non-zero value
then it would be impossible to set the field in the yaml config file explicitly to zero value, because it will always be interpreted as missing (since the code assumes "missing if-and-only-if zero"), and hence always be overwritten with the default (non-zero) value.

to solve this we need a reliable method to find really missing fields (as opposed to present-zero-values). one way to do that is to use viper to also read the config into a map (of things). then, while scanning through the struct looking into its fields, replace the current test with a test that looks up the field in the map - if not found there then it was missing from the yaml config file.

x/pairing/keeper/pairing.go
```diff
@@ -249,7 +249,10 @@ func (k Keeper) CalculateEffectiveAllowedCuPerEpochFromPolicies(policies []*plan
 	var policyTotalCuLimit []uint64
 	for _, policy := range policies {
 		if policy != nil {
-			if policy.EpochCuLimit == 0 || policy.TotalCuLimit == 0 {
+			if policy.EpochCuLimit == 0 {
```
Yaroms, line 252: if policy.EpochCuLimit != 0
   policyEpochCuLimit = append(policyEpochCuLimit, policy.GetEpochCuLimit())

the same for the other limit

x/plans/types/policy.go
```diff
@@ -9,17 +9,16 @@ import (
...
 
 // init policy default values (for fields that their natural zero value is not good)
 // the values were chosen in a way that they will not influence the strictest policy calculation
-func init() {
-	policyDefaultValues = make(map[string]interface{})
-
-	policyDefaultValues["GeolocationProfile"] = uint64(Geolocation_value["GL"])
-	policyDefaultValues["MaxProvidersToPair"] = uint64(math.MaxUint64)
+var policyDefaultValues = map[string]interface{}{
+	"GeolocationProfile": uint64(Geolocation_value["GL"]),
```
Yaroms, line 20: use enum instead of "GL"

common/types/yamlUtils.go
orenl-lava: please mark the panic as "ok".

oren-lava: done. see https://github.com/lavanet/lava/commit/8610fb0f6e29931a011ce961b80ccd7fba66ad98

Yaroms (approved):

PR #633, opened 2023-07-20
[PRT-786] Validate chain id for all rpc interfaces

Aleksao998:
## Description

This PR fixes a bug where we are not validating chainId for all interfaces, but only till first valid

## Test Cases

```
screen -d -m -S provider1 bash -c "source ~/.bashrc; lava-protocol rpcprovider \
$PROVIDER1_LISTENER COS3 rest '$OSMOT_REST' \
$PROVIDER1_LISTENER COS3 tendermintrpc '$OSMO_RPC,$OSMO_RPC' \
$PROVIDER1_LISTENER COS3 grpc '$OSMOT_GRPC' \
$EXTRA_PROVIDER_FLAGS --metrics-listen-address ":7780" --geolocation 1 --log_level debug --from servicer1 2>&1 | tee $LOGS_DIR/PROVIDER1.log" && sleep 0.25
```

```
Jul 20 08:41:51 ERR RPCProvider running with disabled endpoints:
	COS3:rest Network Address:127.0.0.1:2221 Node: https://bekotopopo2:99al9YwXTnvJohMW39ZBqVH5AWnWBDV8@prod-pnet-osmosisnode-1.lavapro.xyz/testnet/rest/ Geolocation:1 Addons:
COS3:grpc Network Address:127.0.0.1:2221 Node: prod-pnet-osmosisnode-1.lavapro.xyz:9092 Geolocation:1 Addons:
```

Also i tried every other combination of these 3 rpc interfaces with correct and wrong node urls

### DIFF
--- a/protocol/rpcprovider/rpcprovider.go
+++ b/protocol/rpcprovider/rpcprovider.go
@@ -166,11 +166,15 @@ func (rpcp *RPCProvider) Start(ctx context.Context, txFactory tx.Factory, client
 				chainMutexes[chainID].Lock()
 				defer chainMutexes[chainID].Unlock()
 
-				// Create temp chain fetcher so we can validate chainID
-				tempChainFetcher := chainlib.NewChainFetcher(ctx, chainRouter, chainParser, rpcProviderEndpoint)
+				var chainFetcher chainlib.ChainFetcherIf
+				if enabled, _ := chainParser.DataReliabilityParams(); enabled {
+					chainFetcher = chainlib.NewChainFetcher(ctx, chainRouter, chainParser, rpcProviderEndpoint)
+				} else {
+					chainFetcher = chainlib.NewDummyChainFetcher(ctx, rpcProviderEndpoint)
+				}
 
 				// Fetch and validate chain id
-				err = tempChainFetcher.Validate(ctx)
+				err = chainFetcher.Validate(ctx)
 				if err != nil {
 					return utils.LavaFormatError("panic severity critical error, aborting support for chain api due to failing to fetch chain ID, continuing with other endpoints", err, utils.Attribute{Key: "endpoint", Value: rpcProviderEndpoint})
 				}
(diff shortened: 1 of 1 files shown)


protocol/rpcprovider/rpcprovider.go
```diff
@@ -165,6 +165,16 @@ func (rpcp *RPCProvider) Start(ctx context.Context, txFactory tx.Factory, client
 			chainCommonSetup := func() error {
 				chainMutexes[chainID].Lock()
 				defer chainMutexes[chainID].Unlock()
+
+				// Create temp chain fetcher so we can validate chainID
+				tempChainFetcher := chainlib.NewChainFetcher(ctx, chainRouter, chainParser, rpcProviderEndpoint)
```
omerlavanet, line 170: we are creating the same chainFetcher twice on the first apiInterface both here and on 189, we can use the one created here instead when storing
so no need to call it temp as we save one of them

omerlavanet, line 170: we are creating the same chainFetcher twice on the first apiInterface both here and on 189, we can use the one created here instead when storing
so no need to call it temp as we save one of them

omerlavanet (changes requested):

protocol/rpcprovider/rpcprovider.go
Aleksao998: Fixed 506689a333e43712a67a17038e9a554f6742eda3

protocol/rpcprovider/rpcprovider.go
```diff
@@ -166,11 +166,15 @@ func (rpcp *RPCProvider) Start(ctx context.Context, txFactory tx.Factory, client
 				chainMutexes[chainID].Lock()
 				defer chainMutexes[chainID].Unlock()
 
-				// Create temp chain fetcher so we can validate chainID
-				tempChainFetcher := chainlib.NewChainFetcher(ctx, chainRouter, chainParser, rpcProviderEndpoint)
+				var chainFetcher chainlib.ChainFetcherIf
+				if enabled, _ := chainParser.DataReliabilityParams(); enabled {
+					chainFetcher = chainlib.NewChainFetcher(ctx, chainRouter, chainParser, rpcProviderEndpoint)
+				} else {
+					chainFetcher = chainlib.NewDummyChainFetcher(ctx, rpcProviderEndpoint)
```
omerlavanet, line 173: problem when we have data reliability disabled but want to validate chainID

omerlavanet (changes requested):

omerlavanet (approved):

PR #637, opened 2023-07-23
CNS-503: cleanup unit tests (part 2)

orenl-lava:


testutil/common/tester.go
```diff
@@ -57,36 +62,89 @@ func NewTester(t *testing.T) *Tester {
...
+	}
+	sort.Strings(names)
+	var accounts []Account
+	for _, k := range names {
+		accounts = append(accounts, ts.accounts[k])
+	}
+	return accounts
+}
+
+func (ts *Tester) StakeAccount(acct Account, spec spectypes.Spec, amount int64, geoloc uint64, moniker string) error {
```
Yaroms, line 120: is this used?
suppose to be deprecated

Yaroms (dismissed):

testutil/common/tester.go
orenl-lava: this stakes a provider, given its common.Account.
similar to StakeProvider(), except the latter takes the string address, i.e. acc.Addr.String()
it gets some clean-up in the PR of part 5; but I can take a step further and replace with StakeProvider where used.

orenl-lava: fixed 4d40a32c

Yaroms (approved):

PR #642, opened 2023-07-23
CNS-489: fixation handle last put of latest entry

orenl-lava:
Every `PutEntry()` must match a prior `GetEntry()`, and an extraneous call to `PutEntry()` is a bug - detected by refcount being already zero. However, if the extraneous call targets a latest entry, we won't detect it easily because the refcount of the latest entry is already elevated by one. 

This PR adds a new field - "entry.IsLatest" - that marks the latest entry (and only the latest entry). This allows `PutEntry()` to test whether it is about to commit a serious error - and if so abort/skip (to avoid panic). 

The new field also allows us to add a test (that was missing before) to prevent retroactive `AppendEntry()` of entries _before_ an existing latest entry. (Retroactive `AppendEntry()` are valid as long as they do not sneak in before an existing LatestEntry).


oren-lava (approved):

common/fixation_entry.go
```diff
@@ -252,10 +252,14 @@ func (fs *FixationStore) AppendEntry(
 	latestEntry, found := fs.getUnmarshaledEntryForBlock(ctx, safeIndex, block)
 
 	var deleteAt uint64 = math.MaxUint64
+	var latest bool
 
 	// if latest entry is not found, this is a first version entry
 	if !found {
 		fs.setEntryIndex(ctx, safeIndex, true)
+		if block <= ctxBlock {
```
Yaroms, line 260: this if is happening 3 times in this code, here, at 281 and 320.
can just set latest = block <= ctxBlock at the beginning and use it

x/plans/keeper/migrations.go
```diff
@@ -91,3 +91,9 @@ func (m Migrator) Migrate4to5(ctx sdk.Context) error {
 func (m Migrator) Migrate5to6(ctx sdk.Context) error {
 	return m.keeper.plansFS.MigrateVersionFrom(ctx, 3)
 }
+
+// Migrate6to7 implements store migration from v6 to v7:
+// -- trigger fixation migration (v4->v5), initialize IsLatest field
+func (m Migrator) Migrate6to7(ctx sdk.Context) error {
+	return m.keeper.plansFS.MigrateVersionFrom(ctx, 4)
```
Yaroms, line 98: why do we use the migrateversionfrom method instead of the general one?

common/fixation_entry.go
orenl-lava: I'm unsure it improves readability or performance in a noticeable way. Anyway, the test is not for "latest" but rather for "(not) future entry". I'll consider this change as I work on allow append-after-delete (CNS-496).

x/plans/keeper/migrations.go
orenl-lava: git-blame suggests you introduces the function MigrateVersionFrom() and its use in `x/plans/keeper/migrations.go:Migrate5to6() - which I used as base. I'm also curious why indeed?

common/types/keys.go
```diff
@@ -31,7 +31,7 @@ const (
 )
 
 const (
-	TimerVersionKey = "Entry_Version"
```
Yaroms, line 34: this needs migration?

orenl-lava: can't do migration, because the timer version is mixed up with that of the fixation store that owns it (if any). Instead, will add logic on the next getVersion() call, to test and if not initialize then initialize ad-hoc. the call to getVersion() would happen naturally upon next upgrade as the logic needs to decided whether migration is needed at all.

Yaroms (approved):

PR #645, opened 2023-07-24
GROW-899 add createKey static method to lavasdk

menash6:


ecosystem/lava-sdk/src/util/create-key.ts
```diff
@@ -0,0 +1,100 @@
...
+  publicKey: string,
+  apiAccessKey: string
+) {
+  const keyParams: DeveloperKeyParams = {
+    projectKey: publicKey,
+    name: `SDK Generated Key-${uuid().toLowerCase()}`,
+  };
+  return await axiosInstance.post(`/developer-keys/`, keyParams, {
+    headers: { "api-access-key": apiAccessKey },
+  });
```
nirtayeb, line 95: why do use axios instead of fetch?
it's going to increase the build size for nothing

ecosystem/lava-sdk/src/util/create-key.ts
```diff
@@ -0,0 +1,100 @@
...
+}
+
+export function timeout(delayMS: number) {
```
nirtayeb, line 98: sleep is a better name

ecosystem/lava-sdk/src/util/create-key.ts
```diff
@@ -0,0 +1,100 @@
...
+  const { data } = await axiosInstance.get(`/developer-keys`, {
+    headers: {
+      "api-access-key": apiAccessKey,
+    },
+  });
+  return data;
+}
+
+export async function createDeveloperKey(
+  publicKey: string,
```
nirtayeb, line 86: developerKey... make it consistent with other arguments
also, change the order to be consistent with getKey.

ecosystem/lava-sdk/src/util/create-key.ts
```diff
@@ -0,0 +1,100 @@
...
+  privateHex: string;
+  mnemonicChecked: EnglishMnemonic;
+};
+
+export type DeveloperKeyParams = {
+  projectKey: string;
+  name: string;
+};
+
+const GW_BACKEND_URL = "https://gateway-master.lava-cybertron.xyz/sdk";
```
nirtayeb, line 38: instead of constant lets pass it as an optional parameter to each function.
default value should be production url.

ecosystem/lava-sdk/src/util/create-key.ts
menash6: I did discuss using axios with @Aleksao998 before using it - of course I can revert back to fetch

PR #652, opened 2023-07-27
CNS-516: fix pairing choice mechanism small bias

oren-lava:
Before this PR, the pairing mechanism had a small bias in the provider choice mechanism. This PR fixes this issue.
I've also added two unit tests that check that the choice is weighted by the score correctly. Note that in those unit tests, the error margin is quite large so when I put a small number of choice iterations (so the tests won't take very long), they'll pass.
If you wish to test that the choice algorithm chooses correctly, change the number of iterations and error margin locally.


x/pairing/keeper/scores/pairing_score.go
```diff
@@ -12,18 +12,17 @@ const (
 
 // PairingScore holds a provider's score with respect to a set of requirements (ScoreReq), indexed by their unique name.
 type PairingScore struct {
-	Provider          *epochstoragetypes.StakeEntry
-	Score             sdk.Uint
-	ScoreComponents   map[string]sdk.Uint
-	ValidForSelection bool
```
Yaroms, line 18: make sure this is false

x/pairing/keeper/pairing_test.go
```diff
@@ -776,6 +776,133 @@ func TestSelectedProvidersPairing(t *testing.T) {
...
+
+	// Calculate the expected count for each provider (should be nearly equal for uniform distribution)
+	expectedCount := (numIterations * providersToPair) / providersCount
+
+	// Define a margin of error for the count (you can adjust this based on your tolerance)
+	marginOfError := int(math.Round(0.1 * float64(expectedCount)))
+
+	// Check that the count for each provider is within the margin of error of the expected count
+	for addr, count := range providerCount {
+		if count < expectedCount-marginOfError || count > expectedCount+marginOfError {
```
Yaroms, line 817: just use abs(count - expectedCount) > marginOfError

oren-lava: done. see https://github.com/lavanet/lava/commit/704a102dd13eb74a97f326646ef4f4d1858f239d

Yaroms (approved):

x/pairing/keeper/scores/score.go
```diff
@@ -188,30 +189,34 @@ func PickProviders(ctx sdk.Context, scores []*PairingScore, groupCount int, hash
...
-		uintHash := sdk.NewUintFromBigInt(bigIntHash)
-		modRes := uintHash.Mod(scoreSum)
+	sum256 := sha256.Sum256(hashData)
+	// Fold the SHA-256 hash into a 64-bit seed using bitwise XOR
+	var seed int64
+	for i := 0; i < len(sum256)/8; i++ {
+		seed ^= int64(binary.BigEndian.Uint64(sum256[i*8 : (i+1)*8]))
+	}
 
+	rand.Seed(seed)
```
orenl-lava, line 199: using the global `rand` directly is not safe in terms of determinism, e.g. if another go routine calls (now or in the future) it to get a random number at the wrong time. also, `rand.Seed` is deprecated in Go (for similar reasons). Instead, we should instantiate a dedicated instance using `rand.New()` - see below.

Also, Go does not guarantee that the PRNG driving the default source for `rand` will always provide the same output for a given seed - across different Go versions. Instead, we should provide our own source (PRNG) that will always be under our control. The [xorshift package](https://github.com/vpxyz/xorshift) offers a selection of such PRNGs. See [Xorshift wiki page](https://en.wikipedia.org/wiki/Xorshift) for details on the different options.

instead, we should instantiate a dedicated instance e.g.
```
   source := xorshift.XoroShiro128StarStar,NewSource(seed)
   rng := rand.New(source)
   ...
   value := rng.UInt63n(...)
```

orenl-lava, line 199: This should go into a helper function/submodule in e.g. `utils/rand`, e.g. something like:
```
    New(seed) *math.Rand  // returns a new RPNG instance properly seeded
    Seed(*math.Rand, seed) *math.Rand  // re-seeds an existing RPNG instance
```

And it should be accompanied with tests to verify uniform distribution, e.g. can be copied from here:
https://qqq.ninja/blog/post/fast-threadsafe-randomness-in-go/#uniform-distribution-test

x/pairing/keeper/pairing_test.go
```diff
@@ -776,6 +777,136 @@ func TestSelectedProvidersPairing(t *testing.T) {
...
+	expectedCount := (numIterations * providersToPair) / providersCount
+
+	// Define a margin of error for the count (you can adjust this based on your tolerance)
+	marginOfError := math.Round(0.1 * float64(expectedCount))
+
+	// Check that the count for each provider is within the margin of error of the expected count
+	for addr, count := range providerCount {
+		if math.Abs(float64(count-expectedCount)) > marginOfError {
+			t.Errorf("Provider with address %s was not picked with the expected weight: count = %d, expected = %d",
+				addr, count, expectedCount)
```
orenl-lava, line 820:
```
    require.InEpsilon(t, count, expectedCount, 0.10) // ensure values are within 10% error
```

x/pairing/keeper/pairing_test.go
```diff
@@ -776,6 +777,136 @@ func TestSelectedProvidersPairing(t *testing.T) {
...
+	for addr, info := range providerCount {
+		// Calculate the expected count based on the provider's stake
+		expectedCount := providersToPair * (numIterations * int(info.stakeAmount)) / int(totalStakes)
+
+		fmt.Printf("count: %d, expected: %d\n", info.count, expectedCount)
+
+		// Define a margin of error for the count (you can adjust this based on your tolerance)
+		marginOfError := math.Round(0.15 * float64(expectedCount))
+
+		if math.Abs(float64(info.count-expectedCount)) > marginOfError {
```
orenl-lava, line 903: Use `require.InEpsilon()`

x/pairing/keeper/pairing_test.go
```diff
@@ -776,6 +777,136 @@ func TestSelectedProvidersPairing(t *testing.T) {
 	}
 }
 
+// Test that the pairing process picks identical providers uniformly
+func TestPairingUniformDistribution(t *testing.T) {
```
orenl-lava, line 781: this test is identical to the next one - except for the extra staking. 
you can refactor the loop part out and both tests would use it, instead of duplication.

x/pairing/keeper/pairing_test.go
```diff
@@ -776,6 +777,136 @@ func TestSelectedProvidersPairing(t *testing.T) {
...
+		for _, provider := range getPairingRes.Providers {
+			if _, ok := providerCount[provider.Address]; !ok {
+				info := providerInfo{count: 0, stakeAmount: provider.Stake.Amount.Int64()}
+				providerCount[provider.Address] = info
+			} else {
+				info := providerCount[provider.Address]
+				info.count++
+				providerCount[provider.Address] = info
+			}
+		}
```
orenl-lava, line 882: how about:
```
    for _, provider := range getPairingRes.Providers {
        info, ok := providerCount[provider.Address]
        if !ok {
            info = providerInfo{count: 0, stakeAmount: provider.Stake.Amount.Int64()}
            providerCount[provider.Address] = info
        }
        info.count++
    }
```

docs/static/openapi.yml
```diff
@@ -30369,16 +30369,6 @@ paths:
           in: query
           required: false
           type: boolean
-        - name: pagination.reverse
```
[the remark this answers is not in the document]
orenl-lava: too late.

x/pairing/keeper/pairing_test.go
orenl-lava: Fixed 0bb8a541

x/pairing/keeper/pairing_test.go
orenl-lava: fixed 0bb8a541

x/pairing/keeper/pairing_test.go
orenl-lava: 0bb8a541fixed

x/pairing/keeper/pairing_test.go
orenl-lava: fixed 0bb8a541

x/pairing/keeper/scores/score.go
orenl-lava: fixed 7944693a

orenl-lava: fixed 7944693a

PR #658, opened 2023-07-30
CNS-552: tech debt identifying missing fields in yaml

oren-lava:
This PR fixes a tech debt in which you couldn't define a field with 0 (it would detect it as a missing field). Now only fields that are absent from the yaml file are regarded as missing fields


orenl-lava:
> You can use mapstructure.Metadata to capture missing (and extraneous) settings in the config vs the struct, either via the Metadata option or using DecodeMetadata() function instead of Decode().

Added a few commits to implement this.

Yaroms (approved):

PR #660, opened 2023-07-30
CNS-496: fixation allow append after delete

orenl-lava:
Changed fixation to allow `AppendEntry()` after `DelEntry()` takes effect, with the following new rules:
* `AppendEntry()` may add retroactive entry only if it does not precede an existing latest entry.
* `AppendEntry()` may not add future entries on or beyond a `DeletedAt` in the future.
* Adding a future entry maturing on a future `DeleteAt` is permitted but useless, because it will be deleted straight away). 
* `DelEntry()` discards all future entries on or beyond the `DeletedAt` (and owing to the previous rule, none will be added until thy delete actually takes place).

Also fix a few glitches in the previous code:
* The check for retroactive invalid Append was off-by-one.
* If the first entry appended was future - the code did not set a time for its maturity block (so it never matures)
* Deleted entries incorrectly retained their `IsLatest` flag
* Entries with `DeleteAt` superseded by maturing future entries incorrectly retained their `DeleteAt`
* Fixation testing advanced multiple block without going calling `AdvanceBlock()` for each block


x/projects/keeper/project_test.go
```diff
@@ -671,6 +671,68 @@ func TestChargeComputeUnits(t *testing.T) {
...
+	// add same key to other project - should fail
+	err = ts.addProjectKeys(projectID2, sub1Addr, types.ProjectDeveloperKey(dev1Addr))
+	require.NotNil(t, err)
+
+	// del key
+	err = ts.delProjectKeys(projectID1, sub1Addr, types.ProjectDeveloperKey(dev1Addr))
+	require.Nil(t, err)
+
+	// del takes effect in next epoch
+	require.True(t, ts.isKeyInProject(projectID1, dev1Addr, types.ProjectKey_DEVELOPER))
```
Yaroms, line 722: adding the key should work even before we advance the epoch and should apply next then, lets test it here

orenl-lava: in the current semantics, adding the same entry again is only possible after the entry has reached its `deleteAt` block. this matches the current semantics of adding projects (or keys) which takes place now (or at the beginning of this epoch) - so cannot complete successfully while a delete is pending.

we can extend this to allow adding the same entry in the future (including at/on the current `deleteAt`) if necessary. but that is more subtle so left for a separate PR.

Yaroms (approved):

PR #667, opened 2023-07-31
CNS-516: fix rand to ensure determinism

orenl-lava:
Using the global rand directly is not safe in terms of determinism, e.g. if
another go routine calls (now or in the future) it to get a random number at
the wrong time. Likewise, rand.Seed() is deprecated in Go (for very similar
reasons).
    
Also, Go does not guarantee that the PRNG driving the default source for rand
will always provide the same output for a given seed - across different Go
versions.
    
Instead, implement our own source (PRNG) that combines rand.New() to provide
a dedicated (private) instance of a PRNG that uses the xoroshiro256** PRNG
from the the xorshift package as a source.
    
(See https://en.wikipedia.org/wiki/Xorshift for more details)

### DIFF
--- a/utils/rand/rand.go
+++ b/utils/rand/rand.go
@@ -1,0 +1,33 @@
+package rand
+
+import (
+	"crypto/sha256"
+	"encoding/binary"
+	"math/rand"
+
+	"github.com/vpxyz/xorshift/xoroshiro128starstar"
+)
+
+func generateSeed(data []byte) (seed int64) {
+	sum256 := sha256.Sum256(data)
+
+	// fold the SHA-256 hash into a 64-bit seed using bitwise XOR
+	for i := 0; i < len(sum256)/8; i++ {
+		seed ^= int64(binary.BigEndian.Uint64(sum256[i*8 : (i+1)*8]))
+	}
+
+	return seed
+}
+
+// returns a new RPNG instance properly seeded
+func New(data []byte) *rand.Rand {
+	seed := generateSeed(data)
+	source := xoroshiro128starstar.NewSource(seed)
+	return rand.New(source)
+}
+
+// re-seeds an existing RPNG instance
+func Seed(rng *rand.Rand, data []byte) {
+	seed := generateSeed(data)
+	rng.Seed(seed)
+}
--- a/utils/rand/rand.go
+++ b/utils/rand/rand.go
@@ -8,25 +8,20 @@ import (
 	"github.com/vpxyz/xorshift/xoroshiro128starstar"
 )
 
-func generateSeed(data []byte) (seed int64) {
+func generateSeed(data []byte) int64 {
 	sum256 := sha256.Sum256(data)
-
-	// fold the SHA-256 hash into a 64-bit seed using bitwise XOR
-	for i := 0; i < len(sum256)/8; i++ {
-		seed ^= int64(binary.BigEndian.Uint64(sum256[i*8 : (i+1)*8]))
-	}
-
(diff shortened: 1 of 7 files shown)


x/pairing/keeper/scores/score.go
```diff
@@ -189,17 +188,10 @@ func PickProviders(ctx sdk.Context, scores []*PairingScore, groupCount int, hash
...
-	for i := 0; i < len(sum256)/8; i++ {
-		seed ^= int64(binary.BigEndian.Uint64(sum256[i*8 : (i+1)*8]))
-	}
-
-	rand.Seed(seed)
+	rng := rand.New(hashData)
 
 	for it := 0; it < groupCount; it++ {
-		randomValue := uint64(rand.Int63n(scoreSum.BigInt().Int64())) + 1
+		randomValue := uint64(rng.Int63n(scoreSum.BigInt().Int64())) + 1
```
Yaroms, line 194: the +1 needs to be inside the random

utils/rand/rand_test.go
```diff
@@ -0,0 +1,70 @@
...
+
+import (
+	"strconv"
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func isUniformlyDistributed(t *testing.T, f func(n int64) int64) {
+	t.Helper()
```
Yaroms, line 11: just so i can learn, what does this do?

x/pairing/keeper/pairing_test.go
```diff
@@ -777,134 +777,94 @@ func TestSelectedProvidersPairing(t *testing.T) {
...
+		mapProviders[provider.Address] = i
+	}
+
+	// calculate the total expected weight
+	var totalWeight int64
+	for _, provider := range allProviders {
+		totalWeight += weight(provider)
+	}
+
+	// calculate calculate the histogram
```
oren-lava, line 800: "calculate calculate"

oren-lava (changes requested):

utils/rand/rand.go
```diff
@@ -0,0 +1,33 @@
...
+	"math/rand"
+
+	"github.com/vpxyz/xorshift/xoroshiro128starstar"
+)
+
+func generateSeed(data []byte) (seed int64) {
+	sum256 := sha256.Sum256(data)
+
+	// fold the SHA-256 hash into a 64-bit seed using bitwise XOR
+	for i := 0; i < len(sum256)/8; i++ {
```
Yaroms, line 15: take only the lsb

x/pairing/keeper/pairing_test.go
orenl-lava: yes, it's a lot of calculus :)

x/pairing/keeper/scores/score.go
orenl-lava: the +1 is at the right place as it is.
we want a random number between 1 and the scoreSum, as values 1-score-of-first-item would mean "pick the first", and the next score-of-second-item values would mean "pick the second" and so on. but `Int63n()` returns a number between 0 and scoreSum-1 - so we need to shift up by one.

utils/rand/rand.go
orenl-lava: fixed 21eca577

x/pairing/keeper/pairing_test.go
orenl-lava: fixed 21eca577

Yaroms (approved):

Yaroms (approved):

PR #678, opened 2023-08-03
CNS fixation genesis

Yaroms:


common/fixation_entry.go
```diff
@@ -996,37 +996,61 @@ func (fs *FixationStore) setVersion(ctx sdk.Context, val uint64) {
...
-	defer iterator.Close()
+func (fs *FixationStore) Export(ctx sdk.Context) types.GenesisState {
+	gs := types.GenesisState{}
 
-	data := []types.RawMessage{}
-	for ; iterator.Valid(); iterator.Next() {
-		data = append(data, types.RawMessage{Key: iterator.Key(), Value: iterator.Value()})
+	gs.Version = fs.getVersion(ctx)
+
+	for _, EntryIndex := range fs.GetAllEntryIndices(ctx) {
```
orenl-lava, line 1004: `GetAllEntryIndices()` skips entries marked deleted (but still in stale period or otherwise held as marker - see details in `deleteStaleEntries()`. You need to use `AllEntryIndicesFilter()` with a nil filter instead.

I suggest you add a test - generate some state with adds, deletes, refcount, future entries, stale, etc. Then do export than init to a second store, then compare the stores using the loop of the previous version of the export.  (That would have caught it).

orenl-lava, line 1004: elsewhere the code usually uses "index" (and anyhow no capital).

common/fixation_entry.go
```diff
@@ -996,37 +996,61 @@ func (fs *FixationStore) setVersion(ctx sdk.Context, val uint64) {
...
+		safeIndex, err := types.SanitizeIndex(entries.Index)
+		if err != nil {
+			utils.LavaFormatPanic("unsafe fixation entry in genesis file", err, utils.Attribute{Key: "Index", Value: entries.Index})
+		}
+
+		fs.setEntryIndex(ctx, safeIndex, entries.IsLive)
 
-	for _, data := range data {
-		store.Set(data.Key, data.Value)
+		for _, entry := range entries.Entries {
```
orenl-lava, line 1048: You validate the EntryIndex above, but you do not validate the actual entries here. Why not validate (or not validate) both?

common/fixation_entry_index.go
```diff
@@ -47,6 +49,14 @@ func (fs FixationStore) removeEntryIndex(ctx sdk.Context, safeIndex types.SafeIn
 	store.Delete(types.KeyPrefix(string(safeIndex)))
 }
 
+// isEntryIndexLive returns isLive from store
+func (fs FixationStore) isEntryIndexLive(ctx sdk.Context, safeIndex types.SafeIndex) bool {
```
orenl-lava, line 53: see `types.IsEntryIndexLive()`

proto/lava/common/fixationEntry.proto
```diff
@@ -14,6 +14,18 @@ message Entry {
     bool is_latest = 7; // is this entry the latest entry now?
 }
 
+message Entries {
```
orenl-lava, line 17: I prefer to make it clear this used purely for genesis by renaming to e.g. "GenesisEntryIndex" - that way we know it's the equivalent of an EntryIndex that is used in Genesis.

orenl-lava (changes requested):

common/fixation_entry.go
Yaroms: done https://github.com/lavanet/lava/pull/678/commits/6ee8bf73cd4a8c56bc341c44c26513105b9f97ec

Yaroms: https://github.com/lavanet/lava/pull/678/commits/6ee8bf73cd4a8c56bc341c44c26513105b9f97ec

common/fixation_entry.go
Yaroms: https://github.com/lavanet/lava/pull/678/commits/6ee8bf73cd4a8c56bc341c44c26513105b9f97ec

common/fixation_entry_index.go
Yaroms: https://github.com/lavanet/lava/pull/678/commits/6ee8bf73cd4a8c56bc341c44c26513105b9f97ec

orenl-lava (approved):

PR #720, opened 2023-08-17
CNS-578: small fixes for init chain scripts

oren-lava:
init_chain.sh - if `jq` is not installed, the script fails. Added jq installation
init_chain_commands.sh - added another `wait_next_block` after proposals to avoid statistical `account sequence mismatch` error

### DIFF
--- a/scripts/init_chain.sh
+++ b/scripts/init_chain.sh
@@ -6,7 +6,7 @@ source $__dir/useful_commands.sh
 
 # Check if jq is not installed
 if ! command_exists jq; then
-    echo "Unable to install jq using apt or brew. Please install jq manually."
+    echo "jq not found. Please install jq using the init_install.sh script or manually."
     exit 1
 fi
 
--- a/scripts/init_chain.sh
+++ b/scripts/init_chain.sh
@@ -1,43 +1,11 @@
 #!/bin/bash
 # make install-all
 killall -9 lavad
-
-# Function to check if a command is available
-command_exists() {
-    command -v "$1" >/dev/null 2>&1
-}
-
-# Flag to track if jq installation is successful
-jq_installed=false
+__dir=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+source $__dir/useful_commands.sh
 
 # Check if jq is not installed
 if ! command_exists jq; then
-    # Try to install jq using apt (for Debian/Ubuntu)
-    if command_exists apt; then
-        echo "Installing jq using apt..."
-        sudo apt update
-        sudo apt install -y jq
-        if command_exists jq; then
-            echo "jq has been successfully installed."
-            jq_installed=true
-        fi
-    fi
-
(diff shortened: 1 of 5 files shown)


scripts/init_chain.sh
```diff
@@ -2,6 +2,46 @@
...
+# Function to check if a command is available
+command_exists() {
+    command -v "$1" >/dev/null 2>&1
+}
+
+# Flag to track if jq installation is successful
+jq_installed=false
+
+# Check if jq is not installed
+if ! command_exists jq; then
```
orenl-lava, line 14: IMO this script should _not_ modify the underlying system. first and foremost, it breaks a user's expectation. second, users usually won't run it as root. and third, it anyhow does not (and will not) cover all possible OS flavors.

if `jq` or any other utility is missing, this script should simply fail with a clear indication of the problem, and how the user could fix it.

instead of auto-install here, add a new script e.g. `init_install.sh` (or whatever) that will check what's missing (for all scripts) and install it, or - if the OS not supported by this script - will indicate what the user needs to do. ideally it will describe what it is about to do and prompt the user for permission prior to doing it. future contributors may enhance it to support more systems.

scripts/useful_commands.sh
```diff
@@ -29,4 +29,20 @@ function wait_next_block {
         break
     fi
   done
+}
+
+function wait_two_blocks {
```
orenl-lava, line 34: would it suffice to simply call `wait_next_block` twice?

actually, rename it to `wait_count_blocks` as follows:
```
function wait_count_blocks {
  for i in $(seq 1 $1); do
    wait_next_block
  done
}
```

orenl-lava (changes requested):

orenl-lava:
while at it, I noticed the OS-specific logic for `sed` (twice) in there. it can be simplified by doing:

```
case "$(uname)" in
  Darwin)
    SED_INLINE="-i ''" ;;
  Linux)
    SED_INLINE="-i" ;;
  *)
    echo "unknown system: $(uname)"
    exit 1 ;;
esac

...
sed $SED_INLINE ...
...
```

with this only one copy of (each relevant) sed command is needed and is useful for both.
(disclaimer: not tested on Darwin, but the idea should be clear).

scripts/init_chain.sh
oren-lava: done. see https://github.com/lavanet/lava/commit/9cd756adfddb049652b4976c240d950a733b1a3c

oren-lava: also this: https://github.com/lavanet/lava/commit/b3bb435b6edeeeb6fa860e9fd70fb3deef53b3f4
sorry for the two commits

scripts/useful_commands.sh
oren-lava: done. see https://github.com/lavanet/lava/commit/461d3dc2bac8029c3f2f3d7bb3664db2158c26e7

Yaroms (approved):

Yaroms (approved):

PR #745, opened 2023-08-29
CNS-570 delegate add new module (part 3)

orenl-lava:


app/app.go
```diff
@@ -204,18 +204,19 @@ var (
...
-		conflictmoduletypes.ModuleName:     {authtypes.Minter, authtypes.Burner, authtypes.Staking},
+		authtypes.FeeCollectorName:               nil,
+		distrtypes.ModuleName:                    nil,
+		minttypes.ModuleName:                     {authtypes.Minter},
+		stakingtypes.BondedPoolName:              {authtypes.Burner, authtypes.Staking},
+		stakingtypes.NotBondedPoolName:           {authtypes.Burner, authtypes.Staking},
+		govtypes.ModuleName:                      {authtypes.Burner},
+		ibctransfertypes.ModuleName:              {authtypes.Minter, authtypes.Burner},
+		epochstoragemoduletypes.ModuleName:       {authtypes.Minter, authtypes.Burner, authtypes.Staking},
+		dualstakingmoduletypes.BondedPoolName:    {authtypes.Burner, authtypes.Staking},
```
Yaroms, line 215: can you explain?
this is stuff i am not familiar with :)

x/dualstaking/keeper/delegate.go
```diff
@@ -0,0 +1,572 @@
+package keeper
+
+// Delegation allows securing funds for a specific provider to effectively increase
+// its stake so it will be paired with consumers more often. The delegators do not
+// transfer the funds to the provider but only bestow the funds with it. In return
+// to locking the funds there, delegators get some of the provider’s profit (after
+// commission deduction).
+//
+// The delegated funds are stored in the module's NotBondedPoolName account. Upon
```
oren-lava, line 9: delegators can delegate or unbond. so shouldn't delegated funds be in the BondedPool? (and not the NotBondedPool)

x/dualstaking/keeper/delegate.go
```diff
@@ -0,0 +1,572 @@
...
+		// new delegation (i.e. not increase of existing one)
+		delegationEntry = types.NewDelegation(delegator, provider, chainID)
+	}
+
+	delegationEntry.AddAmount(amount)
+
+	err := k.delegationFS.AppendEntry(ctx, prefix, nextEpoch, &delegationEntry)
+	if err != nil {
+		// append should never fail here
+		return utils.LavaFormatError("critical: append delegation entry", err,
```
oren-lava, line 75: add the chain ID in the error attributes

x/dualstaking/keeper/delegate.go
```diff
@@ -0,0 +1,572 @@
...
+	return nil
+}
+
+// increaseDelegation increases the delegation of a delegator to a provider for a
+// given chain. It updates the fixation stores for both delegations and delegators,
+// and updates the (epochstorage) stake-entry.
+func (k Keeper) increaseDelegation(ctx sdk.Context, delegator, provider, chainID string, amount sdk.Coin, nextEpoch uint64) error {
+	// get, update and append the delegation entry
+	var delegationEntry types.Delegation
+	prefix := types.DelegationKey(delegator, provider, chainID)
```
Yaroms, line 63: this is an index, not a prefix, can be confusing

x/dualstaking/types/delegate.go
```diff
@@ -0,0 +1,48 @@
...
+import (
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/lavanet/lava/utils/slices"
+)
+
+func NewDelegation(delegator, provider, chainID string) Delegation {
+	return Delegation{
+		Delegator: delegator,
+		Provider:  provider,
+		ChainID:   chainID,
```
Yaroms, line 12: the coin should be init here with the denom of "ulava"

oren-lava (changes requested):

x/dualstaking/keeper/delegate.go
```diff
@@ -0,0 +1,572 @@
...
+		return utils.LavaFormatError("critical: delagate to provider with address mismatch", sdkerrors.ErrInvalidAddress,
+			utils.Attribute{Key: "provider", Value: provider},
+			utils.Attribute{Key: "address", Value: stakeEntry.Address},
+		)
+	}
+
+	stakeEntry.DelegateTotal = stakeEntry.DelegateTotal.Add(amount)
+
+	// fail if amount was negative and the result underflowed the existing total
+	if stakeEntry.DelegateTotal.IsNegative() {
```
Yaroms, line 229: if the amount is not negative than adding it to the total will always be positive

x/dualstaking/keeper/keeper.go
```diff
@@ -22,6 +24,12 @@ type (
 		bankKeeper         types.BankKeeper
 		accountKeeper      types.AccountKeeper
 		epochstorageKeeper types.EpochstorageKeeper
+		specKeeper         types.SpecKeeper
+		stakingKeeper      types.StakingKeeper
+
+		delegationFS common.FixationStore
+		delegatorFS  common.FixationStore
```
Yaroms, line 31: why do we need both of those?
they both show delegator -> providers
we need the opposite,
providers -> delegators like we said in the design

x/dualstaking/keeper/delegate.go
```diff
@@ -0,0 +1,572 @@
...
+	return nil
+}
+
+// decreaseDelegation decreases the delegation of a delegator to a provider for a
+// given chain. It updates the fixation stores for both delegations and delegators,
+// and updates the (epochstorage) stake-entry.
+func (k Keeper) decreaseDelegation(ctx sdk.Context, delegator, provider, chainID string, amount sdk.Coin, nextEpoch uint64) error {
+	// get, update and append the delegation entry
+	var delegationEntry types.Delegation
+	prefix := types.DelegationKey(delegator, provider, chainID)
```
Yaroms, line 112: this is an index, not a prefix

x/dualstaking/keeper/delegate.go
```diff
@@ -0,0 +1,572 @@
...
+}
+
+// setUnbondingTimer creates a timer for an unbonding operation, that will expire
+// in the spec's unbonding-hold-blocks blocks from now. only one unbonding request
+// can be placed per block (for a given delegator/provider/chainID combination).
+func (k Keeper) setUnbondingTimer(ctx sdk.Context, delegator, provider, chainID string, amount sdk.Coin) error {
+	key := encodeForTimer(delegator, provider, chainID)
+
+	period := k.stakingKeeper.UnbondingTime(ctx)
+	timeout := ctx.BlockTime().Add(period).UTC().Unix()
```
Yaroms, line 475: do this for unstakeholdblocks instead of time.
this will be changed once we align the unstake of regulare provider, we need those to be the same

x/dualstaking/keeper/delegate.go
```diff
@@ -0,0 +1,572 @@
...
+
+	// the timer key encodes the unique delegator/provider/chainID combination.
+	// the timer data holds the amount to be released in the future (marshalled).
+
+	if k.unbondingTS.HasTimerByBlockTime(ctx, uint64(timeout), key) {
+		return types.ErrUnbondingInProgress
+	}
+
+	data, _ := yaml.Marshal(amount)
```
Yaroms, line 484: why do we use yaml?
lets align with the rest of the code and marshal with json

x/dualstaking/types/types.go
```diff
@@ -1 +1,13 @@
...
+
+const (
+	DelegateEventName   = "delegate_to_provider"
+	UnbondingEventName  = "unbond_from_provider"
+	RedelegateEventName = "redelegate_between_providers"
+	RefundedEventName   = "refund_to_delegator"
+)
+
+const (
+	NotBondedPoolName = "not_bonded_dualstaking_pool"
```
Yaroms, line 11: the uses for those are switched
bonded are for tokens that are delegated
unbonded are for tokens waiting to be returned to the delegator

proto/lavanet/lava/dualstaking/tx.proto
```diff
@@ -1,7 +1,47 @@
...
-service Msg {}
+service Msg {
+      rpc Delegate(MsgDelegate) returns (MsgDelegateResponse);
+      rpc Redelegate(MsgRedelegate) returns (MsgRedelegateResponse);
+      rpc Unbond(MsgUnbond) returns (MsgUnbondResponse);
+// this line is used by starport scaffolding # proto/tx/rpc
+}
+
+message MsgDelegate {
+  string delegator = 1;
```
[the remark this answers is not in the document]
orenl-lava: fixed:  cd29aacc

x/dualstaking/keeper/delegate.go
orenl-lava: fixed  96a898ea

x/dualstaking/keeper/delegate.go
orenl-lava: fixed  96a898ea

x/dualstaking/keeper/delegate.go
orenl-lava: fixed 96a898ea

x/dualstaking/keeper/delegate.go
orenl-lava: fixed  96a898ea

x/dualstaking/keeper/delegate.go
```diff
@@ -0,0 +1,572 @@
...
+		if amount.IsNegative() {
+			return nil
+		}
+		// otherwise, we return an error because delegation to a provider that is
+		// not stake does not make sense.
+		return types.ErrProviderNotStaked
+	}
+
+	// sanity check
+	if stakeEntry.Address != provider {
```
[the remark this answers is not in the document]
orenl-lava: sanity test modeled after similar test in our `x/pairing/keeper/staking.go`

x/dualstaking/keeper/delegate.go
orenl-lava: this function may be called with positive amount or negative amount (e.g. redelegate,unbond)
in case of the latter, this test checks that the amount was not too big to have causes negative outcome - which would mean there were not enough funds deletegated to the provider to begin with.

x/dualstaking/keeper/delegate.go
orenl-lava: discussed with Omer who recommended to use time based hold-period like cosmos's `x/staking`, and specifically use the same period from that module.

x/dualstaking/keeper/delegate.go
orenl-lava: originally modeled after other code in `x/staking`.
fixed 96a898ea

x/dualstaking/keeper/keeper.go
orenl-lava: `delegationFS` maps providers+chainID to delegations (key is provider_chainID_delegator; entry is delegation data).
`delegatorFS` maps delegators to their providers (key is delegator; entry is list of providers)

it's already documented in the code (in `x/dualstaking/keeper/delegate.go`), but added short comments here too (see 96a898ea)

x/dualstaking/types/delegate.go
orenl-lava: fixed 96a898ea

x/dualstaking/types/types.go
orenl-lava: fixed 96a898ea

x/dualstaking/keeper/delegate.go
orenl-lava: changed in 5c318faa

x/dualstaking/keeper/delegate.go
```diff
@@ -0,0 +1,563 @@
...
+	if _, err := sdk.AccAddressFromBech32(provider); err != nil {
+		utils.LavaFormatError("critical: finalizeBonding invalid provider", err, attrs...)
+		return
+	}
+
+	// sanity: saved amount is valid
+	if err := validateCoins(amount); err != nil {
+		utils.LavaFormatError("critical: finalizeBonding invalid amount", err, attrs...)
+		return
+	} else if amount.IsZero() || amount.IsNegative() {
```
omerlavanet, line 517: validateCoins checks that, redundant

x/dualstaking/keeper/delegate.go
```diff
@@ -0,0 +1,563 @@
...
+
+	return nil
+}
+
+var space = " "
+
+// encodeForTimer generates timer key unique to a specific delegation; thus it
+// must include the delegator, provider, and chainID.
+func encodeForTimer(delegator, provider, chainID string) []byte {
+	dlen, plen, clen := len(delegator), len(provider), len(chainID)
```
omerlavanet, line 436: can we sanitize the separator here? since we decode from it later

x/dualstaking/keeper/delegate.go
```diff
@@ -0,0 +1,563 @@
...
+			utils.Attribute{Key: "chainID", Value: chainID},
+		)
+	}
+
+	// note: if spec was not found, the default choice is Spec_dynamic == 0
+
+	block := uint64(ctx.BlockHeight())
+	if providerType == spectypes.Spec_static {
+		return k.epochstorageKeeper.UnstakeHoldBlocksStatic(ctx, block)
+	} else {
```
omerlavanet, line 558: remove the else, then we don't need unreached

x/epochstorage/keeper/migrations.go
```diff
@@ -33,3 +35,36 @@ func (m Migrator) Migrate2to3(ctx sdk.Context) error {
...
+
+// Migrate3to4 implements store migration from v3 to v4:
+// - initialize DelegateTotal, DelegateLimit, DelegateCommission
+func (m Migrator) Migrate3to4(ctx sdk.Context) error {
+	const ProviderKey = "provider"
+
+	utils.LavaFormatDebug("migrate: epochstorage")
+
+	storage := m.keeper.GetAllStakeStorage(ctx)
+	for _, storage := range storage {
```
omerlavanet, line 47: can we rename one of them

x/pairing/keeper/unstaking.go
```diff
@@ -130,15 +126,24 @@ func (k Keeper) creditUnstakingEntries(ctx sdk.Context, entriesToUnstake []epoch
 	}
 }
 
-func (k Keeper) unstakeHoldBlocks(ctx sdk.Context, chainID string) (uint64, error) {
-	spec, found := k.specKeeper.GetSpec(ctx, chainID)
+// TODO: this is duplicated in x/dualstaking/keeper/delegate.go; merge into one call
```
omerlavanet, line 129: do this TODO :)

omerlavanet (dismissed):
very good job! minor stuff

x/dualstaking/keeper/delegate.go
orenl-lava: unsure what you mean by "sanitize"?
delegator, provider and chainID were all validated by now, and therefore may not include spaces.
see x/dualstaking/types/keys.go line 35 for rationale of choosing "space" as separator.

x/dualstaking/keeper/delegate.go
```diff
@@ -0,0 +1,563 @@
...
+// can be placed per block (for a given delegator/provider/chainID combination).
+func (k Keeper) setUnbondingTimer(ctx sdk.Context, delegator, provider, chainID string, amount sdk.Coin) error {
+	key := encodeForTimer(delegator, provider, chainID)
+
+	unholdBlocks := k.getUnbondHoldBlocks(ctx, chainID)
+	timeout := uint64(ctx.BlockHeight()) + unholdBlocks
+
+	// the timer key encodes the unique delegator/provider/chainID combination.
+	// the timer data holds the amount to be released in the future (marshalled).
+	if k.unbondingTS.HasTimerByBlockHeight(ctx, timeout, key) {
```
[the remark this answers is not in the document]
Yaroms: this is to in case there is unbonding tx in the same block,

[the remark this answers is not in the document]
orenl-lava: currently there isn't an API to get a timer's data or to modify a timer.  (both are needed: a second unbond in the same block would require to raise the unbond amount associated with the timer).  I don't think the scenario of two unbonds (same delegator, same provider) in the same block is worth the extra trouble - so we just disallow; user can retry the next block.

x/dualstaking/keeper/delegate.go
orenl-lava: this is originally _your_ code; I just added the comment to clarify it.

x/dualstaking/keeper/delegate.go
```diff
@@ -0,0 +1,563 @@
...
+
+	if err := k.decreaseStakeEntry(ctx, provider, chainID, amount); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+// increaseStakeEntry increases the (epochstorage) stake-entry of the provider for a chain.
+func (k Keeper) increaseStakeEntry(ctx sdk.Context, provider, chainID string, amount sdk.Coin) error {
```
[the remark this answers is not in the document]
orenl-lava: fixed 94aaec19

x/dualstaking/keeper/delegate.go
orenl-lava: fixed 94aaec19

x/epochstorage/keeper/migrations.go
orenl-lava: fixed 94aaec19

x/pairing/keeper/unstaking.go
orenl-lava: at a second glance this isn't so useful, because the first parts anyhow depends on the calling module; also, will need to add epochstorage as dependency to spec module, which is more trouble than this refactor would save. so instead, I changed to comment to note that any changes should be reflected in both sites. 

fixed 31633a97

Yaroms (approved):

PR #765, opened 2023-09-03
CNS-593: fix "make proto-gen" breakage

orenl-lava:
The build 'make proto-gen' stopped working because the 'proto-build' docker image from cosmos no longer supports bash. Fix by adjusting the script to be compatible with 'sh' instead of 'bash'.

### DIFF
--- a/scripts/protocgen.sh
+++ b/scripts/protocgen.sh
@@ -22,7 +22,7 @@ cd proto
 proto_dirs=$(find . -path -prune -o -name '*.proto' -print0 | xargs -0 -n1 dirname | sort | uniq)
 for dir in $proto_dirs; do
   for file in $(find "${dir}" -maxdepth 1 -name '*.proto'); do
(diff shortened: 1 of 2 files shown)


Yaroms:
using this branch, trying protogenc.sh on my env doesnt work:
./scripts/protocgen.sh
./scripts/protocgen.sh: 3: source: not found
./scripts/protocgen.sh: 5: command_exists: not found

orenl-lava:
> using this branch, trying protogenc.sh on my env doesnt work:

sigh, that was incomplete commit. fixed.

oren-lava:
Not a must, but when I'm running ./scripts/protocgen.sh from this branch, it prints "option" strings like this: `option go_package = "github.com/lavanet/lava/common/types";`
Can you make it not print that (like before)?

orenl-lava:
Done.

Yaroms (approved):

PR #767, opened 2023-09-04
CNS-525: query delegator providers (take three)

oren-lava:
Implemented a query that accepts a delegator and prints all the providers it's delegated to


proto/lavanet/lava/dualstaking/query.proto
```diff
@@ -14,6 +15,11 @@ service Query {
   rpc Params(QueryParamsRequest) returns (QueryParamsResponse) {
     option (google.api.http).get = "/lavanet/lava/dualstaking/params";
   }
+
+  // Queries a list of ListProjects items.
+	rpc DelegatorProviders(QueryDelegatorProvidersRequest) returns (QueryDelegatorProvidersResponse) {
```
orenl-lava, line 20: indentation in proto should be 2 spaces

x/dualstaking/keeper/delegate.go
```diff
@@ -561,3 +561,23 @@ func (k Keeper) getUnbondHoldBlocks(ctx sdk.Context, chainID string) uint64 {
 
 	// NOT REACHED
 }
+
+// GetDelegatorProviders gets all the providers the delegator is delegated to
+func (k Keeper) GetDelegatorProviders(ctx sdk.Context, delegator string) ([]string, error) {
+	var delegatorEntry types.Delegator
+	prefix := types.DelegatorKey(delegator)
+	nextEpoch, err := k.getNextEpoch(ctx)
```
orenl-lava, line 569: what is the reason to return the future value (of next epoch) and not the current value?
and if so, should we not clarify that some delegations will be in effect only in next epoch?
or better, add a flag to select between the state now and by next epoch.

x/dualstaking/keeper/grpc_query_delegator_providers.go
```diff
@@ -0,0 +1,50 @@
...
+	}
+
+	providers, err := k.GetDelegatorProviders(ctx, req.Delegator)
+	if err != nil {
+		return nil, err
+	}
+
+	var delegations []types.Delegation
+	for _, provider := range providers {
+		indices := k.delegationFS.GetAllEntryIndicesWithPrefix(ctx, provider)
```
orenl-lava, line 32: this will give all delegations of the provider, not only those for the specific delegator we're interested in. so if the provider has e.g. 2 delegations, we will return both even though one does is not relevant to the query.

you need to decode the index into <provider, chain, delegator>, and pick only those with the matching delegator.

note that this will be become simpler when delegation is per provider and not per <provider,chainID> - once the provider is known we could dig up <provider,delegator> directly with FindEntry() without needing to loop over the indices.

x/dualstaking/keeper/grpc_query_delegator_providers.go
```diff
@@ -0,0 +1,50 @@
...
+		for _, ind := range indices {
+			var delegation types.Delegation
+			found := k.delegationFS.FindEntry(ctx, ind, nextEpoch, &delegation)
+			if !found {
+				delegator, provider, chainID := types.DelegationKeyDecode(ind)
+				utils.LavaFormatError("provider found in delegatorFS but not in delegationFS", fmt.Errorf("provider delegation not found"),
```
orenl-lava, line 38: this should be critical

x/dualstaking/keeper/grpc_query_delegator_providers_test.go
```diff
@@ -0,0 +1,59 @@
...
+	_, delegator := ts.GetAccount(common.CONSUMER, 0)
+	_, provider := ts.GetAccount(common.PROVIDER, 0)
+
+	spec := ts.Spec("mock")
+
+	amountUint64 := uint64(100)
+	amount := sdk.NewCoin(epochstoragetypes.TokenDenom, sdk.NewIntFromUint64(amountUint64))
+
+	// delegate and query
+	_, err := ts.TxDualstakingDelegate(delegator, provider, spec.Index, amount)
```
orenl-lava, line 26: please also test:
- one provider with multiple delegator, with multiple chains
- one delegator with more than one provider
- one delegator to unstaking, and to unstaked, provider.

orenl-lava (changes requested):

proto/lavanet/lava/dualstaking/query.proto
oren-lava: done. see https://github.com/lavanet/lava/commit/f64a33890597cbbf9002d2b66c3daf435abfe569

x/dualstaking/keeper/delegate.go
oren-lava: added `--show-pending` flag that lets you see pending delegators (applied in the next epoch). see https://github.com/lavanet/lava/commit/aaefdd332ea2388ff5aa28e3e320f79875219754

x/dualstaking/keeper/grpc_query_delegator_providers.go
oren-lava: done. see https://github.com/lavanet/lava/commit/2f9f3f255ed7898d378b890621825f64a86bcb3b

x/dualstaking/keeper/grpc_query_delegator_providers_test.go
```diff
@@ -0,0 +1,120 @@
...
+	require.Nil(t, err)
+	ts.AdvanceEpoch()
```
orenl-lava, line 29: with this, the test does not cover the case of query with "pending=false" before the change takes effect.
I think it's better to do both pending cases (false, true) here, like:

    delegate
    query without pending (should not show)
    query with pending (should show)
    advance epoch
    query without pending (should show)
    query with pending (should show)

and repeat that after each change to delegation. then you can drop the new test function.
(maybe refactor a helper for queries that will do both).

x/dualstaking/types/expected_keepers.go
```diff
@@ -32,6 +32,7 @@ type EpochstorageKeeper interface {
 	ModifyStakeEntryCurrent(ctx sdk.Context, chainID string, stakeEntry epochstoragetypes.StakeEntry, removeIndex uint64)
 	UnstakeHoldBlocks(ctx sdk.Context, block uint64) (res uint64)
 	UnstakeHoldBlocksStatic(ctx sdk.Context, block uint64) (res uint64)
+	GetEpochStartForBlock(ctx sdk.Context, block uint64) (epochStart, blockInEpoch uint64, err error)
```
orenl-lava, line 35: (not needed: see commend where it is used)

x/dualstaking/keeper/grpc_query_delegator_providers.go
```diff
@@ -0,0 +1,59 @@
...
+	providers, err := k.GetDelegatorProviders(ctx, req.Delegator, epoch)
+	if err != nil {
+		return nil, err
+	}
+
+	var delegations []types.Delegation
+	for _, provider := range providers {
+		indices := k.delegationFS.GetAllEntryIndicesWithPrefix(ctx, provider)
+		for _, ind := range indices {
+			delegator, providerDecode, chainID := types.DelegationKeyDecode(ind)
```
orenl-lava, line 40: `providerDecode` is not needed (must be same as `provider`), so replace with _

x/dualstaking/keeper/grpc_query_delegator_providers.go
```diff
@@ -0,0 +1,59 @@
...
+			if delegator != req.Delegator {
+				continue
+			}
+			var delegation types.Delegation
+			found := k.delegationFS.FindEntry(ctx, ind, epoch, &delegation)
+			if !found {
+				utils.LavaFormatError("critical: provider found in delegatorFS but not in delegationFS", fmt.Errorf("provider delegation not found"),
+					utils.Attribute{Key: "delegator", Value: delegator},
+					utils.Attribute{Key: "provider", Value: providerDecode},
```
orenl-lava, line 49: can stick to `provider` (they must be the same)

x/dualstaking/keeper/delegate.go
```diff
@@ -561,3 +561,19 @@ func (k Keeper) getUnbondHoldBlocks(ctx sdk.Context, chainID string) uint64 {
...
 }
+
+// GetDelegatorProviders gets all the providers the delegator is delegated to
+func (k Keeper) GetDelegatorProviders(ctx sdk.Context, delegator string, epoch uint64) (providers []string, err error) {
+	var delegatorEntry types.Delegator
+	prefix := types.DelegatorKey(delegator)
+
+	found := k.delegatorFS.FindEntry(ctx, prefix, epoch, &delegatorEntry)
+	if !found {
+		return nil, utils.LavaFormatWarning("could not get delegator providers", fmt.Errorf("delegator not found"),
```
orenl-lava, line 572: why is this an error?
I'd think that querying for delegations of a (potential) delegator would return empty list if that (potential) delegator doesn't have any delegations, rather than an error.

orenl-lava: also, in comparison querying for provider's delegation indeed return empty list if there are none (and not an error)

orenl-lava: also, the code should validate the input delegator

proto/lavanet/lava/dualstaking/query.proto
```diff
@@ -24,3 +30,12 @@ message QueryParamsResponse {
   // params holds all the parameters of this module.
   Params params = 1 [(gogoproto.nullable) = false];
 }
+
+message QueryDelegatorProvidersRequest {
+  string delegator = 1;
+  bool show_pending_delegators = 2;
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/4b8f71ec561ce22c1d4d4e982a0c9dee619c01ec

x/dualstaking/keeper/grpc_query_delegator_providers_test.go
oren-lava: fixed unit tests. see https://github.com/lavanet/lava/commit/8a4c37623f94c6d470444b0e3edde0db7168d081 for final version

x/dualstaking/keeper/grpc_query_delegator_providers.go
oren-lava: done. see https://github.com/lavanet/lava/commit/d1e2d3c1afe278bf342ca8f975d03ca5326020c6

x/dualstaking/keeper/grpc_query_delegator_providers.go
oren-lava: done. see https://github.com/lavanet/lava/commit/d1e2d3c1afe278bf342ca8f975d03ca5326020c6

x/dualstaking/types/expected_keepers.go
oren-lava: done. see https://github.com/lavanet/lava/commit/d1e2d3c1afe278bf342ca8f975d03ca5326020c6

x/dualstaking/keeper/delegate.go
oren-lava: done. see https://github.com/lavanet/lava/commit/d044fe62749f4d88f2bde153789d8316749618bd

x/dualstaking/keeper/grpc_query_delegator_providers_test.go
oren-lava: I updated the WithPending test. see https://github.com/lavanet/lava/commit/4564b1721989d3de340f742ac8e86c0e7a93e348

x/dualstaking/keeper/grpc_query_delegator_providers.go
```diff
@@ -0,0 +1,57 @@
...
+	}
+
+	providers, err := k.GetDelegatorProviders(ctx, req.Delegator, epoch)
+	if err != nil {
+		return nil, err
+	}
+
+	var delegations []types.Delegation
+	for _, provider := range providers {
+		indices := k.delegationFS.GetAllEntryIndicesWithPrefix(ctx, provider)
```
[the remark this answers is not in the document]
orenl-lava: won't work, because the key is currently <provider, chain-id, delegator>. for this to work, the key needs to be <provider, delegator, chain-id>.

we can change the key order now; but keep in mind that feature will change to be per provider (instead of per provider per chain-id). when that happens, it would work (as I mentioned in the other comment).

orenl-lava (approved):

x/dualstaking/keeper/grpc_query_delegator_providers.go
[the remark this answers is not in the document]
Yaroms: do we lose anything from changing it now?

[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/9ac02e1108b5717f64fbd634f72bb539a88c4bdf

x/dualstaking/keeper/grpc_query_delegator_providers.go
```diff
@@ -33,19 +33,15 @@ func (k Keeper) DelegatorProviders(goCtx context.Context, req *types.QueryDelega
 
 	var delegations []types.Delegation
 	for _, provider := range providers {
-		indices := k.delegationFS.GetAllEntryIndicesWithPrefix(ctx, provider)
+		indices := k.delegationFS.GetAllEntryIndicesWithPrefix(ctx, provider+" "+req.Delegator)
```
Yaroms, line 36: use the function to create the key

oren-lava: done. see https://github.com/lavanet/lava/commit/54bfb017febb0c4111849702c621d895731b9718

Yaroms (approved):

PR #777, opened 2023-09-11
PRT-847: Providerside Error Handling

candostyavuz:
**Changelog**

- Added a check for response status codes 504 and 429 in REST and Tendermint URI calls. (Note: For JSON-RPC and Tendermint-RPC, this is already handled by the doRequest function in http.go).
- Implemented a JSON format validation for all REST API responses.
- Introduced an ID verification step for both JSON-RPC and Tendermint-RPC.

**Manual Testing**

- Conducted tests by sending requests to a server configured to always return status codes 504 and 429. Verified that the new status code checks are operational.
- Reviewed provider logs and confirmed that for successful relay transactions, no errors were generated due to JSON formatting issues or ID discrepancies.

### DIFF
--- a/protocol/chainlib/chainproxy/rpcclient/client.go
+++ b/protocol/chainlib/chainproxy/rpcclient/client.go
@@ -389,7 +389,7 @@ func (c *Client) Notify(ctx context.Context, method string, args ...interface{})
 	msg.ID = nil
 
 	if c.isHTTP {
-		return c.sendHTTP(ctx, op, msg, false)
+		return c.sendHTTP(ctx, op, msg, true)
 	}
 	return c.send(ctx, op, msg)
 }
--- a/protocol/chainlib/chainproxy/rpcclient/client.go
+++ b/protocol/chainlib/chainproxy/rpcclient/client.go
@@ -272,7 +272,7 @@ func (c *Client) SetHeader(key, value string) {
 //
 // The result must be a pointer so that package json can unmarshal into it. You
 // can also pass nil, in which case the result is ignored.
-func (c *Client) CallContext(ctx context.Context, id json.RawMessage, method string, params interface{}) (*JsonrpcMessage, error) {
+func (c *Client) CallContext(ctx context.Context, id json.RawMessage, method string, params interface{}, isJsonRPC bool) (*JsonrpcMessage, error) {
 	var msg *JsonrpcMessage
 	var err error
 	switch p := params.(type) {
@@ -293,7 +293,7 @@ func (c *Client) CallContext(ctx context.Context, id json.RawMessage, method str
 	op := &requestOp{ids: []json.RawMessage{msg.ID}, resp: make(chan *JsonrpcMessage, 1)}
 
 	if c.isHTTP {
-		err = c.sendHTTP(ctx, op, msg)
+		err = c.sendHTTP(ctx, op, msg, isJsonRPC)
 	} else {
 		err = c.send(ctx, op, msg)
 	}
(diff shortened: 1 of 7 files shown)


protocol/chainlib/tendermintRPC.go
```diff
@@ -528,6 +528,11 @@ func (cp *tendermintRpcChainProxy) SendURI(ctx context.Context, nodeMessage *rpc
 		defer res.Body.Close()
 	}
 
+	if res.StatusCode == 504 || res.StatusCode == 429 {
+		utils.LavaFormatError("Received invalid status code", nil, utils.Attribute{Key: "Status Code", Value: res.StatusCode})
+		return nil, "", nil, err
+	}
+
```
ranlavanet, line 535: in URI we need to check json format as well.

protocol/chainlib/rest.go
```diff
@@ -440,6 +441,11 @@ func (rcp *RestChainProxy) SendNodeMsg(ctx context.Context, ch chan interface{},
 		defer res.Body.Close()
 	}
 
+	if res.StatusCode == 504 || res.StatusCode == 429 {
```
ranlavanet, line 444: can you export this to the errorHandler struct? so we can change it in the future in one place for all occurrences?

ranlavanet (changes requested):

protocol/chainlib/tendermintRPC.go
candostyavuz: done : 5fb6ebf93a2aa1aebc13b40e0cfd51fd8d779160

protocol/chainlib/chainproxy/rpcclient/client.go
```diff
@@ -389,7 +389,7 @@ func (c *Client) Notify(ctx context.Context, method string, args ...interface{})
 	msg.ID = nil
 
 	if c.isHTTP {
-		return c.sendHTTP(ctx, op, msg)
+		return c.sendHTTP(ctx, op, msg, false)
```
ranlavanet, line 392: here we are true I dont think its being used but just in case set it to true.

protocol/chainlib/chainproxy/rpcclient/http.go
```diff
@@ -163,7 +163,7 @@ func (c *Client) sendBatchHTTP(ctx context.Context, op *requestOp, msgs []*Jsonr
 	if !ok {
 		return fmt.Errorf("sendBatchHTTP - c.writeConn.(*httpConn) - type assertion failed, type:" + fmt.Sprintf("%s", c.writeConn))
 	}
-	respBody, err := hc.doRequest(ctx, msgs)
+	respBody, err := hc.doRequest(ctx, msgs, false)
```
ranlavanet, line 166: here true as well as this is a default jsonrpc client not tendermint.

protocol/chainlib/chainproxy/rpcclient/http.go
```diff
@@ -200,7 +200,7 @@ func (hc *httpConn) doRequest(ctx context.Context, msg interface{}) (io.ReadClos
 	if err != nil {
 		return nil, err
 	}
-	if resp.StatusCode < 200 || resp.StatusCode >= 300 {
```
ranlavanet, line 203: I wanted to validate the HandleStatusError here for tendermint rpc + jsonrpc. we get it here for free. add a method to the client expose it to the error handler and reuse it there

ranlavanet: because now we dont check 429 and 504 for tendermint and jsonrpc RPC clients.

ranlavanet (changes requested):

protocol/chainlib/chainproxy/rpcclient/http.go
candostyavuz: done: 716bc2e03e4756f6168f912bd89afd04b1a615e7

ranlavanet (approved):

PR #795, opened 2023-09-14
PRT-899 cherry pick optmizing probing to providers

ranlavanet:


ecosystem/lava-sdk/src/lavasession/consumerSessionManager.ts
```diff
@@ -786,61 +769,104 @@ export class ConsumerSessionManager {
...
-        //Logger.error(err);
-      }
+            const lavaEpoch = probeReply.getLavaEpoch();
+            if (epoch == 0 && this.currentEpoch == 0) {
+              this.currentEpoch = lavaEpoch; // setting the epoch for initialization.
+            }
+            Logger.debug(
+              `Lava Epoch for provider ${consumerSessionWithProvider.publicLavaAddress}: ${lavaEpoch}`
+            );
+            this.epochTracker.setLatestBlockFor(
```
Aleksao998, line 818: Here we are missing:
```
consumerSessionWithProvider.setPairingEpoch(lavaEpoch);
```
If we do not setup pairingEpoch it will always be 0, so we need to to update it

Aleksao998: Which makes sense, when we probe the provider we have a lava epoch block, which should be used as the pairing epoch for this consumerSessionWithProvider. And not the one we set on init

omerlavanet (dismissed):

ranlavanet:
Merge only tomorrow

omerlavanet (dismissed):

protocol/rpcprovider/rpcprovider_server.go
```diff
@@ -382,7 +382,8 @@ func (rpcps *RPCProviderServer) verifyRelaySession(ctx context.Context, request
 		} else if request.RelaySession.Epoch == 0 {
 			errorMessage = "user reported lava block 0, either it's test rpcprovider or a consumer that has no node access"
 		}
-		return nil, nil, utils.LavaFormatWarning(errorMessage, lavasession.EpochMismatchError,
+		return nil, nil, utils.LavaFormatInfo(errorMessage,
```
omerlavanet, line 385: this can;t be changed you have to return lavasession.EpochMismatchError here not a generic error, also your print of a specific constant does not provide information

omerlavanet (changes requested):

omerlavanet (approved):

PR #802, opened 2023-09-18
CNS v0.23.2

Yaroms:


ranlavanet (changes requested):
upgrade protocol version as well to 0.23.2? if its already .2 change all to .3

ranlavanet (approved):

PR #805, opened 2023-09-19
PRT - fix_jsonrpc_ids_parsing_on_error_handling

candostyavuz:


protocol/chainlib/jsonRPC.go
```diff
@@ -484,6 +484,18 @@ func (cp *JrpcChainProxy) SendNodeMsg(ctx context.Context, ch chan interface{},
 			return nil, "", nil, utils.LavaFormatError("jsonRPC error", err, utils.Attribute{Key: "GUID", Value: ctx})
 		}
 		replyMsg = *replyMessage
+
+		reqId, idErr := rpcInterfaceMessages.IdFromRawMessage(nodeMessage.ID)
```
ranlavanet, line 488: let's create a common method that we use and add it to the error handler class. for both tendermint and jsonrpc

ranlavanet (changes requested):

protocol/chainlib/jsonRPC.go
candostyavuz: fixed in: 92fb9142c18e1e95113212df7301f824f46e1e14

protocol/chainlib/jsonRPC.go
```diff
@@ -484,6 +484,11 @@ func (cp *JrpcChainProxy) SendNodeMsg(ctx context.Context, ch chan interface{},
 			return nil, "", nil, utils.LavaFormatError("jsonRPC error", err, utils.Attribute{Key: "GUID", Value: ctx})
 		}
 		replyMsg = *replyMessage
+
+		err := cp.HandleIDMismatchError(nodeMessage.ID, replyMessage.ID)
```
ranlavanet, line 488: lets change the name of this to ValidateRequestAndResponseIds 
so its clear its a validation.

ranlavanet (changes requested):

protocol/chainlib/jsonRPC.go
candostyavuz: fixed: 3e029e715b1a279a51f3bf50120ead79aaa4cffe

ranlavanet (approved):

PR #807, opened 2023-09-19
json array response parse error for rest (Aptos /transactions API fix)

candostyavuz:


ranlavanet:
for some reason this doesnt point to branch

PR #809, opened 2023-09-19
PRT-891 Reward Persistency - ft-agilefreaks | ft-distributed-lab

shleikes:
This PR adds the Badger DB as a snapshot database for the rewards server, to save the rewards that hadn't been confirmed yet, so in the rare case that the provider or the rewards server fail, or restarted, the rewards will be restored from the filesystem and the providers wont lose their rewards.


protocol/rpcprovider/rewardserver/reward_server.go
```diff
@@ -272,23 +354,136 @@ func (rws *RewardServer) PaymentHandler(payment *PaymentRequest) {
...
+			rws.resetSnapshotTimerAndSaveRewardsSnapshotToDBAndResetTimer()
+		case <-rws.rewardsSnapshotThresholdCh:
+			rws.resetSnapshotTimerAndSaveRewardsSnapshotToDBAndResetTimer()
+		}
+	}
+}
+
+func (rws *RewardServer) resetSnapshotTimerAndSaveRewardsSnapshotToDBAndResetTimer() {
+	// We lock without defer because the DB is already locking itself
+	rws.lock.RLock()
```
ranlavanet, line 402: rws.rewardsSnapshotTimer.Reset needs write privileges so not rlock -> lock.

protocol/rpcprovider/rewardserver/reward_server.go
```diff
@@ -272,23 +354,136 @@ func (rws *RewardServer) PaymentHandler(payment *PaymentRequest) {
...
+		return
+	}
+
+	utils.LavaFormatDebug("saving rewards snapshot to the DB", utils.Attribute{Key: "proofs", Value: len(rewardEntities)})
+
+	var err error
+	for i := 0; i < MaxDBSaveRetries; i++ {
+		err = rws.rewardDB.BatchSave(rewardEntities)
+		if err == nil {
+			return
```
ranlavanet, line 431: let's add a success print INFO level

protocol/rpcprovider/rewardserver/reward_server.go
```diff
@@ -272,23 +354,136 @@ func (rws *RewardServer) PaymentHandler(payment *PaymentRequest) {
...
+		}
+
+		utils.LavaFormatWarning("failed saving proofs snapshot to rewardDB. Retrying...", err,
```
ranlavanet, line 434: lets change it to INFO level instead of warning.

ranlavanet (changes requested):

protocol/rpcprovider/rewardserver/reward_server.go
shleikes: What about the error?

protocol/rpcprovider/rewardserver/badger_db.go
```diff
@@ -16,7 +16,7 @@ type BadgerDB struct {
 	shardString  string
 	rewards      map[string]*entryWithTtl
 	db           *badger.DB
-	lock         sync.RWMutex
+	lock         sync.Mutex
```
ranlavanet, line 19: why change to Mutex? we can still Rlock for read and lock for write, it will be faster

ranlavanet (changes requested):

protocol/rpcprovider/rewardserver/badger_db.go
shleikes: After a second look, you're right, I'll change that backl

shleikes: Fixed

ranlavanet (approved):

PR #824, opened 2023-09-25
Helm improvements

easeev:
1. When using the `latest` tag, `pullPolicy` should be set to `Always` to avoid image caching. Overall using `latest` is not recommended and it's better to use a particular pinned version
2. Added checksum for the configmap so that the pod restarts when the configmap is changed
3. Fixed syntax for probe cmd which was not working correctly


praveenperera:
Thanks @easeev these are great improvements!

PR #827, opened 2023-09-26
[PRT-899] Lava sdk provider optimizer

bunea:


ecosystem/lava-sdk/src/providerOptimizer/providerOptimizer.ts
```diff
@@ -0,0 +1,656 @@
...
+  }
+
+  public chooseProvider(
+    allAddresses: string[],
+    ignoredProviders: string[],
+    cu: number,
+    requestedBlock: number,
+    perturbationPercentage: number
+  ): string[] {
+    const returnedProviders: string[] = [];
```
omerlavanet, line 213: in the go implementation first provider is initialized to "" and here it's not, this causes a bug

omerlavanet (changes requested):

ecosystem/lava-sdk/src/rpcconsumer/rpcconsumer_server.ts
```diff
@@ -181,63 +191,103 @@ export class RPCConsumerServer {
...
+          return new Error("unsupported error " + err);
+        });
+
+      promises.push(promise);
+    }
+
+    const timeout = new Promise<Error | undefined>((resolve) => {
+      setTimeout(
+        () => resolve(SDKErrors.noResponseTimeout),
+        relayTimeout + secondsToMillis(2)
```
omerlavanet, line 276: there is no need to create another timeout here, relayInner has it's own timeout

ecosystem/lava-sdk/src/rpcconsumer/rpcconsumer_server.ts
```diff
@@ -140,27 +140,37 @@ export class RPCConsumerServer {
...
+
+    if (consumerSessionsMap.size == 0) {
+      return new Error("returned empty consumerSessionsMap");
+    }
+
+    let finalRelayResult: RelayResult | RelayError[] | Error | undefined;
+    let responsesReceived = 0;
+
+    const trySetFinalRelayResult = (
+      res: RelayResult | RelayError[] | Error
```
omerlavanet, line 162: what happens here is that we take only the last reply, this is not the golang functionality in which we take the first reply that is not an error

ecosystem/lava-sdk/src/rpcconsumer/rpcconsumer_server.ts
```diff
@@ -181,63 +191,103 @@ export class RPCConsumerServer {
...
       );
-      return relayResult;
-    } catch (err) {
-      if (err instanceof Error) {
-        return err;
-      }
-      return new Error("unsupported error " + err);
+    });
+
+    const response = await Promise.race([Promise.all(promises), timeout]);
```
omerlavanet, line 280: we do not want to Promise.All all providers, this means we wait for all of them, so even if one of them is down we will reach the timeout, while we already got a valid response long ago
wanted behaviour like the golang: we return the repsponse once one of the promises settled correctly

omerlavanet (changes requested):

omerlavanet:
merged in another PR

PR #854, opened 2023-10-03
PRT-944: Chaging lavavisor block parsing to just query the params and validate the version that way we wont miss blocks

candostyavuz:
**This PR changes Lavavisor upgrade detection mechanism**

- Old Mechanism: monitoring the upgrade events to detect a need for an upgrade. (If we miss the event, then gg)
- New Mechanism: Queries the protocol version params from blockchain in Update() and compares it with the lastKnownVersion. When the mismatch occurs, lastKnown version parameter is updated inside Update() function so that Lavavisor will detect the mismatch inside its own `ValidateProtocolVersion` implementation.

**This PR also adds block height info to lavavisor info logs as follows:**

Screenshot 2023-10-03 at 14 29 41


ranlavanet:
merged in #876

ranlavanet:
merged in https://github.com/lavanet/lava/pull/876

PR #886, opened 2023-10-12
CNS-separate-cli-test

Yaroms:


.github/workflows/consensus_tests.yml
```diff
@@ -64,10 +64,4 @@ jobs:
...
 
     - name: lava subscription unit Tests
-      run: go test ./x/subscription/... 
-    
-    - name: lavad cli test
-      run: |
-        chmod +x ./scripts/cli_test.sh    
-        ./scripts/cli_test.sh     
-        killall lavad    
+      run: go test ./x/subscription/...
```
oren-lava, line 67: why don't we also run tests from the modules: downtime, dualstaking and fixationstore?

oren-lava (changes requested):

.github/workflows/consensus_tests.yml
Yaroms: we should, ill add it now

oren-lava (approved):

PR #895, opened 2023-10-16
Add emergency mode to rpcprovider/rpcconsumer

slandymani:


protocol/statetracker/consumer_state_tracker.go
```diff
@@ -48,6 +48,13 @@ func (cst *ConsumerStateTracker) RegisterConsumerSessionManagerForPairingUpdates
 	if !ok {
 		utils.LavaFormatFatal("invalid updater type returned from RegisterForUpdates", nil, utils.Attribute{Key: "updater", Value: pairingUpdaterRaw})
 	}
+
+	// register for updates in case of emergency mode is enabled
+	pairingUpdaterWithEmergencyRaw := cst.StateTracker.RegisterForEmergencyModeUpdates(ctx, pairingUpdater)
```
omerlavanet, line 53: don't do this code between the casting to pairing updater and pairingUpdater.RegisterPairing, making it complex and weird, do both registers without affecting one another

protocol/statetracker/downtime_params_updater.go
```diff
@@ -0,0 +1,78 @@
...
+
+	(*downtimeParamsUpdatable).SetDowntimeParams(*params)
+	dpu.downtimeParamsUpdatables = append(dpu.downtimeParamsUpdatables, downtimeParamsUpdatable)
+	return nil
+}
+
+func (dpu *DowntimeParamsUpdater) Update(latestBlock int64) {
+	dpu.lock.Lock()
+	defer dpu.lock.Unlock()
+	paramsUpdated := dpu.eventTracker.getLatestDowntimeParamsUpdateEvents()
```
omerlavanet, line 65: missing a check event tracker has the block you need

protocol/statetracker/downtime_params_updater.go
```diff
@@ -0,0 +1,78 @@
...
+	if paramsUpdated {
+		// fetch updated downtime params from consensus
+		params, err := dpu.downtimeParamsStateQuery.GetDowntimeParams(context.Background())
```
omerlavanet, line 68: do a retry if we fail we lose the change and can't recover it until the next change

protocol/statetracker/epoch_updater.go
```diff
@@ -100,3 +104,24 @@ func (eu *EpochUpdater) Update(latestBlock int64) {
...
+// update virtual epoch for registered updatables during emergency mode
+func (eu *EpochUpdater) EmergencyModeUpdate(virtualEpoch uint64) {
+	eu.lock.RLock()
+	defer eu.lock.RUnlock()
+
+	if virtualEpoch <= eu.currentVirtualEpoch {
+		return
+	}
+
+	utils.LavaFormatDebug("Emergency mode is turn on", utils.Attribute{Key: "virtual_epoch", Value: virtualEpoch})
```
omerlavanet, line 117: fix grammar

protocol/statetracker/epoch_updater_test.go
```diff
@@ -32,6 +32,8 @@ func (esqt *epochUpdatable) UpdateEpoch(epoch uint64) {
 	esqt.updatedBlock = epoch
 }
 
+func (esqt *epochUpdatable) UpdateVirtualEpoch(epoch uint64, virtualEpoch uint64) {}
```
omerlavanet, line 35: where are the tests?

protocol/statetracker/provider_state_tracker.go
```diff
@@ -104,7 +124,20 @@ func (pst *ProviderStateTracker) LatestBlock() int64 {
 }
 
 func (pst *ProviderStateTracker) GetMaxCuForUser(ctx context.Context, consumerAddress, chainID string, epoch uint64) (maxCu uint64, err error) {
-	return pst.stateQuery.GetMaxCuForUser(ctx, consumerAddress, chainID, epoch)
+	latestBlockTime := pst.EventTracker.getLatestBlockTime()
```
omerlavanet, line 127: state tracker shouldnt modify the responses, it's an interface to fetch the data

omerlavanet: also event tracker is not the place to save latest block time, there is chain tracker for lava for it

protocol/statetracker/event_tracker.go
```diff
@@ -74,6 +84,12 @@ func (et *EventTracker) updateBlockResults(latestBlock int64) (err error) {
 	return nil
 }
 
+func (et *EventTracker) getLatestBlockTime() time.Time {
```
omerlavanet, line 87: event tracker tracks events not blocks and their time, remove it form here you can use chaintracker time

protocol/statetracker/event_tracker.go
```diff
@@ -66,6 +69,13 @@ func (et *EventTracker) updateBlockResults(latestBlock int64) (err error) {
 	et.lock.Lock()
 	defer et.lock.Unlock()
 	if latestBlock > et.latestUpdatedBlock {
+		// if latestBlock from args != 0, we don't fetch time from chain,
+		// so we believe that lastBlockTime is equal to time now
+		if latestBlockTime == nil {
```
omerlavanet, line 74: remove

protocol/statetracker/event_tracker.go
```diff
@@ -25,11 +25,13 @@ type EventTracker struct {
 	clientCtx          client.Context
 	blockResults       *ctypes.ResultBlockResults
 	latestUpdatedBlock int64
+	latestBlockTime    time.Time
```
omerlavanet, line 28: does not belong here, you have chaintracker to save the block time

protocol/statetracker/pairing_updater.go
```diff
@@ -154,3 +171,11 @@ func (pu *PairingUpdater) filterPairingListByEndpoint(ctx context.Context, pairi
 	// replace previous pairing with new providers
 	return pairing, nil
 }
+
+func (pu *PairingUpdater) EmergencyModeUpdate(virtualEpoch uint64) {
+	pu.lock.Lock()
+	defer pu.lock.Unlock()
+	pu.updateConsumerSessionManagerCULimits(virtualEpoch)
+
+	pu.currentVirtualEpoch = virtualEpoch
```
omerlavanet, line 180: we are not using this member, you can remove it

protocol/statetracker/pairing_updater.go
```diff
@@ -17,6 +17,7 @@ type PairingUpdater struct {
 	lock                       sync.RWMutex
 	consumerSessionManagersMap map[string][]*lavasession.ConsumerSessionManager // key is chainID so we don;t run getPairing more than once per chain
 	nextBlockForUpdate         uint64
+	currentVirtualEpoch        uint64
```
omerlavanet, line 20: this member is not used, remove

protocol/statetracker/pairing_updater.go
```diff
@@ -51,9 +52,13 @@ func (pu *PairingUpdater) UpdaterKey() string {
 }
 
 func (pu *PairingUpdater) Update(latestBlock int64) {
-	pu.lock.RLock()
-	defer pu.lock.RUnlock()
+	pu.lock.Lock()
```
omerlavanet, line 55: no need for lock, removing the unused member makes this all read

protocol/statetracker/pairing_updater.go
```diff
@@ -51,9 +52,13 @@ func (pu *PairingUpdater) UpdaterKey() string {
...
+	defer pu.lock.Unlock()
 	ctx := context.Background()
+
+	// reset virtual epoch after the end of emergency mode
+	pu.currentVirtualEpoch = 0
```
omerlavanet, line 60: why?

protocol/statetracker/pairing_updater.go
```diff
@@ -85,6 +90,18 @@ func (pu *PairingUpdater) Update(latestBlock int64) {
 	pu.nextBlockForUpdate = nextBlockForUpdateMin
 }
```
omerlavanet, line 92: can we add all new functions next to each other and maintain hierarchy? for internal functions caller above calle

protocol/statetracker/pairing_updater.go
```diff
@@ -85,6 +90,18 @@ func (pu *PairingUpdater) Update(latestBlock int64) {
...
 
+// updateConsumerSessionManagerCULimits for new virtual epoch when emergency mode is enabled
+func (pu *PairingUpdater) updateConsumerSessionManagerCULimits(virtualEpoch uint64) {
+	for _, consumerSessionManagers := range pu.consumerSessionManagersMap {
+		for _, consumerSessionManager := range consumerSessionManagers {
+			if consumerSessionManager == nil {
+				continue
+			}
+			consumerSessionManager.UpdateMaxCULimit(virtualEpoch, pu.currentVirtualEpoch)
```
omerlavanet, line 100: should implement update virtual epoch on consumer sessions manager instead of directly calling the max CU function, for modularity, in case we need to add more stuff to virtual epoch changes, also lower coupling with pairing updater and consumer sessions manager

protocol/statetracker/provider_state_tracker.go
```diff
@@ -104,7 +124,20 @@ func (pst *ProviderStateTracker) LatestBlock() int64 {
...
+	downtimeParams := pst.chainTracker.GetDowntimeParams()
```
omerlavanet, line 128: get downtime params from downtime params updater not from chaintracker

protocol/statetracker/provider_state_tracker.go
```diff
@@ -104,7 +124,20 @@ func (pst *ProviderStateTracker) LatestBlock() int64 {
...
+
+	delay := time.Now().UTC().Sub(latestBlockTime)
```
omerlavanet, line 130: time.Since

protocol/statetracker/provider_state_tracker.go
```diff
@@ -104,7 +124,20 @@ func (pst *ProviderStateTracker) LatestBlock() int64 {
...
+
+	epochDuration := downtimeParams.EpochDuration.Milliseconds()
+	virtualEpoch := uint64(0)
+	// check if emergency mode is enabled
+	if delay > downtimeParams.DowntimeDuration {
+		// division delay by epoch duration rounded up, subtract one to skip regular epoch
+		virtualEpoch = uint64((delay.Milliseconds()+epochDuration-1)/epochDuration - 1)
+	}
+
+	return pst.stateQuery.GetMaxCuForUser(ctx, consumerAddress, chainID, epoch, virtualEpoch)
```
omerlavanet, line 140: with your current designs we have two mechanisms, one for changing existing cus and one for new cus, it's a more complex design than just saving the virtual epoch and when opening a session and checking cu left just regard it

protocol/statetracker/state_query.go
```diff
@@ -205,7 +217,8 @@ func (psq *ProviderStateQuery) GetMaxCuForUser(ctx context.Context, consumerAddr
 		psq.ResponsesCache.SetWithTTL(MaxCuResponseKey+key, userEntryRes, 1, DefaultTimeToLiveExpiration)
 	}
 
-	return userEntryRes.GetMaxCU(), err
+	// increase CU during emergency mode for passed virtual epochs
+	return userEntryRes.GetMaxCU() * (virtualEpoch + 1), err
```
omerlavanet, line 221: this multiplication shouldn't be here but rather when using the session

protocol/statetracker/state_tracker.go
```diff
@@ -21,17 +21,23 @@ const (
...
 // it allows also to query specific data form the blockchain and acts as a single place to send transactions
 type StateTracker struct {
-	chainTracker         *chaintracker.ChainTracker
-	registrationLock     sync.RWMutex
-	newLavaBlockUpdaters map[string]Updater
-	EventTracker         *EventTracker
+	chainTracker          *chaintracker.ChainTracker
+	registrationLock      sync.RWMutex
+	newLavaBlockUpdaters  map[string]Updater
+	emergencyModeUpdaters map[string]EmergencyModeUpdater
```
omerlavanet, line 27: all updaters are in the same map, and need to implement Updater interface, your usage is wrong, convert EmergencyModeUpdater to implement Updater interface and use the regular function

protocol/statetracker/state_tracker.go
```diff
@@ -21,17 +21,23 @@ const (
...
+	EventTracker          *EventTracker
 }
 
 type Updater interface {
 	Update(int64)
 	UpdaterKey() string
 }
 
+type EmergencyModeUpdater interface {
```
omerlavanet, line 36: remove

protocol/statetracker/state_tracker.go
```diff
@@ -63,9 +69,15 @@ func NewStateTracker(ctx context.Context, txFactory tx.Factory, clientCtx client
 		})
 	}
 
-	cst := &StateTracker{newLavaBlockUpdaters: map[string]Updater{}, EventTracker: eventTracker}
+	cst := &StateTracker{
+		newLavaBlockUpdaters:  map[string]Updater{},
+		emergencyModeUpdaters: map[string]EmergencyModeUpdater{},
```
omerlavanet, line 74: you are cluttering state tracker for no reason

protocol/statetracker/state_tracker.go
```diff
@@ -89,6 +101,31 @@ func (st *StateTracker) newLavaBlock(latestBlock int64, hash string) {
...
 
+// block handler for emergency mode checking
+func (st *StateTracker) oldLavaBlock(latestBlock int64) {
+	st.registrationLock.RLock()
+	defer st.registrationLock.RUnlock()
+
+	latestBlockTime := st.EventTracker.getLatestBlockTime()
+	downtimeParams := st.chainTracker.GetDowntimeParams()
+
+	delay := time.Now().UTC().Sub(latestBlockTime)
```
omerlavanet, line 112: i see the same code over and over for calculation virtual epochs, use a common function

protocol/statetracker/state_tracker.go
```diff
@@ -89,6 +101,31 @@ func (st *StateTracker) newLavaBlock(latestBlock int64, hash string) {
...
+		return
+	}
+
+	epochDuration := downtimeParams.EpochDuration.Milliseconds()
+
+	// division delay by epoch duration rounded up, subtract one to skip regular epoch
+	virtualEpoch := (delay.Milliseconds()+epochDuration-1)/epochDuration - 1
+
+	for _, updater := range st.emergencyModeUpdaters {
+		updater.EmergencyModeUpdate(uint64(virtualEpoch))
```
omerlavanet, line 125: state tracker shouldnt be dealing with emergency updates, you need to create a new EmergencyTracker class to do it have it register for downtime params, have it be the source of truth for vitrtual epochs, have other classes embed an interface to it so you don't have to register for emergency updates

omerlavanet (commented):
the current design is flawed
1. once a block arrives, the previous data for emergency mode is wrong, if on epoch 60 there was emergency mode with 3 virtual epochs, if 67 block arrives suddenly it becomes 0 and cus are incorrectly calculated
2. a lot of complexity tracking what is oging on where you need to both update every entry and modify new entries and modify returned responses form state tracker, and it's different between consumer and provider

design changes necessary:
1. emergencyTracker class to register for downtime changes, and virtual epochs
2. all epoch updaters can just embed an interface from this class exporting the virtual epoch for each epoch
3. consumer and provider code to use the exact same functions up to the point of calculating the max CU where the virtual epoch is to be used
4. virtual epoch function needs to get an epoch as argument, not only track the latest

omerlavanet (changes requested):
1. design change by my comments
2. there are no tests at all!
3. needs to pass the existing tests as a PR

protocol/statetracker/emergency_tracker.go
```diff
@@ -0,0 +1,84 @@
...
+	cs.downtimeParams = params
+}
+
+func (cs *EmergencyTracker) GetDowntimeParams() downtimev1.Params {
+	cs.lock.RLock()
+	defer cs.lock.RUnlock()
+	return cs.downtimeParams
+}
+
+func (cs *EmergencyTracker) UpdateEpoch(epoch uint64) {
```
omerlavanet, line 37: if this is called by the consumer, this will result in a problem, i assume we create two classes inheriting form one another, one for the consumer tracking only the latest the other for the provider, this way we can't confuse it

protocol/statetracker/emergency_tracker.go
```diff
@@ -0,0 +1,84 @@
...
+
+import (
+	"sync"
+	"time"
+
+	"github.com/lavanet/lava/utils"
+	downtimev1 "github.com/lavanet/lava/x/downtime/v1"
+)
+
+type EmergencyTracker struct {
```
omerlavanet, line 11: let's separate last epoch tracking from full epochs tracking the provider needs, having them together separated only by wether we call updateEpoch is confusing

protocol/statetracker/provider_state_tracker.go
```diff
@@ -130,3 +149,7 @@ func (pst *ProviderStateTracker) GetEpochSizeMultipliedByRecommendedEpochNumToCo
 func (pst *ProviderStateTracker) GetProtocolVersion(ctx context.Context) (*ProtocolVersionResponse, error) {
 	return pst.stateQuery.GetProtocolVersion(ctx)
 }
+
+func (pst *ProviderStateTracker) GetEmergencyTracker() *EmergencyTracker {
```
omerlavanet, line 153: remove, after moving the usage to the constructor

omerlavanet (changes requested):
need to fix:
1. review comments
2. missing tests:
    * unitests for identifying emergency and getting out of it
    * unitests for a full session usage between consumer and provider, then activating emergency on the consumer only (allow usage until the maximum CU)
    * unitests for a full session usage between consumer and provider, then activating emergency on the provider only (allow usage until the maximum CU)
    * unitests for a full session usage between consumer and provider, get to the limit, trigger emergency on one, try to use and fail, trigger on the second, allow usage, get to the limit, increase virtual epoch on one, try to use more and fail, increase virtual epoch on the second too and succeed
    * an e2e flow for emergency: starting up the node, with all the chains, start usage, kill the node, continue usage bypassing the limit, resume the node
3. metrics: current virtual epoch, how many times it was triggered in prometheus both on consumer and provider
4. a recorded demo of it working in a local flow

protocol/statetracker/consumer_state_tracker.go
```diff
@@ -39,12 +40,15 @@ func NewConsumerStateTracker(ctx context.Context, txFactory tx.Factory, clientCt
...
-		ConsumerTxSenderInf: txSender,
-		EmergencyTracker:    emergencyTracker,
+		StateTracker:                stateTrackerBase,
+		stateQuery:                  NewConsumerStateQuery(ctx, clientCtx),
+		ConsumerTxSenderInf:         txSender,
+		ConsumerEmergencyTrackerInf: emergencyTracker,
 	}
-	return cst, nil
+
+	cst.RegisterForPairingUpdates(ctx, emergencyTracker)
```
omerlavanet, line 49: love it good job

protocol/statetracker/pairing_updater.go
```diff
@@ -77,6 +95,18 @@ func (pu *PairingUpdater) Update(latestBlock int64) {
 			}
 		}
 	}
+
+	// get latest epoch from cache
```
omerlavanet, line 99: very nice!

omerlavanet (approved):

PR #901, opened 2023-10-17
Cns 600 add verifications to evmos spec

Yaroms:


omerlavanet (changes requested):
pruning and net version checks don't make a lot of sense to me

omerlavanet (approved):
must notify community before these go live

omerlavanet (approved):

PR #914, opened 2023-10-24
CNS-648: delegators aggregate rewards after a month

oren-lava:
In this PR I've implemented a feature that lets delegators aggregate rewards only after a month from their original delegation. Also, when the delegator redelegates, it's treated as a new delegation (meaning he'll have to wait a month again).

Note that the delegation starts effecting the pairing process (increasing the provider's effective stake) immediately, and not after a month.


x/dualstaking/types/delegate.go
```diff
@@ -6,12 +6,15 @@ import (
 	epochstoragetypes "github.com/lavanet/lava/x/epochstorage/types"
 )
 
-func NewDelegation(delegator, provider, chainID string) Delegation {
+const MONTH_DURATION = 2678395 // number of seconds in a month (assuming 30 days in a month)
```
Yaroms, line 9: use the nextmonth function in the subscription

x/dualstaking/types/delegate.go
```diff
@@ -37,6 +40,10 @@ func (delegation *Delegation) Equal(other *Delegation) bool {
 	return true
 }
 
+func (delegation *Delegation) IsFirstMonthPassed(currentTimestamp int64) bool {
+	return delegation.Timestamp+MONTH_DURATION <= currentTimestamp
```
Yaroms, line 44: save the timestamp as time+month so you wont have to do this calculation every time

oren-lava:
Closing this to avoid difficult merge with main. See updated version (same functionality) here: https://github.com/lavanet/lava/pull/972

PR #917, opened 2023-10-25
CNS-661-add-verification-to-starknet

Yaroms:


omerlavanet (approved):

omerlavanet:
please make sure to give notice right now about this change, not only when this lands

PR #938, opened 2023-11-01
Fix typo in startup message

sigv:
The message "all endpoints up an running" is missing a 'd' for "and".


Yaroms:
Thanks for the fix @sigv

Yaroms (approved):

PR #945, opened 2023-11-07
CNS-668: let users see pending policy changes

oren-lava:
policy changes apply only on the next epoch. When users query for their project, they will now see if there's a pending change in the future


x/projects/keeper/grpc_query_developer.go
```diff
@@ -21,5 +21,15 @@ func (k Keeper) Developer(goCtx context.Context, req *types.QueryDeveloperReques
...
 	}
 
-	return &types.QueryDeveloperResponse{Project: &project}, nil
+	nextEpoch, err := k.epochstorageKeeper.GetNextEpoch(ctx, uint64(ctx.BlockHeight()))
+	if err != nil {
+		return nil, err
+	}
+
+	pendingProject, err := k.GetProjectForDeveloper(ctx, req.Developer, nextEpoch)
+	if project.Equal(pendingProject) || err != nil {
```
Yaroms, line 30: error should be checked first in case of project == nil

x/projects/keeper/grpc_query_info.go
```diff
@@ -21,5 +21,15 @@ func (k Keeper) Info(goCtx context.Context, req *types.QueryInfoRequest) (*types
...
 	}
 
-	return &types.QueryInfoResponse{Project: &project}, nil
+	nextEpoch, err := k.epochstorageKeeper.GetNextEpoch(ctx, uint64(ctx.BlockHeight()))
+	if err != nil {
+		return nil, err
+	}
+
+	pendingProject, err := k.GetProjectForBlock(ctx, req.Project, nextEpoch)
+	if project.Equal(pendingProject) || err != nil {
```
Yaroms, line 30: same here

x/projects/keeper/grpc_query_developer.go
oren-lava: done. see https://github.com/lavanet/lava/commit/dec5f61dbc57b5f56416130660a3dcf465410efa

x/projects/keeper/grpc_query_info.go
oren-lava: done. see https://github.com/lavanet/lava/commit/dec5f61dbc57b5f56416130660a3dcf465410efa

Yaroms (approved):

PR #960, opened 2023-11-09
CNS-667: Subscription auto-renewal

oren-lava:
Added an auto-renewal feature for subscriptions.

To enable auto-renewal you can either:
1. Buy a new subscription with the flag `--enable-auto-renewal`
2. Use the `auto-renewal` TX (can be used to disable auto-renewal too)

When a subscription with enabled auto-renewal expires, it's extended by an additional month automatically. It can fail if:
1. there are not enough funds to pay for the extension
2. the subscription's plan's price increased by more than 5% (policy changes do not influence renewal)


x/subscription/keeper/msg_server_auto_renewal.go
```diff
@@ -0,0 +1,33 @@
...
+
+	sub, found := k.GetSubscription(ctx, msg.Creator)
+	if !found {
+		return nil, utils.LavaFormatWarning("could not change auto-renewal of subscription", fmt.Errorf("subscription not found"),
+			utils.Attribute{Key: "sub_consumer", Value: msg.Creator},
+		)
+	}
+
+	sub.AutoRenewal = msg.Enable
+	err := k.subsFS.AppendEntry(ctx, msg.Creator, uint64(ctx.BlockHeight()), &sub)
```
Yaroms, line 23: this should be at the same block as the original subscription, we dont want another instance in the fixation

oren-lava: done. see https://github.com/lavanet/lava/commit/b9acd69c93fc42fce0e164674af4de9c33525130

Yaroms (approved):

PR #962, opened 2023-11-12
CNS-683: make timer store genesis explicit

oren-lava:
I've made the timer store data that's saved in the genesis readable.
I've also added unit test of import/export of genesis for the fixation and timer stores


x/fixationstore/helpers_test.go
```diff
@@ -0,0 +1,31 @@
+package fixationstore_test
+
+import (
```
Yaroms, line 3: delete me

oren-lava: done. see https://github.com/lavanet/lava/commit/d667226d20162b69428d236c77b32d8df82f533e

Yaroms (approved):

PR #967, opened 2023-11-14
CNS-692: make fixation use BlocksToSave instead of a constant

oren-lava:


Yaroms:
the e2e tests fails only here, check why?

Yaroms (dismissed):

Yaroms (approved):

PR #983, opened 2023-11-19
CNS-655: Hook validator slashing

oren-lava:
In this PR, I've implemented the `BeforeValidatorSlashed` that happens right before a validator is slashed. The hook makes sure that the providers-validators delegations balance is kept intact.


oren-lava:
Closing this. See CNS-655 updated

PR #984, opened 2023-11-19
CNS-655 updated: hook validator slashing

oren-lava:
** UPDATED VERSION **

In this PR, I've implemented the BeforeValidatorSlashed that happens right before a validator is slashed. The hook makes sure that the providers-validators delegations balance is kept intact.


x/dualstaking/keeper/hooks_test.go
```diff
@@ -157,3 +159,180 @@ func TestReDelegateToProvider(t *testing.T) {
...
+			dualstakingkeeper.EMPTY_PROVIDER,
+			provider,
+			dualstakingkeeper.EMPTY_PROVIDER_CHAINID,
+			ts.spec.Index,
+			sdk.NewCoin(epochstoragetypes.TokenDenom, redelegateAmts[i]))
+		require.Nil(t, err)
+	}
+
+	// sanity check: redelegate from val2 to val1 and check delegations balance
+	_, err = ts.TxReDelegateValidator(delegatorAcc, validators[1], validators[0], sdk.NewInt(10))
```
Yaroms, line 299: what i meant for the testing is to test redelegate from one provider to another, not from validator
so use txdualstakingredelegate instead

x/dualstaking/keeper/hooks_test.go
```diff
@@ -157,3 +159,180 @@ func TestReDelegateToProvider(t *testing.T) {
...
+	require.Nil(t, err)
+	diff, err := ts.Keepers.Dualstaking.VerifyDelegatorBalance(ts.Ctx, delegatorAcc.Addr)
+	require.Nil(t, err)
+	require.True(t, diff.IsZero())
+
+	// sanity check: unbond some of val1's funds and check delegations balance
+	_, err = ts.TxUnbondValidator(delegatorAcc, validators[0], sdk.NewInt(10))
```
Yaroms, line 306: same here

x/dualstaking/keeper/hooks_test.go
oren-lava: done. see https://github.com/lavanet/lava/commit/117a1c20140e60fe7004324b1e4e2dfe7d720351

x/dualstaking/keeper/hooks_test.go
oren-lava: done. see https://github.com/lavanet/lava/commit/117a1c20140e60fe7004324b1e4e2dfe7d720351

PR #992, opened 2023-11-22
CNS-686 Add CLI to query from timer store

shleikes:


common/types/time.go
```diff
@@ -0,0 +1,11 @@
+package types
+
+import "time"
+
+const (
+	TimeFormat = "2006-January-02 15:04:05"
+)
+
+func ConvertUnixTimestampToString(val uint64) string {
+	return time.Unix(int64(val), 0).Format(TimeFormat)
```
Yaroms, line 10: what hapens if you just time.Unix(int64(val), 0).string() ?

proto/lavanet/lava/timerstore/query.proto
```diff
@@ -0,0 +1,60 @@
...
+  }
+}
+
+message QueryAllTimersRequest {
+  string store_key = 1;
+  string prefix = 2;
+}
+
+message QueryAllTimersResponse {
+  string blockTimeTimers = 1;
```
Yaroms, line 34: this should be repeated string?

x/timerstore/keeper/grpc_query_all_timers.go
```diff
@@ -0,0 +1,39 @@
...
+				BlockHeightTimers: store.DumpAllTimers(ctx, types.BlockHeight),
+				Tick:              BeginBlock,
+			}, nil
+		}
+	}
+
+	for _, store := range k.timerStoresEnd {
+		if store.GetStoreKey().Name() == req.StoreKey && store.GetStorePrefix() == req.Prefix {
+			return &types.QueryAllTimersResponse{
+				BlockTimeTimers:   store.DumpAllTimers(ctx, types.BlockTime),
```
Yaroms, line 31: make dumpalltimers to return list of strings

x/timerstore/keeper/grpc_query_next.go
```diff
@@ -0,0 +1,59 @@
...
+	"math"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	commontypes "github.com/lavanet/lava/common/types"
+	"github.com/lavanet/lava/x/timerstore/types"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+)
+
+func createQueryNextResponse(nextBlockHeight, nextBlockTime uint64, tick string) *types.QueryNextResponse {
```
Yaroms, line 14: tick is never used, needs to put in the tick

x/timerstore/keeper/keeper.go
```diff
@@ -1,9 +1,15 @@
...
+package keeper
 
 import (
 	"github.com/cosmos/cosmos-sdk/codec"
 	storetypes "github.com/cosmos/cosmos-sdk/store/types"
 	sdk "github.com/cosmos/cosmos-sdk/types"
+	timerstoretypes "github.com/lavanet/lava/x/timerstore/types"
+)
+
+const (
```
Yaroms, line 10: move these consts to the query files

common/types/time.go
shleikes: It will use the default format. I chose not to use the default format since in Hebrew format, the month come first, and in other languages, the month comes second after the day. I wanted it to be clear what is the month and what is the day, and that is why I've created the TimeFormat for.

proto/lavanet/lava/timerstore/query.proto
shleikes: This is how I use it in the code:
```
// ...
BlockTimeTimers:   store.DumpAllTimers(ctx, types.BlockTime),
// ...
```

The function `store.DumpAllTimers` returns a single string

x/timerstore/keeper/grpc_query_all_timers.go
shleikes: Even better: an object

proto/lavanet/lava/timerstore/query.proto
shleikes: Repeated object!

x/timerstore/keeper/grpc_query_next.go
shleikes: Oops

x/timerstore/keeper/keeper.go
shleikes: 👍

Yaroms (approved):

PR #1000, opened 2023-11-26
CNS-Fix-Migrator

Yaroms:


x/dualstaking/keeper/hooks.go
```diff
@@ -133,7 +133,7 @@ func (h Hooks) BeforeDelegationRemoved(ctx sdk.Context, delAddr sdk.AccAddress,
 	if err != nil {
 		return nil
 	}
-	amount := validator.TokensFromShares(delegation.Shares).TruncateInt()
+	amount := validator.TokensFromSharesRoundUp(delegation.Shares).TruncateInt()
 	err = h.k.UnbondUniformProviders(ctx, delAddr.String(), sdk.NewCoin(epochstoragetypes.TokenDenom, amount))
```
omerlavanet, line 137: if unbond uniform fails, do we really want to fail the hook? didn't we say we will allow some cases? maybe let's allow failing unbond uniform in the special case it's the last delegation, i think this should cover our edge case

x/dualstaking/keeper/migrations.go
```diff
@@ -63,6 +62,12 @@ func (m Migrator) ConvertProviderStakeToSelfDelegation(ctx sdk.Context) error {
 		}
 	}
 
+	moduleBalance := m.keeper.bankKeeper.GetBalance(ctx, m.keeper.accountKeeper.GetModuleAddress(types.ModuleName), epochstoragetypes.TokenDenom)
+	err := m.keeper.bankKeeper.BurnCoins(ctx, types.ModuleName, sdk.NewCoins(moduleBalance))
```
omerlavanet, line 66: we are only burning dual staking, we will need to add pairing module burn in a migration (doesn't have to be in this one)

x/dualstaking/keeper/migrations.go
```diff
@@ -74,14 +79,17 @@ func (m Migrator) HandleProviderDelegators(ctx sdk.Context) error {
...
-			return err
-		}
+	// burn all coins from the pools
+	moduleBalance := m.keeper.bankKeeper.GetBalance(ctx, m.keeper.accountKeeper.GetModuleAddress(dualstakingtypes.NotBondedPoolName), epochstoragetypes.TokenDenom)
+	err := m.keeper.bankKeeper.BurnCoins(ctx, dualstakingtypes.NotBondedPoolName, sdk.NewCoins(moduleBalance))
+	if err != nil {
+		return err
+	}
+
+	moduleBalance = m.keeper.bankKeeper.GetBalance(ctx, m.keeper.accountKeeper.GetModuleAddress(dualstakingtypes.NotBondedPoolName), epochstoragetypes.TokenDenom)
```
omerlavanet, line 89: why are we burning twice the same NotBondedPoolName looks like a mistake

omerlavanet (changes requested):

x/dualstaking/keeper/hooks.go
Yaroms: unbonduniform will fail only if something is wrong with the delegations,
so if we want to unbond more than we have it wont sent an error

x/dualstaking/keeper/migrations.go
Yaroms: why?
mpdule pairing after this migration shouldnt hold any coins

x/dualstaking/keeper/migrations.go
Yaroms: done

omerlavanet (approved):

omerlavanet (approved):

PR #1011, opened 2023-11-29
CNS-721: disabled minting module

oren-lava:


Yaroms (changes requested):
this needs to come with an upgrade handler that removes a module.
better to do it here, we will forget it later

Yaroms (approved):

PR #1023, opened 2023-12-05
PRT-fix linter issues

Yaroms:


ecosystem/lava-sdk/src/util/apiCollection.ts
```diff
@@ -1,5 +1,4 @@
 import {
-  Api,
```
shleikes, line 2: You have another lint error here

shleikes (changes requested):

shleikes (approved):

PR #1053, opened 2023-12-18
CNS - 766: Implement subscription auto renewel

shleikes:


oren-lava (dismissed):

x/subscription/keeper/subscription.go
```diff
@@ -316,10 +324,10 @@ func (k Keeper) advanceMonth(ctx sdk.Context, subkey []byte) {
 			sub.MonthCuTotal = plan.PlanPolicy.TotalCuLimit
 
 			k.resetSubscriptionDetailsAndAppendEntry(ctx, &sub, block, false)
-		} else if sub.AutoRenewal {
+		} else if sub.AutoRenewalNextPlan != types.AUTO_RENEWAL_PLAN_NONE {
```
Yaroms, line 327: can you move this to the subscription object so we can call:
sub.AutoRenewal()?

x/subscription/keeper/subscription.go
```diff
@@ -252,7 +228,6 @@ func (k Keeper) upgradeSubscriptionPlan(ctx sdk.Context, duration uint64, sub *t
 	// The "old" subscription's duration is now expired
 	// If called from CreateSubscription, the duration will reset to the duration bought
 	sub.DurationLeft = 0
-	sub.Cluster = types.GetClusterKey(*sub)
```
Yaroms, line 255: why?

x/subscription/client/cli/tx_auto_renewal.go
```diff
@@ -12,34 +13,76 @@ import (
 
 var _ = strconv.Itoa(0)
 
+const (
+	EnableAutoRenewalFlag  = "enable"
```
Yaroms, line 17: why dont we create a flag that can be either true or false?

x/subscription/keeper/msg_server_auto_renewal.go
```diff
@@ -12,22 +13,73 @@ import (
...
+	if !msg.Enable && sub.AutoRenewalNextPlan == types.AUTO_RENEWAL_PLAN_NONE {
+		return nil, utils.LavaFormatWarning("could not change auto-renewal of subscription", fmt.Errorf("auto-renewal is already disabled"),
+			utils.Attribute{Key: "creator", Value: msg.Creator},
+			utils.Attribute{Key: "consumer", Value: msg.Consumer},
+		)
+	}
+
+	// If msg.Enable == true, verify plan index
+	if msg.Enable {
+		if strings.TrimSpace(msg.Index) == "" {
```
Yaroms, line 42: why do we need to trim spaces?

x/subscription/keeper/subscription.go
```diff
@@ -234,9 +193,26 @@ func (k Keeper) createNewSubscription(ctx sdk.Context, plan *planstypes.Plan, cr
 	return sub, nil
 }
 
-func (k Keeper) upgradeSubscriptionPlan(ctx sdk.Context, duration uint64, sub *types.Subscription, toPlan *planstypes.Plan) error {
+func (k Keeper) upgradeSubscriptionPlan(ctx sdk.Context, consumer string, duration uint64, sub *types.Subscription, newPlan *planstypes.Plan) error {
```
Yaroms, line 196: why do you need to pass the consumer if you have it in sub?

x/subscription/keeper/subscription.go
```diff
@@ -234,9 +193,26 @@ func (k Keeper) createNewSubscription(ctx sdk.Context, plan *planstypes.Plan, cr
...
 	block := uint64(ctx.BlockHeight())
 
+	currentPlan, err := k.GetPlanFromSubscription(ctx, consumer, block)
```
Yaroms, line 199: you already have the sub from the input

x/subscription/keeper/subscription.go
```diff
@@ -263,11 +238,44 @@ func (k Keeper) upgradeSubscriptionPlan(ctx sdk.Context, duration uint64, sub *t
 	// Remove one refcount for previous plan
 	k.plansKeeper.PutPlan(ctx, sub.PlanIndex, sub.PlanBlock)
 
-	sub.PlanIndex = toPlan.Index
-	sub.PlanBlock = toPlan.Block
-	sub.MonthCuTotal = toPlan.PlanPolicy.TotalCuLimit
+	sub.PlanIndex = newPlan.Index
+	sub.PlanBlock = newPlan.Block
+	sub.MonthCuTotal = newPlan.PlanPolicy.TotalCuLimit
```
Yaroms, line 243: update cluster

Yaroms: done in the next function :)

x/subscription/keeper/subscription.go
```diff
@@ -263,11 +238,44 @@ func (k Keeper) upgradeSubscriptionPlan(ctx sdk.Context, duration uint64, sub *t
...
+
+	k.resetSubscriptionDetailsAndAppendEntry(ctx, sub, nextEpoch, true)
```
Yaroms, line 245: inside the function we delete the timer if we have error, instead, create the timer only after the append is done

x/subscription/keeper/subscription.go
```diff
@@ -76,63 +76,16 @@ func (k Keeper) CreateSubscription(
...
-					utils.Attribute{Key: "original_plan_block", Value: strconv.FormatUint(subPlan.Block, 10)},
-					utils.Attribute{Key: "original_plan_price", Value: subPlan.Price.String()},
-					utils.Attribute{Key: "current_plan_block", Value: strconv.FormatUint(plan.Block, 10)},
-					utils.Attribute{Key: "current_plan_block", Value: plan.Price.String()},
-				)
-			}
-
-			// update sub plan
-			sub.PlanBlock = plan.Block
+			return k.CreateFutureSubscription(ctx, creator, consumer, planIndex, duration)
```
Yaroms, line 88: we have this flow already in the advance buy, if the user want to use advance buy let him use the flag.
instead return error here

x/subscription/keeper/subscription.go
Yaroms: you also dont use duration here

x/subscription/keeper/subscription.go
```diff
@@ -154,7 +107,7 @@ func (k Keeper) CreateSubscription(
 		return utils.LavaFormatWarning("create subscription failed", err)
 	}
 
-	if !found || sub.AutoRenewal {
+	if !found || sub.AutoRenewalNextPlan != types.AUTO_RENEWAL_PLAN_NONE {
```
Yaroms, line 110: sub.AutoRenewalNextPlan != types.AUTO_RENEWAL_PLAN_NONE i think is not needed anymore since we dont get here from the begin block

Yaroms (commented):
nice :)

x/subscription/keeper/subscription.go
shleikes: Because we already do it later in the function `resetSubscriptionDetailsAndAppendEntry`

x/subscription/client/cli/tx_auto_renewal.go
shleikes: This is just the name of the flag that presented to the user.

I'm not sure I know what you mean...

x/subscription/keeper/msg_server_auto_renewal.go
shleikes: Just in case? I wanted to see if the given arg is an empty string

shleikes:
> x/subscription/keeper/subscription.go

Thank you very much sir 😁

x/subscription/keeper/subscription.go
shleikes: True. Fixed in 1269175585ae3fbc37add86b0d59b43294e9abd4

x/subscription/client/cli/tx_auto_renewal.go
Yaroms: i mean, you are setting 2 flags for 1 boolean.
why not just use one flag?

x/subscription/keeper/subscription.go
Yaroms: this flow lets the user use advance buy without using the flag  --advance-purchase
if the user wants to but something in advance let him do it with the flag so it will be explicit

x/subscription/client/cli/tx_auto_renewal.go
shleikes: Fixed in b715be8644c4eddaf5744f238f7c690dd1120e3f

x/subscription/keeper/subscription.go
shleikes: Fixed in 7fd7523aedabccabf7f8c7e5c34d25a7426e1e58

Yaroms (approved):

PR #1071, opened 2023-12-26
Grow 1144 Adding the --relay-server-address to lava

oneeuroquestion:
For the new usage server under portalapp


protocol/rpcconsumer/rpcconsumer.go
```diff
@@ -97,12 +97,14 @@ type RPCConsumer struct {
 }
 
 // spawns a new RPCConsumer server with all it's processes and internals ready for communications
-func (rpcc *RPCConsumer) Start(ctx context.Context, txFactory tx.Factory, clientCtx client.Context, rpcEndpoints []*lavasession.RPCEndpoint, requiredResponses int, cache *performance.Cache, strategy provideroptimizer.Strategy, metricsListenAddress string, maxConcurrentProviders uint) (err error) {
+func (rpcc *RPCConsumer) Start(ctx context.Context, txFactory tx.Factory, clientCtx client.Context, rpcEndpoints []*lavasession.RPCEndpoint, requiredResponses int, cache *performance.Cache, strategy provideroptimizer.Strategy, metricsListenAddress string, maxConcurrentProviders uint, usageServerAddr string) (err error) {
```
ranlavanet, line 100: i dont like the fact that we just propagate these type of info. next time we need another address we will also propagate it? seems like its time to move to a struct with all the fields so we propagate this struct instead both merics and relay analytics (by the way lets agree on one name and stick with it ? |

"usageServerAddr", in most places you called it relay server here its usage server.

I think a proper name would be "analytics server" / "usage analytic server"

protocol/metrics/rpcconsumerlogs.go
```diff
@@ -28,26 +28,27 @@ const (
...
-	excludeMetricsReferrers string
-	excludedUserAgent       []string
-	consumerMetricsManager  *ConsumerMetricsManager
+	newRelicApplication       *newrelic.Application
+	MetricService             *MetricService
+	StoreMetricData           bool
+	excludeMetricsReferrers   string
+	excludedUserAgent         []string
+	consumerMetricsManager    *ConsumerMetricsManager
+	ConsumerRelayserverClient *ConsumerRelayserverClient
```
ranlavanet, line 37:
```suggestion
	ConsumerRelayServerClient *ConsumerRelayServerClient
```

ranlavanet: please do this fix to all names :) naming conventions are important

protocol/metrics/metrics_provider_manager.go
```diff
@@ -13,6 +13,7 @@ import (
 
 const (
 	MetricsListenFlagName = "metrics-listen-address"
+	RelayServerFlagName   = "relay-server-address"
```
ranlavanet, line 16:
```suggestion
	RelayServerFlagName   = "usage-analytic-server-address "
```

protocol/metrics/consumer_relayserver_client.go
```diff
@@ -0,0 +1,196 @@
...
+	if cuc == nil {
+		return nil
+	}
+
+	// Create a map to hold the aggregated data
+	aggregated := make(map[string]*UpdateMetricsRequest)
+
+	for _, req := range reqs {
+		// Use the combination of RecordDate, Hash, Chain, and ApiType as the key
+		key := req.RecordDate + req.Hash + req.Chain + req.ApiType
```
ranlavanet, line 170: always good practice to have a method generate the key. pass the request and get the key back.

protocol/metrics/consumer_relayserver_client.go
```diff
@@ -0,0 +1,196 @@
...
+	return cuc
+}
+
+func (cuc *ConsumerRelayserverClient) processQueue() {
+	if cuc == nil {
+		return
+	}
+
+	for range cuc.ticker.C {
+		cuc.lock.Lock()
```
ranlavanet, line 60: thats a really bad usage of locks. 

best practice of locks is to have wrapper methods that lock and defer unlock. 
do what you need in the method and return. 
there shouldn't be multiple locking and unlocking in the same function body that's a garenteed deadlock recepie

protocol/metrics/consumer_relayserver_client.go
```diff
@@ -0,0 +1,196 @@
...
+		if !cuc.isSendQueueRunning && len(cuc.addQueue) > 0 {
+			cuc.sendQueue, cuc.addQueue = cuc.addQueue, make([]UpdateMetricsRequest, 0)
+			cuc.isSendQueueRunning = true
+			cuc.sendID++
+			utils.LavaFormatDebug(fmt.Sprintf("CUC: Swapped queues (sendQueue length: %d) and set isSendQueueRunning to true - CCC iter:%d.", len(cuc.sendQueue), cuc.sendID))
+			cuc.lock.Unlock()
+
+			utils.LavaFormatDebug(fmt.Sprintf("CUC: Sending requests (sendQueue length: %d) - async start - CCC iter:%d.", len(cuc.sendQueue), cuc.sendID))
```
ranlavanet, line 68: reading from cuc while not locked is not allowed. 
better use of locks would be change the locks to a read write lock sync.RWMutex 
and read lock when can. reading is allowed in parallel by the lock itself so multiple reads are faster.

protocol/metrics/consumer_relayserver_client.go
```diff
@@ -0,0 +1,196 @@
...
+			cuc.sendRequests(cuc.sendQueue)
```
ranlavanet, line 69: reading while not locked.

protocol/metrics/consumer_relayserver_client.go
```diff
@@ -0,0 +1,196 @@
...
+			utils.LavaFormatDebug(fmt.Sprintf("CUC: Sending requests - async end - CCC iter:%d.", cuc.sendID))
```
ranlavanet, line 70: reading while not locked.

protocol/metrics/consumer_relayserver_client.go
```diff
@@ -0,0 +1,196 @@
...
+	jsonData, err := json.Marshal(aggregatedRequests)
+	if err != nil {
+		utils.LavaFormatDebug("CUC: Error marshalling sendQueue to JSON")
+		return
+	}
+
+	var resp *http.Response
+	for i := 0; i < 3; i++ {
+		utils.LavaFormatDebug(fmt.Sprintf("CUC: Attempting to post request - Attempt: %d", i+1))
+		resp, err = client.Post(cuc.endPointAddress+"/updateMetrics", "application/json", bytes.NewBuffer(jsonData))
```
ranlavanet, line 141: reading while not locked.

protocol/metrics/consumer_relayserver_client.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+
+	if err != nil {
+		utils.LavaFormatDebug("CUC: Failed to send requests after 3 attempts")
+		return
+	}
+
+	defer resp.Body.Close()
+
+	utils.LavaFormatDebug(fmt.Sprintf("CUC: Successfully sent requests - CCC iter:%d.", cuc.sendID))
```
ranlavanet, line 157: reading while not locked.

ranlavanet (changes requested):

protocol/rpcconsumer/rpcconsumer.go
OrigamiMike: i would first delete some other code - and then make the struct - there are things you can delete after this new relay server works

oneeuroquestion: there is some code we can delete before - all the old relay server
then it will be easier to see which struct will be best

protocol/metrics/consumer_relayserver_client.go
```diff
@@ -31,18 +32,18 @@ type UpdateMetricsRequest struct {
...
 	}
 
-	cuc := &ConsumerRelayserverClient{
+	cuc := &ConsumerRelayServerClient{
 		endPointAddress: endPointAddress,
 		ticker:          time.NewTicker(30 * time.Second),
 		sendQueue:       make([]UpdateMetricsRequest, 0),
 		addQueue:        make([]UpdateMetricsRequest, 0),
-		lock:            sync.Mutex{},
+		lock:            sync.RWMutex{},
```
ranlavanet, line 46: FYI no need to initialize locks.

protocol/metrics/consumer_relayserver_client.go
```diff
@@ -51,7 +52,7 @@ func NewConsumerRelayserverClient(endPointAddress string) *ConsumerRelayserverCl
 	return cuc
 }
 
-func (cuc *ConsumerRelayserverClient) processQueue() {
+func (cuc *ConsumerRelayServerClient) processQueue() {
```
ranlavanet, line 55: refactor this method, its kinda. dead locks can kill our service so that's why I'm very specific on how it should look. we had too many in the past and its hell to bebug 

create an inner method. 
```go 
func (cuc *ConsumerRelayServerClient) shouldSendInfo() bool {
    cuc.lock.RLock()
    defer cuc.lock.RUnlock()
    return !cuc.isSendQueueRunning && len(cuc.addQueue) > 0
}
```

now on main flow:

if cuc.shouldSendInfo() {
    ....
    ....
} 

every time you need to read do an rlock with a defer 
every time you need to write lock. 

if you need a flow with multiple reads and write just lock the entire flow. 

best practice is a method starting with lock + defer unlock and do what ever you need inside.

protocol/metrics/consumer_relayserver_client.go
```diff
@@ -135,13 +142,25 @@ func (cuc *ConsumerRelayserverClient) sendRequests(sendQueue []UpdateMetricsRequ
 		return
 	}
 
+	cuc.lock.RLock()
+	var cucEndpointAddress string = cuc.endPointAddress
```
ranlavanet, line 146: helper function. 
read lock, defer read unlock return info.

protocol/metrics/consumer_relayserver_client.go
```diff
@@ -154,10 +173,13 @@ func (cuc *ConsumerRelayserverClient) sendRequests(sendQueue []UpdateMetricsRequ
 
 	defer resp.Body.Close()
 
+	cuc.lock.RLock()
```
ranlavanet, line 176: read all the info you need at the beginning its not changing anyway

protocol/rpcconsumer/rpcconsumer.go
```diff
@@ -97,12 +97,12 @@ type RPCConsumer struct {
 }
 
 // spawns a new RPCConsumer server with all it's processes and internals ready for communications
-func (rpcc *RPCConsumer) Start(ctx context.Context, txFactory tx.Factory, clientCtx client.Context, rpcEndpoints []*lavasession.RPCEndpoint, requiredResponses int, cache *performance.Cache, strategy provideroptimizer.Strategy, metricsListenAddress string, maxConcurrentProviders uint, usageServerAddr string) (err error) {
+func (rpcc *RPCConsumer) Start(ctx context.Context, txFactory tx.Factory, clientCtx client.Context, rpcEndpoints []*lavasession.RPCEndpoint, requiredResponses int, cache *performance.Cache, strategy provideroptimizer.Strategy, metricsListenAddress string, maxConcurrentProviders uint, relayServerAddr string) (err error) {
```
ranlavanet, line 100: lets create a struct that passes both relay server address and metrics listen address ? will be more readable and easier to add more stuff in the future 

```go
type AnalyticsListenAddressess struct {
    MetricsListenAddress string 
    RelayServerAddress string
}
```

ranlavanet (changes requested):

scripts/pre_setups/init_lava_only_with_node.sh
```diff
@@ -50,11 +50,18 @@ screen -d -m -S provider1 bash -c "source ~/.bashrc; lavap rpcprovider \
 $PROVIDER1_LISTENER LAV1 rest '$LAVA_REST' \
 $PROVIDER1_LISTENER LAV1 tendermintrpc '$LAVA_RPC,$LAVA_RPC' \
 $PROVIDER1_LISTENER LAV1 grpc '$LAVA_GRPC' \
-$EXTRA_PROVIDER_FLAGS --geolocation 1 --log_level debug --from servicer1 --chain-id lava --metrics-listen-address ":7776" 2>&1 | tee $LOGS_DIR/PROVIDER1.log" && sleep 0.25
```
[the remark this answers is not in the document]
oneeuroquestion: it's back now - i made my own script

protocol/metrics/consumer_relayserver_client.go
oneeuroquestion: done

protocol/metrics/consumer_relayserver_client.go
ranlavanet: what about this?

protocol/metrics/consumer_relayserver_client.go
```diff
@@ -0,0 +1,241 @@
...
+		Hash:            relayMetric.ProjectHash,
+		Chain:           relayMetric.ChainID,
+		ApiType:         relayMetric.APIType,
+		RelaysInc:       1,
+		CuInc:           int(relayMetric.ComputeUnits),
+		LatencyToAdd:    uint64(relayMetric.Latency),
+		LatencyAvgCount: 1,
+	}
+
+	cuc.lock.Lock()
```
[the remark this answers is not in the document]
ranlavanet: what about this?

ranlavanet (changes requested):

protocol/metrics/consumer_relayserver_client.go
```diff
@@ -0,0 +1,239 @@
...
+			cuc.lock.Unlock()
+		}()
+	} else {
+		utils.LavaFormatDebug(fmt.Sprintf("CUC: Skipped this iteration - CCC iter:%d.", cuc.sendID))
+	}
+}
+
+func (cuc *ConsumerRelayServerClient) SetRelayMetrics(relayMetric *RelayMetrics) error {
+	if cuc == nil {
+		utils.LavaFormatWarning("CUC: Warning - CUC is nil. called from SetRelayMetrics", nil)
```
ranlavanet, line 101: this is gonna spam our logs like hell. every relay. please clean all logs from nil case. 
please clean all logs that might spam from warning / error levels ( debug you can leave but only if its not too spammy)

oneeuroquestion: ok - this is easy to remove

ranlavanet (approved):

PR #1103, opened 2024-01-04
CNS: Make sure we do not modulo by zero

shleikes:


x/epochstorage/keeper/params.go
```diff
@@ -107,12 +107,23 @@ func (k Keeper) BlocksToSave(ctx sdk.Context, block uint64) (res uint64, erro er
...
-	epochStartInGrid := fixtedParams.FixationBlock // fixation block is always <= block
+	fixatedParams, err := k.GetFixatedParamsForBlock(ctx, string(types.KeyEpochBlocks), block)
+	if err != nil {
+		return 0, err
+	}
+
+	var epochBlocks uint64
+	utils.Deserialize(fixatedParams.Parameter, &epochBlocks)
+
+	if epochBlocks == 0 {
```
omerlavanet, line 118: we don't need to validate a param on every action if we have validate basic that protects from 0

omerlavanet (changes requested):

x/epochstorage/keeper/params.go
oren-lava: just to clarify Omer's comment, the check for 0 already exists in validateEpochBlocks()

oren-lava (dismissed):

x/epochstorage/keeper/params.go
shleikes: Talked to @omerlavanet about this. Decided to leave it as it is.

Yaroms (approved):

PR #1134, opened 2024-01-11
CNS-806: align subscription deletion and CU tracker start to next epoch

oren-lava:
Almost every subscription related state change was applied in the next epoch. Now I fix the loose ends. Both subscription deletion (upon expiry) and CU tracker timer start will be from the next epoch.


x/subscription/keeper/subscription.go
```diff
@@ -687,7 +687,7 @@ func (k Keeper) DelProjectFromSubscription(ctx sdk.Context, consumer, name strin
 func (k Keeper) delAllProjectsFromSubscription(ctx sdk.Context, consumer string) {
 	allProjectsIDs := k.projectsKeeper.GetAllProjectsForSubscription(ctx, consumer)
 	for _, projectID := range allProjectsIDs {
-		err := k.projectsKeeper.DeleteProject(ctx, consumer, projectID)
+		err := k.projectsKeeper.PurgeProject(ctx, projectID)
```
Yaroms, line 690: what happens on an error? this is called in begin block

oren-lava: on the next line below you can see that it's simply printed. Should I change anything regarding this?

x/projects/keeper/creation.go
```diff
@@ -151,6 +156,45 @@ func (k Keeper) DeleteProject(ctx sdk.Context, creator, projectID string) error
 	return k.projectsFS.DelEntry(ctx, project.Index, nextEpoch)
 }
 
+// PurgeProject forces the deletion of projects of an expired subscription (happens immediately)
+func (k Keeper) PurgeProject(ctx sdk.Context, projectID string) error {
```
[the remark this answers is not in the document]
oren-lava: done. see https://github.com/lavanet/lava/commit/23843b75aab78fa5e9756f52ef30318991920279

x/projects/keeper/creation.go
```diff
@@ -85,8 +85,13 @@ func (k Keeper) CreateProject(ctx sdk.Context, subAddr string, projectData types
...
 
 	// project name per subscription is unique: check for duplicates
+	// we also check !isDeleted because when a new subscription is created
+	// in the same epoch that a subscription from the same creator was expired,
+	// we'll find the old admin project but since it's deleted we want to ignore
+	// it and allow creating a new admin project
 	var emptyProject types.Project
-	if found := k.projectsFS.FindEntry(ctx, project.Index, epoch, &emptyProject); found {
+	_, isDeleted, _, found := k.projectsFS.FindEntryDetailed(ctx, project.Index, epoch, &emptyProject)
+	if found && !isDeleted {
```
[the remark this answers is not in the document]
oren-lava: yes there will be two entries with the same index. The differences between them:
Entry A: `Refcount`=0, `DeleteAt`=blockWithinTheEpoch, `StaleAt`= blockWithinTheEpoch+stalePeriod, `IsLatest`=false
Entry B: `Refcount`=1, `DeleteAt`=MaxUint64, `StaleAt`= MaxUint64, `IsLatest`=true

since entry B is the latest, it will be found first in every query. All the unit tests pass but I do admit that it may require extra care. @Yaroms what do you think?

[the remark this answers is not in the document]
oren-lava: After talks with @Yaroms the whole purpose of this PR changed. Now all deletions are triggered from the next epoch, so this code doesn't exist anymore.

x/subscription/keeper/cu_tracker.go
```diff
@@ -243,10 +220,9 @@ func (k Keeper) RewardAndResetCuTracker(ctx sdk.Context, cuTrackerTimerKeyBytes
 			utils.LogLavaEvent(ctx, k.Logger(ctx), types.MonthlyCuTrackerProviderRewardEventName, map[string]string{
 				"provider":         provider,
 				"sub":              sub,
-				"plan":             subObj.PlanIndex,
 				"tracked_cu":       strconv.FormatUint(trackedCu, 10),
 				"credit_used":      creditToSub.String(),
-				"credit_remaining": subObj.Credit.String(),
+				"credit_remaining": timerData.Credit.String(),
```
Yaroms, line 225: is it the remaining credit?

oren-lava: done. see https://github.com/lavanet/lava/commit/258ba6c80e921517193f08b73232259ec4a942a5

Yaroms (approved):

PR #1137, opened 2024-01-11
PRT-1078: Fix SDK crash on EMPTY PARSER_FUNC

shleikes:


ecosystem/lava-sdk/src/parser/parser.ts
```diff
@@ -492,8 +490,11 @@ export class Parser {
       case PARSER_FUNC.DEFAULT:
         retval = this.parseDefault(blockParser.getParserArgList());
         break;
+      case PARSER_FUNC.EMPTY:
```
ranlavanet, line 493: now this flow prints an error. Empty parser function is not an error. its a valid option. we just skip parsing. 
add to each API Interface if the instance of the error is the empty parser func error type dont print the error just st the requested block to NOT_APPLICABLE.

ranlavanet (changes requested):

ranlavanet (approved):

PR #1140, opened 2024-01-14
Lava spec update / Updating script 

nimrod-teich:


scripts/automation_scripts/update_lava_spec.py
```diff
@@ -0,0 +1,155 @@
...
+        interface_type = api_collection.get("collection_data", {}).get(
+            "api_interface", ""
+        )
+
+        if interface_type not in endpoints:
+            continue
+
+        for api in api_collection.get("apis", []):
+            api_name = api.get("name", "")
+            if "lavanet" in api_name:
```
Yaroms, line 29: instead of checking for lavanet (this only covers lava api's and is not generic)
lets use: not cosmwasm or not cosmos

Yaroms (approved):

PR #1148, opened 2024-01-18
CNS-817: uncapped repeated fields

oren-lava:
In this PR I capped the following TXs arguments:
- StakeProvider: max number of 5 endpoints per geolocation
- AddKeys/DelKeys: max number of 10 keys

All restrictions can be changed by changing the appropriate constant.
Note, other repeated fields in TXs were not changed as instructed by @Yaroms.

### DIFF
--- a/x/pairing/keeper/staking.go
+++ b/x/pairing/keeper/staking.go
@@ -56,6 +56,29 @@ func (k Keeper) StakeNewEntry(ctx sdk.Context, validator, creator, chainID strin
 			utils.Attribute{Key: "geolocation", Value: geolocation},
 		)
 	}
+
+	// validate there are no more than 5 endpoints per geolocation
+	endp_geo_hist := map[int32]int{}
+	for _, endp := range endpointsVerified {
+		counter, ok := endp_geo_hist[endp.Geolocation]
+		if !ok {
+			endp_geo_hist[endp.Geolocation] = 1
+		} else {
+			counter++
+			if counter > types.MAX_ENDPOINTS_AMOUNT_PER_GEO {
+				return utils.LavaFormatWarning("stake provider failed", fmt.Errorf("number of endpoint for geolocation exceeded limit"),
+					utils.LogAttr("creator", creator),
+					utils.LogAttr("chain_id", chainID),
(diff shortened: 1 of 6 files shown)


x/pairing/keeper/staking.go
```diff
@@ -56,6 +56,29 @@ func (k Keeper) StakeNewEntry(ctx sdk.Context, validator, creator, chainID strin
 			utils.Attribute{Key: "geolocation", Value: geolocation},
 		)
 	}
+
+	// validate there are no more than 5 endpoints per geolocation
+	endp_geo_hist := map[int32]int{}
```
Yaroms, line 61: do it much more simple:
if len(enpoints) > len(geolocations)*5
     error

x/projects/keeper/msg_server_add_project_keys.go
```diff
@@ -18,6 +18,13 @@ func (k msgServer) AddKeys(goCtx context.Context, msg *types.MsgAddKeys) (*types
 		)
 	}
 
+	if len(msg.ProjectKeys) > types.MAX_KEYS_AMOUNT {
```
Yaroms, line 21: isnt this a double check since we already did it in k.AddKeysToProject?

x/pairing/keeper/staking.go
oren-lava: this is a different restriction.
a user can have 6 endpoints in US and one in EU - thus bypassing the 5 endpoints per geolocation restriction but passing this if

x/projects/keeper/msg_server_add_project_keys.go
oren-lava: removed. see https://github.com/lavanet/lava/commit/df5a276856b7e6bdf5405bc56de276d195ddde62

x/pairing/keeper/staking.go
oren-lava: done. see https://github.com/lavanet/lava/commit/0f853db7a5ebf23bf8455e136b67aa18ff476d8e

Yaroms (approved):

PR #1153, opened 2024-01-21
CNS-631-add-ibc-modules

Yaroms:


app/app.go
```diff
@@ -556,6 +572,7 @@ func New(
 		// TODO: Check alternatives to this
 		// AddRoute(distrtypes.RouterKey, distr.NewCommunityPoolSpendProposalHandler(app.DistrKeeper)).
 		AddRoute(upgradetypes.RouterKey, upgrade.NewSoftwareUpgradeProposalHandler(&app.UpgradeKeeper)).
+		// AddRoute(ibcexported.RouterKey, ibcclient.NewClientProposalHandler(app.IBCKeeper.ClientKeeper))
```
oren-lava, line 575: should it be in a comment? these lines didn't exist before

scripts/init_chain.sh
```diff
@@ -11,7 +11,8 @@ if ! command_exists jq; then
 fi
 
 rm -rf ~/.lava
-lavad init validator --chain-id lava
+chainID="lava-local-1"
```
oren-lava, line 14: why? what was wrong with the original chain-id "lava"?

scripts/relayer/init_chain_2.sh
```diff
@@ -0,0 +1,123 @@
+#!/bin/bash
```
oren-lava, line 1: I'm guessing this is used to start another node in parallel to a running node. Can you put a small comment here to the script usage for future reference?

scripts/relayer/init_chain_2.sh
```diff
@@ -0,0 +1,123 @@
...
+        | jq '.app_state.gov.params.voting_period = "4s"' \
+        | jq '.app_state.gov.params.expedited_voting_period = "3s"' \
+        | jq '.app_state.gov.params.expedited_min_deposit[0].denom = "ulava"' \
+        | jq '.app_state.gov.params.expedited_min_deposit[0].amount = "200"' \
+        | jq '.app_state.gov.params.expedited_threshold = "0.67"' \
+        | jq '.app_state.mint.params.mint_denom = "ulava"' \
+        | jq '.app_state.staking.params.bond_denom = "ulava"' \
+        | jq '.app_state.crisis.constant_fee.denom = "ulava"' \
+        | jq '.app_state.epochstorage.params.epochsToSave = "5"' \
+        | jq '.app_state.epochstorage.params.epochBlocks = "4"' \
```
oren-lava, line 42: add downtime_duration=1s

scripts/relayer/init_relay.sh
```diff
@@ -0,0 +1,24 @@
+#!/bin/bash
```
oren-lava, line 1: can you add a comment about what this script does?

oren-lava (changes requested):

app/app.go
Yaroms: yes, currently we dont enable this, maybe in the future

scripts/init_chain.sh
Yaroms: omer want to differentiate it from mainnet chain id so there wont be mix ups

scripts/relayer/init_chain_2.sh
Yaroms: done

scripts/relayer/init_chain_2.sh
Yaroms: done

scripts/init_chain.sh
omerlavanet: we can do lava since mainnet isn't lava-1

PR #1162, opened 2024-01-23
CNS-829: fix disableHooks bug

oren-lava:
keep the DisableDualstakingHook var on state and not as a global


Yaroms (approved):

Yaroms (changes requested):
one more thing, can you make a migrator that sets the flag to false?
this is to make sure we dont get a panic if we do an upgrade

Yaroms (approved):

PR #1169, opened 2024-01-24
Added compound and save add-on

sidlavanet:


cookbook/specs/spec_add_sqdsubgraph.json
```diff
@@ -96,11 +96,43 @@
                             "api_interface": "rest",
                             "internal_path": "",
                             "type": "POST",
-                            "add_on": "convex"
+                            "add_on": "compound-v3"
                         },
                         "apis": [
                             {
-                                "name": "/subgraphs/name/convex",
+                                "name": "/subgraphs/name/compound-v3",
```
Yaroms, line 103: is covex not supported anymore?

cookbook/specs/spec_add_sqdsubgraph.json
```diff
@@ -96,11 +96,43 @@
...
+                        "parse_directives": [],
+                        "verifications": []
+                    },
+                    {
+                        "enabled": true,
+                        "collection_data": {
+                            "api_interface": "rest",
+                            "internal_path": "",
+                            "type": "POST",
+                            "add_on": "aave-v3"
```
Yaroms, line 131: do we need 2 different addons for these apis?

cookbook/specs/spec_add_sqdsubgraph.json
sidlavanet: Yes, we are going with Aave and compound for now

cookbook/specs/spec_add_sqdsubgraph.json
sidlavanet: Omer suggested that we add these subgraphs as add-ons. @omerlavanet anything you would like to add here?

Yaroms: the question is, do node providers need to do something special to support this api? or does all the nodes have them by defualt?
for example trace addon is an additional config for providers to support extra apis than regular node.

cookbook/specs/spec_add_sqdsubgraph.json
```diff
@@ -96,11 +96,43 @@
                             "api_interface": "rest",
                             "internal_path": "",
                             "type": "POST",
-                            "add_on": "convex"
+                            "add_on": "compound-v3"
```
Yaroms, line 99: example:
curl --request POST \
  --url https://sqd-subgraph.net/subgraphs/name/papercliplabs/compound-v3 \
  --header 'Content-Type: application/json' \
  --data '{"query":"{\n  tokens(first: 5) {\n    id\n    name\n    symbol\n    decimals\n  }\n}"}'

PR #1178, opened 2024-01-28
CNS-825: pruning verifications to non cosmos/eth chains

oren-lava:
Didn’t do the following specs:

Starknet - used starknet_getBlockWithTxHashes for block 0 and regular RPC node worked - there is no difference between archive and regular node.

Moralis, Solana, Koii, Subsquid - no RPC API that returns a different result for archive (two cases: doesn’t exist or return an error for old block)


cookbook/specs/spec_add_aptos.json
```diff
@@ -455,6 +455,39 @@
                                         "expected_value": "1"
                                     }
                                 ]
+                            },
+                            {
+                                "name": "pruning",
```
Yaroms, line 460: set this to warning severity and later we will make it error.
that will make sure that if we mess with the verifications providers wont go down

oren-lava: done. see https://github.com/lavanet/lava/commit/048b31fa857659cd70dc4d7057a13d45466c3b5a

Yaroms (approved):

PR #1182, opened 2024-01-28
PRT-1103: Health test - add support for add-ons and extensions

shleikes:


protocol/monitoring/types.go
```diff
@@ -0,0 +1,15 @@
+package monitoring
+
+import "strconv"
+
+type HealthRPCEndpoint struct {
+	NetworkAddress string   `yaml:"network-address,omitempty" json:"network-address,omitempty" mapstructure:"network-address"` // HOST:PORT
```
ranlavanet, line 6: use embedding instead 

```go

type bobo struct {
	a string
}

type bebe struct {
	*bobo
}

func TestCalculateAvailabilityScore2(t *testing.T) {
	a := &bobo{a: "a"}
	fmt.Println(a.a)
}

```

out: "a"

ranlavanet (changes requested):

ranlavanet (approved):

PR #1183, opened 2024-01-30
CNS-840: archive pruning verifications for rest and grpc

oren-lava:
In this PR, I added support for pruning verifications for rest and grpc API interfaces in Cosmos-based chains. This was not supported previously because unlike jsonrpc, rest and grpc can't get an "earliest" argument to the "getBlockByNumber" API.

All verifications have a "Warning" severity and all were tested locally.


cookbook/specs/spec_add_cosmossdk.json
```diff
@@ -4299,6 +4333,20 @@
                                     "encoding": "hex"
                                 },
                                 "api_name": "block"
+                            },
+                            {
+                                "function_template": "{\"jsonrpc\":\"2.0\",\"method\":\"status\",\"params\":[],\"id\":1}",
```
Yaroms, line 4338: we set up the whole thing here and it is never used except for the function_tag.
we need to find a more elegant place to mark that it supports earliest

cookbook/specs/spec_add_ethereum.json
```diff
@@ -1014,6 +1014,19 @@
                                     "encoding": "hex"
                                 },
                                 "api_name": "eth_getBlockByNumber"
+                            },
```
Yaroms, line 1017: same here

Yaroms (commented):
this is a good step, lets find another way to do this more elegant

Yaroms:
consulted with omer, we decided that instead of defining the parse directive twice, in the verification parse directive we should set only the function tag as "earliest" and in the code, we go fetch that parse directive from the parse_directives list in the apicollection

protocol/chainlib/chain_fetcher.go
```diff
@@ -116,6 +118,23 @@ func (cf *ChainFetcher) Verify(ctx context.Context, verification VerificationCon
 		}...)
 		return nil
 	}
+
+	// craft data for GET_BLOCK_BY_NUM verification that cannot use "earliest"
+	// also check for %d because the data constructed assumes its presence
+	if verification.ParseDirective.FunctionTag == spectypes.FUNCTION_TAG_GET_BLOCK_BY_NUM {
+		if verification.LatestDistance != 0 && latestBlock != 0 {
+			if latestBlock >= verification.LatestDistance {
+				data = []byte(fmt.Sprintf(parsing.FunctionTemplate, latestBlock-verification.LatestDistance))
```
omerlavanet, line 127: we are not checking for %d

omerlavanet (changes requested):

protocol/chainlib/base_chain_parser.go
```diff
@@ -402,6 +402,9 @@ func getServiceApis(spec spectypes.Spec, rpcInterface string) (retServerApis map
 						Addon:     apiCollection.CollectionData.AddOn,
 					}
 
+					if verification.ParseDirective.FunctionTag != spectypes.FUNCTION_TAG_VERIFICATION {
+						verification.ParseDirective = taggedApis[verification.ParseDirective.FunctionTag].Parsing
```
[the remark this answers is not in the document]
Yaroms: a valid spec should always have this. this is tested on chain.

protocol/chainlib/chain_fetcher.go
Yaroms: FUNCTION_TAG_GET_BLOCK_BY_NUM will always have %d, this check is never done in other places in the provider.
should i still add it?

protocol/chainlib/chain_fetcher.go
```diff
@@ -140,7 +162,7 @@ func (cf *ChainFetcher) Verify(ctx context.Context, verification VerificationCon
 			{Key: "Response", Value: string(reply.Data)},
 		}...)
 	}
-	if verification.LatestDistance != 0 && latestBlock != 0 {
+	if verification.LatestDistance != 0 && latestBlock != 0 && verification.ParseDirective.FunctionTag == spectypes.FUNCTION_TAG_GET_EARLIEST_BLOCK {
```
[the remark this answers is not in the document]
Yaroms: done

protocol/chainlib/chain_fetcher.go
```diff
@@ -116,6 +118,23 @@ func (cf *ChainFetcher) Verify(ctx context.Context, verification VerificationCon
...
+			if latestBlock >= verification.LatestDistance {
+				data = []byte(fmt.Sprintf(parsing.FunctionTemplate, latestBlock-verification.LatestDistance))
+			} else {
+				return utils.LavaFormatWarning("[-] verify failed getting non-earliest block for chainMessage", fmt.Errorf("latestBlock is smaller than latestDistance"),
+					utils.LogAttr("path", path),
+					utils.LogAttr("latest_block", latestBlock),
+					utils.LogAttr("Latest_distance", verification.LatestDistance),
+				)
+			}
+		}
```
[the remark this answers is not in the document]
Yaroms: should the verification fail due to bad spec?

protocol/chainlib/base_chain_parser.go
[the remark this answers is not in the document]
omerlavanet: and you can guarantee that forever in the future there are no bugs in the protocol code when constructing tagged apis object? does it cost a lot to add a check?

protocol/chainlib/chain_fetcher.go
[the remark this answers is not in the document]
omerlavanet: yes

protocol/chainlib/chain_fetcher.go
omerlavanet: yes, add it in the consensus, then you don't need it here

protocol/chainlib/base_chain_parser.go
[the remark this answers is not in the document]
Yaroms: what to do if it doesnt exist?

protocol/chainlib/chain_fetcher.go
Yaroms: done

protocol/chainlib/chain_fetcher.go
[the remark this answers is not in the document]
Yaroms: done

protocol/chainlib/base_chain_parser.go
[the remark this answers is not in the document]
Yaroms: done

protocol/chainlib/chain_fetcher.go
```diff
@@ -78,9 +78,8 @@ func (cf *ChainFetcher) Validate(ctx context.Context) error {
 				}
 			}
 			if err != nil {
-				err := utils.LavaFormatError("invalid Verification on provider startup", err, utils.Attribute{Key: "Addons", Value: addons}, utils.Attribute{Key: "verification", Value: verification.Name})
 				if verification.Severity == spectypes.ParseValue_Fail {
-					return err
+					return utils.LavaFormatError("invalid Verification on provider startup", err, utils.Attribute{Key: "Addons", Value: addons}, utils.Attribute{Key: "verification", Value: verification.Name})
```
omerlavanet, line 82: verifications can run not on provider startup, why did you add these attributes? aren't they added outside?

x/spec/types/spec.go
```diff
@@ -96,6 +96,11 @@ func (spec Spec) ValidateSpec(maxCU uint64) (map[string]string, error) {
 					return details, fmt.Errorf("unsupported api encoding %s in apiCollection %v ", parsing.ResultParsing.Encoding, apiCollection.CollectionData)
 				}
 			}
+			if parsing.FunctionTag == FUNCTION_TAG_GET_BLOCK_BY_NUM {
+				if !strings.Contains(parsing.FunctionTemplate, "%") {
```
omerlavanet, line 100: % and not %d?

omerlavanet (changes requested):

x/spec/types/spec.go
Yaroms: yes, there might be %x

protocol/chainlib/chain_fetcher.go
Yaroms: they were there, i just moved the error print only if the is fail

Yaroms: these attributes are not added anywhere

omerlavanet (approved):

PR #1188, opened 2024-02-04
CNS-848: when staking, add to delegate total past delegations

oren-lava:
Fix a bug in which when a provider unstakes and then re-stakes, we didn't consider past delegations to it.

Now, when a provider stakes, we check the delegationFS for existing delegations and add it to its DelegateTotal field

### DIFF
--- a/x/pairing/keeper/staking.go
+++ b/x/pairing/keeper/staking.go
@@ -145,12 +145,19 @@ func (k Keeper) StakeNewEntry(ctx sdk.Context, validator, creator, chainID strin
 
 	// if there are registered delegations to the provider, count them in the delegateTotal
 	delegateTotal := sdk.ZeroInt()
-	epoch := k.epochStorageKeeper.GetEpochStart(ctx)
-	delegations, err := k.dualstakingKeeper.GetProviderDelegators(ctx, senderAddr.String(), epoch)
+	nextEpoch, err := k.epochStorageKeeper.GetNextEpoch(ctx, uint64(ctx.BlockHeight()))
+	if err != nil {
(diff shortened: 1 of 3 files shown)


x/pairing/keeper/staking.go
```diff
@@ -143,6 +143,25 @@ func (k Keeper) StakeNewEntry(ctx sdk.Context, validator, creator, chainID strin
 		{Key: "geolocation", Value: geolocation},
 	}
 
+	// if there are registered delegations to the provider, count them in the delegateTotal
+	delegateTotal := sdk.ZeroInt()
+	epoch := k.epochStorageKeeper.GetEpochStart(ctx)
```
Yaroms, line 148: use the next epoch instead of current one

oren-lava: done. see https://github.com/lavanet/lava/pull/1188/commits/ed65639f3db931943b5b007ffea26726ce1b5e93

x/pairing/keeper/staking.go
```diff
@@ -143,6 +143,32 @@ func (k Keeper) StakeNewEntry(ctx sdk.Context, validator, creator, chainID strin
 		{Key: "geolocation", Value: geolocation},
 	}
 
+	// if there are registered delegations to the provider, count them in the delegateTotal
+	delegateTotal := sdk.ZeroInt()
+	nextEpoch, err := k.epochStorageKeeper.GetNextEpoch(ctx, uint64(ctx.BlockHeight()))
```
Yaroms, line 148: return and fail the tx on error

Yaroms (approved):

PR #1212, opened 2024-02-12
PRT-1135: Fix the error that caused the health test to fail (init_chain_commands)

shleikes:


cookbook/projects/policy_eth_with_addon.yml
```diff
@@ -0,0 +1,9 @@
+Policy:
+  selected_providers_mode: ALLOWED
+  chain_policies:
```
omerlavanet, line 3: you disabled all other chains, that needs to be fixed

omerlavanet (changes requested):

cookbook/projects/policy_eth_with_addon.yml
shleikes: I thought that this is the purpose of `selected_providers_mode`.

image

omerlavanet: it's not, that flag is on how we select providers not how we define the chains we can use, look at the policy we are using it has a wildcard

omerlavanet (approved):

PR #1218, opened 2024-02-12
CNS-884-audit-validate-chainid-on-delegation-key

Yaroms:


common/types/ascii.go
```diff
@@ -59,6 +60,10 @@ func ValidateString(s string, restrictType charRestrictionEnum, disallowedChars
 				if !unicode.IsLetter(r) && r != ' ' && r != '_' && !unicode.IsDigit(r) {
 					return false
 				}
+			case INDEX_RESTRICTIONS:
+				if !unicode.IsLetter(r) && !unicode.IsDigit(r) {
```
oren-lava, line 64: why not only capital letters?

oren-lava (changes requested):

common/types/ascii.go
Yaroms: why limit?

oren-lava (approved):

PR #1227, opened 2024-02-14
CNS-869: IPRPC part 3 - IPRPC fund TX

oren-lava:
In this PR I implemented the `iprpc-fund` TX.

This TX allows users to fund the IPRPC pool with any kind of tokens to be vested as IPRPC rewards for a selected number of months.

Note, this PR doesn't contain unit tests. They will be implemented later on.


x/rewards/keeper/iprpc.go
```diff
@@ -0,0 +1,72 @@
...
+	err = k.bankKeeper.SendCoinsFromAccountToModule(ctx, addr, string(types.IprpcPoolName), fund)
+	if err != nil {
+		return utils.LavaFormatError(types.ErrFundIprpc.Error()+"for funding iprpc pool", err,
+			utils.LogAttr("creator", creator),
+			utils.LogAttr("fund", fund.String()),
+		)
+	}
+
+	// check if to add funds from next month or to current month
+	fromNextMonth := false
```
Yaroms, line 63: fromNextMonth = utils.IsMiddleOfMonthPassed(ctx.BlockTime())

x/rewards/keeper/iprpc_reward.go
```diff
@@ -119,12 +119,12 @@ func (k Keeper) addSpecFunds(ctx sdk.Context, spec string, fund sdk.Coins, durat
 			}
 			// did not find spec in IPRPC reward -> create a new one
 			iprpcReward.SpecFunds = append(iprpcReward.SpecFunds, types.Specfund{Spec: spec, Fund: fund})
+			k.SetIprpcReward(ctx, iprpcReward)
```
Yaroms, line 122: do it outside of the if

x/rewards/keeper/iprpc.go
oren-lava: outdated

x/rewards/keeper/iprpc_reward.go
oren-lava: outdated

x/rewards/keeper/iprpc.go
```diff
@@ -3,13 +3,71 @@ package keeper
...
+			utils.LogAttr("fund_ulava_amount", fund.AmountOf(k.stakingKeeper.BondDenom(ctx))),
+		)
+	}
+
+	// check creator has enough balance
+	addr, err := sdk.AccAddressFromBech32(creator)
+	if err != nil {
+		return utils.LavaFormatWarning("invalid creator address", types.ErrFundIprpc)
+	}
+	creatorUlavaBalance := k.bankKeeper.GetBalance(ctx, addr, k.stakingKeeper.BondDenom(ctx))
```
Yaroms, line 36: what about the other coins?

oren-lava: it doesn't matter. The IPRPC min cost is by definition in ulava. If the creator has enough ulava to pay for the min fee but doesn't have enough coins for the funding, it'll fail a few lines below in SendCoinsFromAccountToModule

Yaroms: than why check the balance for ulava?
it will fail on sending the coins

oren-lava: Assuming min iprpc cost is 100ulava
I want to fund the pool with 10ulava and I have 15ulava in my account.

If I dismiss the check for min iprpc price, the tx will wrongly pass 

My point is that it's not enough that you have the balance for the tx, you have to have enough balance for your desired find plus min iprpc cost

x/rewards/keeper/iprpc.go
```diff
@@ -2,12 +2,70 @@ package keeper
...
+func (k Keeper) FundIprpc(ctx sdk.Context, creator string, duration uint64, fund sdk.Coins, spec string) error {
+	// verify spec exists and active
+	foundAndActive, _, _ := k.specKeeper.IsSpecFoundAndActive(ctx, spec)
+	if !foundAndActive {
+		return utils.LavaFormatWarning("spec not found or disabled", types.ErrFundIprpc)
+	}
+
+	// check fund consists of minimum amount of ulava (duration * min_iprpc_cost)
+	minIprpcFundCost := k.GetMinIprpcCost(ctx).Amount.MulRaw(int64(duration))
+	if fund.AmountOf(k.stakingKeeper.BondDenom(ctx)).LT(minIprpcFundCost) {
```
omerlavanet, line 21: we should compare one month fund to one month min cost not one month fund to duration times min cost

omerlavanet: also please add a unitest that would find this by setting cost = mincost *1.1 and duration 2

omerlavanet (changes requested):

x/rewards/keeper/iprpc.go
Yaroms: will add the test to the unitests branch

PR #1228, opened 2024-02-15
CNS-871: IPRPC part 4 - Queries

oren-lava:
In this PR I implemented two queries:
- `iprpc-provider-rewards`: shows the estimated IPRPC rewards for a specific provider (relative to its serviced CU)
- `iprpc-spec-rewards`: shows all/specific spec of the IPRPC rewards

Note, this PR doesn't contain unit tests. They will be implemented later on.


proto/lavanet/lava/rewards/query.proto
```diff
@@ -31,6 +32,16 @@ service Query {
   rpc ShowIprpcData(QueryShowIprpcDataRequest) returns (QueryShowIprpcDataResponse) {
     option (google.api.http).get = "/lavanet/lava/rewards/show_iprpc_data";
   }
+
+  // IprpcProviderReward queries for a provider's current IPRPC reward (relative to its serviced CU)
+  rpc IprpcProviderReward(QueryIprpcProviderRewardRequest) returns (QueryIprpcProviderRewardResponse) {
```
Yaroms, line 37: i think we should make it clear it is only an estimation, lets add it to the names of the querie

x/rewards/keeper/base_pay.go
```diff
@@ -54,7 +54,7 @@ func (k Keeper) SetAllBasePay(ctx sdk.Context, list []types.BasePayGenesis) {
 	}
 }
 
-func (k Keeper) popAllBasePayForChain(ctx sdk.Context, chainID string) (list []types.BasePayWithIndex) {
+func (k Keeper) popAllBasePayForChain(ctx sdk.Context, chainID string, removeAfterPop bool) (list []types.BasePayWithIndex) {
```
omerlavanet, line 57: let's separate pop and get, this implementation of them together is risky

omerlavanet (changes requested):

x/rewards/keeper/base_pay.go
Yaroms: done

PR #1240, opened 2024-02-19
CNS-899 stargaze spec and deployment

Yaroms:


scripts/automation_scripts/update_lava_spec.py
```diff
@@ -3,7 +3,7 @@
 import os
 import re
 
-LAVA_PUBLIC_RPC = "public-rpc.lavanet.xyz:9090"
+LAVA_PUBLIC_RPC = "grpc-1.elgafar-1.stargaze-apis.com:26660"
```
omerlavanet, line 6: let's not put this here for everyone to see what we used

omerlavanet (changes requested):

PR #1241, opened 2024-02-19
PRT: Consumer "in <<<" logs truncate

shleikes:


utils/lavalog.go
```diff
@@ -317,3 +317,10 @@ func FormatStringerList[T fmt.Stringer](description string, listToPrint []T, sep
 	st = fmt.Sprintf(description+"\n\t%s", st)
 	return st
 }
+
+func FormatLongString(msg string, maxCharacters int) string {
+	if len(msg) > maxCharacters {
+		return msg[:maxCharacters] + "..."
```
omerlavanet, line 323: usually we keep last 20 characters something like abcsefg....truncated...uvwxyz

omerlavanet (changes requested):

utils/lavalog.go
shleikes: Fixed in a9b803d6d883758240ecb91cefe9d7e129bef59c

omerlavanet (approved):

PR #1261, opened 2024-02-28
PRT-1153: Consumer sometimes get non archive provider even though it specifically asked only for archive

shleikes:


ranlavanet (approved):

ranlavanet:
well done @shleikes . good catch

PR #1278, opened 2024-03-05
Send additional referrer usage information

nimrod-teich:


protocol/chainlib/common.go
```diff
@@ -383,15 +384,36 @@ type RefererData struct {
 	ReferrerClient *metrics.ConsumerReferrerClient
 }
 
-func (rd *RefererData) SendReferer(refererMatchString string) error {
+func (rd *RefererData) SendRefererForHttp(refererMatchString string, chainId string, headers map[string][]string) error {
```
ranlavanet, line 387: you can still use one method, pass the websocket pointer and check for nil to combine them both to the same flow :)

ranlavanet (changes requested):

ranlavanet (approved):

PR #1288, opened 2024-03-10
CNS-923: added local cache for validate pairing

oren-lava:
When iterating over relays, instead of validating the pairing for each relay, we keep track of the validated provider-consumer-spec trio in a local map and skip the validation if it was already validated in a previous relay

### DIFF
--- a/x/pairing/keeper/msg_server_relay_payment.go
+++ b/x/pairing/keeper/msg_server_relay_payment.go
@@ -22,11 +22,6 @@ type BadgeData struct {
 	BadgeSigner sdk.AccAddress
 }
 
-type PairingData struct {
-	allowedCU uint64
-	providers []epochstoragetypes.StakeEntry
-}
-
 func (k msgServer) RelayPayment(goCtx context.Context, msg *types.MsgRelayPayment) (*types.MsgRelayPaymentResponse, error) {
 	if len(msg.LatestBlockReports) > len(msg.Relays) {
 		return nil, utils.LavaFormatError("RelayPayment_invalid_latest_block_reports", fmt.Errorf("invalid latest block reports"),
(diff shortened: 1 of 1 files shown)


x/pairing/keeper/msg_server_relay_payment.go
```diff
@@ -169,24 +176,36 @@ func (k msgServer) RelayPayment(goCtx context.Context, msg *types.MsgRelayPaymen
...
+		var providers []epochstoragetypes.StakeEntry
+		allowedCU := uint64(0)
+		pairingData, ok := validatePairingCache[validatePairingKey]
+		if ok {
+			allowedCU = pairingData.allowedCU
+			providers = pairingData.providers
+		} else {
+			isValidPairing := false
+			isValidPairing, allowedCU, providers, err = k.Keeper.ValidatePairingForClient(
+				ctx,
```
Yaroms, line 190: can allowed CU change in between iterations? if so we need to recalculate it

oren-lava: done. see https://github.com/lavanet/lava/commit/e33ae32dd77042944191f696e87d51636ed548db

x/pairing/keeper/msg_server_relay_payment.go
```diff
@@ -271,6 +294,9 @@ func (k msgServer) RelayPayment(goCtx context.Context, msg *types.MsgRelayPaymen
 		rejected_relays_num--
 	}
 
+	// save all payment objects
+	paymentHandler.Flush(ctx)
```
oren-lava, line 298: wouldn't it be better if we put the flush after the `rejected_relays_num != 0` check? In other words, we don't wanna update the KVStore for a failed relay TX

x/pairing/keeper/unique_payment_storage_client_provider.go
```diff
@@ -73,8 +73,8 @@ func (k Keeper) IsDoubleSpend(ctx sdk.Context, chainID string, block uint64, pro
 	return found
 }
 
-func (k Keeper) GetConsumerFromUniquePayment(uniquePaymentStorageClientProvider *types.UniquePaymentStorageClientProvider) string {
-	key := uniquePaymentStorageClientProvider.Index
+func (k Keeper) GetConsumerFromUniquePayment(uniquePaymentStorageClientProvider string) string {
+	key := uniquePaymentStorageClientProvider
```
oren-lava, line 77: name the function's argument `key` and remove this

Yaroms (approved):

PR #1295, opened 2024-03-13
Add workflow for typo check 

miles-six:
I have made many typo corrections and would like to add a workflow for typos checks to prevent future typos if it works.


Yaroms:
Hey @miles-six thanks for your contribution and the time.
our policy is to use only go linters for spell checking and not open source library we are not familiar with.

PR #1300, opened 2024-03-13
PRT-1166 cache hash key migration.

ranlavanet:


protocol/rpcprovider/rpcprovider_server.go
```diff
@@ -766,7 +781,21 @@ func (rpcps *RPCProviderServer) TryRelay(ctx context.Context, request *pairingty
...
+					return
+				}
+				err = cache.SetEntry(new_ctx, &pairingtypes.RelayCacheSet{
+					RequestHash:      hash,
+					RequestedBlock:   requestedBlock,
+					BlockHash:        requestedBlockHash,
+					ChainId:          rpcps.rpcProviderEndpoint.ChainID,
+					Response:         copyReply,
+					Finalized:        finalized,
+					OptionalMetadata: ignoredMetadata,
```
omerlavanet, line 797: you forgot average block time and seen block

protocol/rpcprovider/rpcprovider_server.go
```diff
@@ -722,13 +722,28 @@ func (rpcps *RPCProviderServer) TryRelay(ctx context.Context, request *pairingty
...
+		if hashErr != nil {
+			utils.LavaFormatError("TryRelay Failed computing hash for cache request", hashErr)
+		} else {
+			cacheCtx, cancel := context.WithTimeout(ctx, common.CacheTimeout)
+			cacheReply, err = cache.GetEntry(cacheCtx, &pairingtypes.RelayCacheGet{
+				RequestHash:    hashKey,
+				RequestedBlock: request.RelayData.RequestBlock,
+				ChainId:        rpcps.rpcProviderEndpoint.ChainID,
+				BlockHash:      requestedBlockHash,
+				Finalized:      finalized,
```
omerlavanet, line 736: seen block missing here

omerlavanet (changes requested):

omerlavanet (changes requested):
let's add the following test too:
set up a provider, cache using chainFetcher, see that we hit the cache by requesting from a user (with seen block)

protocol/rpcprovider/rpcprovider_server.go
ranlavanet: agreed that its ok.

omerlavanet (approved):

PR #1312, opened 2024-03-18
fix: define `--delegate-limit` var

RuslanGlaznyov:
## What does this PR do? 
Fix error `Error: accepts 5 arg(s), received 6` when `spect_test_full.sh` run 

The $STAKE var is undefined, this PR defines var for `delegate-limit`, also `--delegate-limit`  can't be removed from the command, because cli rising error that this param is required


Yaroms:
Thank you for the contribution

Yaroms (approved):

PR #1343, opened 2024-03-29
Update README.md

jiOnederfull:
It seems that `lavad_auto_completion_install.sh` is no longer available and has been replaced by `lava_auto_completion_install.sh`.


Yaroms:
Thanks for opening the PR.

PR #1365, opened 2024-04-10
feat: spec add NAMADA

RuslanGlaznyov:
Tested with `./scripts/test_spec_full.sh` works as expected on my side
run script logs:
 
[run1.log](https://github.com/lavanet/lava/files/14948675/run1.log)

Forum: https://community.lavanet.xyz/t/add-spec-namada/119


cookbook/specs/spec_add_namada.json
```diff
@@ -0,0 +1,651 @@
...
+            "enabled": true,
+            "collection_data": {
+              "api_interface": "tendermintrpc",
+              "internal_path": "",
+              "type": "",
+              "add_on": ""
+            },
+            "apis": [
+              {
+                "name": "dump_consensus_state",
```
omerlavanet, line 33: if it's a tendermint chain, we are inheriting apis, rather than rewrite them for maintainability and readability
if you want to inherit only a part of the api collections you just overwrite the other imports to enabled:false by specifying the apiCollection with the same collection data
are there no other apis?

omerlavanet (changes requested):

cookbook/specs/spec_add_namada.json
RuslanGlaznyov: Thx for the response! I didn't know that we could inherit only one API. I will prepare fixes 

> Are there no other APIs?

Namada is based on commetbft, but the application layer doesn't use cosmosSDK, the app layer is written from scratch in Rust. Only RPC available

RuslanGlaznyov: Maybe for this purpose, it would be great to have inherited spec only for terndemint/commetBFT API? 
https://docs.tendermint.com/v0.34/rpc/

omerlavanet: Yes if chains start using tendermint separately to cosmos definitely
Need to create a base tendermint and cosmos to
Inherit from it

RuslanGlaznyov: Refoctered spec with inheritance but got the error in GitHub actions, but `test_spec_full.sh` works well on my side

omerlavanet: the sdk has a statistical unitest error we are aware of

cookbook/specs/spec_add_namada.json
```diff
@@ -0,0 +1,57 @@
...
+        "block_distance_for_finalized_data": 0,
+        "blocks_in_finalization_proof": 1,
+        "average_block_time": 6500,
+        "allowed_block_lag_for_qos_sync": 2,
+        "shares": 1,
+        "min_stake_provider": {
+          "denom": "ulava",
+          "amount": "47500000000"
+        },
+        "api_collections": [
```
omerlavanet, line 25: missing the chain-d verification, purning verification to make sure providers are serving the right chain with the right pruning configuration, inheriting from cosmos sdk also enables archive so it should be set as well

omerlavanet (changes requested):

cookbook/specs/spec_add_namada.json
RuslanGlaznyov: Any standard configuration for pruning the archive and unarchive node?

omerlavanet: 24h for the verification, 23.8h for the archive routing

RuslanGlaznyov: I need clarification about pruning  
```json
{
                "name": "pruning",
                "values": [
                  {
                    "latest_distance": <DEFAULT_DISTANCE>,
                  },
                    {
                    "extension": "archive",
                    "latest_distance": <ARCHIVE_DISTANCE>
                  }
                ]
              }
            ],
            "extensions": [
              {
                "name": "archive",
                "cu_multiplier": 5,
                "rule": {
                  "block":  <BLOCK_FOR_ARCHIVE>
                }
              }
            ]
```
For example block creation in chain each 5 seconds so: 
86400 = 24h in seconds 
17280 = 86400/5
17280blocks in 24h
so DEFAULT_DISTANCE is 17280. 

To calculate 23.8h 
85680s =  23.8h in seconds
17136 = 85680/5
17136blocks in 23.8 

So ARCHIVE_DISTANCE should be 17136? 


I can understand what we should put in BLOCK_FOR_ARCHIVE because in the osmosis example this 1. In Evmos 9990. 

I can't figure out the pattern, in osmosis the DEFAULT_DISTANCE is 17280. But I don't understand where we got this number if block creation takes 3.7 seconds in OSMOSIS regarding mintscan, 17280blocks * 3.7seconds_block = 17.76hours 🤷‍♂️

oren-lava: Hi @RuslanGlaznyov, sorry for the late response, let me clarify:
The structure of the verifications should be as follows (taken from Osmosis):
```
                        "verifications": [
                            {
                                "name": "chain-id",
                                "values": [
                                    {
                                        "expected_value": "osmosis-1"
                                    }
                                ]
                            },
                            {
                                "name": "pruning",
                                "values": [
                                    {
                                        "latest_distance": 17280
                                    },
                                    {
                                        "extension": "archive",
                                        "expected_value": "1"
                                    }
                                ]
                            }
                        ],
                        "extensions": [
                            {
                                "name": "archive",
                                "cu_multiplier": 5,
                                "rule": {
                                    "block": 17040
                                }
                            }
                        ]
```

Note that in your comment you put `latest_distance` under the `"extension": "archive"` line instead of `"expected_block`.

Let's go over the verifications one by one:
1. `chain-id`: the expected value should be the chain's ID (usually part of the response of a `status` query)
2. `pruning`: this verification refers to the minimal history of blocks a pruned node should have access to. The `latest_distance` field is that number of blocks. For all chains, we calculate the number of blocks in 24H. Specifically in Osmosis case, the calculation was: 86400 [sec in 24H] / 5 [sec per block] = 17280 [blocks]. Since you pointed out that Osmosis block creation is 3.7 seconds, we might need to fix that, but the logic stays the same. Also, specifically for Evmos, we didn't use this calculation for other reasons but it's only like this in Evmos.
3. archive extension under `pruning`: this field indicates the expected earliest value in an archived node. Usually it's 1 (the first block) but sometimes chains experience hard forks so archival nodes can only reach an earliest block which is larger than 1.
4. `rule` of archive extension: this expected value here (of `block`) is the threshold in which we determine whether a request is considered archived or not and should cost more. As a general rule, we consider requests for blocks that are older than the last 23H + 40M to be archive. In Osmosis case, the calculation was: (86400-1200) [sec in 23H+40M] / 5 [sec per block] = 17040 [blocks]

I hope this clears things out. If you have further questions you can comment here or reach me on Discord :)

RuslanGlaznyov: Thanks this is super helpful! I did updates thanks to your clarifications. 
- The block creation in NAMADA is around 10s regarding this message in [discord](https://discord.com/channels/833618405537218590/1222159147353772122/1222174575517892619)
- The expected value is `237907`, because NAMADA SE did hard. See [discord announcement](https://discord.com/channels/833618405537218590/1196488611613376574/1222554358965600327)

But, with my updates, I got an error from ./test_spec_full.sh. 

Here is shorted lines: 
```log
finished waiting at block "17"
waiting for next block "17"
-Error: rpc error: code = Unknown desc = rpc error: code = Unknown desc = recovered: runtime error: invalid memory address or nil pointer dereference
stack:
goroutine 320 [running]:
runtime/debug.Stack()
        runtime/debug/stack.go:24 +0x65
github.com/cosmos/cosmos-sdk/baseapp.newDefaultRecoveryMiddleware.func1({0x2a788a0, 0x57b2550})
        github.com/cosmos/cosmos-sdk@v0.47.10/baseapp/recovery.go:71 +0x27
github.com/cosmos/cosmos-sdk/baseapp.newRecoveryMiddleware.func1({0x2a788a0?, 0x57b2550?})
        github.com/cosmos/cosmos-sdk@v0.47.10/baseapp/recovery.go:39 +0x30
github.com/cosmos/cosmos-sdk/baseapp.processRecovery({0x2a788a0, 0x57b2550}, 0xc00513e580?)
        github.com/cosmos/cosmos-sdk@v0.47.10/baseapp/recovery.go:28 +0x37
github.com/cosmos/cosmos-sdk/baseapp.processRecovery({0x2a788a0, 0x57b2550}, 0x3eb4260?)
        github.com/cosmos/cosmos-sdk@v0.47.10/baseapp/recovery.go:33 +0x5e
github.com/cosmos/cosmos-sdk/baseapp.(*BaseApp).runTx.func1()
        github.com/cosmos/cosmos-sdk@v0.47.10/baseapp/baseapp.go:646 +0xf0
panic({0x2a788a0, 0x57b2550})
        runtime/panic.go:890 +0x263
```

Full log: 
[run_error.log](https://github.com/lavanet/lava/files/15120165/run_error.log)

RuslanGlaznyov: @oren-lava 
Fixed error.. Was type in `chain-id` verification. Now /test_spec_full.sh works as expected

omerlavanet (approved):

omerlavanet:
@RuslanGlaznyov very well done, beautiful use of inheritance

PR #1372, opened 2024-04-12
feat: spec-manta

BabyScope:
Hi i tested out with the script and it is working i added the test result as screenshot
Schermafbeelding 2024-04-12 205729
Schermafbeelding 2024-04-12 210119
Schermafbeelding 2024-04-12 210138

### DIFF
--- a/cookbook/specs/spec_add_manta.json
+++ b/cookbook/specs/spec_add_manta.json
@@ -45,7 +45,7 @@
                   "name": "pruning",
                   "values": [
                     {
-                      "latest_distance": 0
+                      "latest_distance": 8640
                     },
                     {
                       "extension": "archive",
@@ -59,7 +59,7 @@
                   "name": "archive",
                   "cu_multiplier": 5,
                   "rule": {
-                    "block": 0
+                    "block": 8550
                   }
                 }
               ]
--- a/cookbook/specs/spec_add_manta.json
+++ b/cookbook/specs/spec_add_manta.json
@@ -1,0 +1,115 @@
+{
+    "proposal": {
+      "title": "Add Specs: Manta",
+      "description": "Adding new specification support for relaying data on Manta",
+      "specs": [
+        {
+          "index": "MANTA",
+          "name": "manta mainnet",
+          "enabled": true,
+          "imports": ["ETH1"],
+          "reliability_threshold": 268435455,
+          "data_reliability_enabled": true,
+          "block_distance_for_finalized_data": 1,
+          "blocks_in_finalization_proof": 3,
+          "average_block_time": 10000,
+          "allowed_block_lag_for_qos_sync": 5,
+          "shares": 1,
+          "min_stake_provider": {
+            "denom": "ulava",
+            "amount": "47500000000"
+          },
+          "api_collections": [
+            {
+              "enabled": true,
+              "collection_data": {
+                "api_interface": "jsonrpc",
+                "internal_path": "",
+                "type": "POST",
+                "add_on": ""
+              },
+              "apis": [],
+              "headers": [],
+              "inheritance_apis": [],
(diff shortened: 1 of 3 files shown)


BabyScope:
Union also working 
Schermafbeelding 2024-04-12 215648

cookbook/specs/spec_add_manta.json
```diff
@@ -0,0 +1,115 @@
...
+                    }
+                  ]
+                }
+              ],
+              "extensions": [
+                {
+                  "name": "archive",
+                  "cu_multiplier": 5,
+                  "rule": {
+                    "block": 0
```
omerlavanet, line 62: needs to be set otherwise archive won't work, rule of thumb is a 24h - 15m

cookbook/specs/spec_add_manta.json
```diff
@@ -0,0 +1,115 @@
...
+                    {
+                      "expected_value": "0xa9"
+                    }
+                  ]
+                },
+                {
+                  "name": "pruning",
+                  "values": [
+                    {
+                      "latest_distance": 0
```
omerlavanet, line 48: 0 here means any pruning value will work, this is not working well with the archive value, this works together with the extension archive block rule, rule of thumb is 24h of blocks

cookbook/specs/spec_add_union.json
```diff
@@ -0,0 +1,99 @@
...
+            "api_collections": [
+              {
+                "enabled": true,
+                "collection_data": {
+                  "api_interface": "rest",
+                  "internal_path": "",
+                  "type": "GET",
+                  "add_on": ""
+                },
+                "apis": [],
```
omerlavanet, line 33: inheriting rest without adding api's mean that only the basic cosmos rest calls will work, and any additional added modules not in the basic sdk will be missing here, is that the intention?

cookbook/specs/spec_add_union.json
```diff
@@ -0,0 +1,99 @@
...
+              },
+              {
+                "enabled": true,
+                "collection_data": {
+                  "api_interface": "grpc",
+                  "internal_path": "",
+                  "type": "",
+                  "add_on": ""
+                },
+                "apis": [],
```
omerlavanet, line 56: same as rest, when you add rest apis grpc should have new ones too

omerlavanet (changes requested):

cookbook/specs/spec_add_manta.json
oren-lava: just to clarify, @omerlavanet means that the value that you need to put is:
<manta_average_block_time_in_seconds> * <number_of_seconds_in_23H_and_45M> = number of blocks in 23H and 45M

cookbook/specs/spec_add_manta.json
oren-lava: just to clarify, @omerlavanet means that the value that you need to put is:
<manta_average_block_time_in_seconds> * <number_of_seconds_in_24H> = number of blocks in 24H

cookbook/specs/spec_add_manta.json
BabyScope: I added 8550 
10 seconds average a block meaning 8550 block on 24H-15MIN

cookbook/specs/spec_add_manta.json
BabyScope: Also changed this to 8640 block

cookbook/specs/spec_add_union.json
BabyScope: Yes i checked them all other then the cosmos chain the are importing ibc that is already being imported in COSMOS so is it good to only import COSMOSSDK or also need to add IBC?

PR #1376, opened 2024-04-14
fix: PRT-fix context timeout when sending relay retries

ranlavanet:


nimrod-teich (approved):

shleikes (approved):

protocol/rpcconsumer/rpcconsumer_server.go
```diff
@@ -388,13 +393,13 @@ func (rpccs *RPCConsumerServer) ProcessRelaySend(ctx context.Context, directiveH
...
 			}
-			err := rpccs.sendRelayToProvider(ctx, chainMessage, relayRequestData, dappID, consumerIp, relayProcessor)
+			err := rpccs.sendRelayToProvider(processingCtx, chainMessage, relayRequestData, dappID, consumerIp, relayProcessor)
 			go validateReturnCondition(err)
 			go readResultsFromProcessor()
 		case <-startNewBatchTicker.C:
 			// only trigger another batch for non BestResult relays
 			if relayProcessor.selection != BestResult {
-				err := rpccs.sendRelayToProvider(ctx, chainMessage, relayRequestData, dappID, consumerIp, relayProcessor)
+				err := rpccs.sendRelayToProvider(processingCtx, chainMessage, relayRequestData, dappID, consumerIp, relayProcessor)
```
omerlavanet, line 402: let's add a condition that if processingCtx time left is less than relay timeout we do relayTimeout? so we don;t have an absurd condition we have 1ms left and we open a session then timeout

omerlavanet (changes requested):

protocol/rpcconsumer/rpcconsumer_server.go
ranlavanet: done

Yaroms (approved):

omerlavanet (approved):

PR #1391, opened 2024-04-26
chore: Use errors.New instead of fmt.Errorf with no parameters

ChengenH:


omerlavanet:
does not meet criteria for a contribution

PR #1392, opened 2024-04-26
feat: Add Manta Pacific Spec

datanexus-vincent:
The full test worked just fine and from what I can tell Manta Pacific doesn't have any custom RPC methods on top of the Ethereum EVM spec.

Testing command:
```
./scripts/test_spec_full.sh cookbook/specs/spec_add_manta_pacific.json jsonrpc https://pacific-rpc.manta.network/http jsonrpc https://pacific-rpc.testnet.manta.network/http
```

FYI, it looks like there hasn't been a new block on the testnet in around a month, so there won't be any new block log lines.

### DIFF
--- a/cookbook/specs/spec_add_manta_pacific.json
+++ b/cookbook/specs/spec_add_manta_pacific.json
@@ -1,0 +1,118 @@
+{
+    "proposal": {
+        "title": "Add Specs: Manta Pacific",
+        "description": "Adding new specification support for relaying Manta Pacific data on Lava",
+        "specs": [
+            {
+                "index": "MANTAPACIFIC",
+                "name": "manta pacific mainnet",
+                "enabled": true,
+                "imports": [
+                    "ETH1"
+                ],
+                "reliability_threshold": 268435455,
+                "data_reliability_enabled": true,
+                "block_distance_for_finalized_data": 1,
+                "blocks_in_finalization_proof": 3,
+                "average_block_time": 10000,
+                "allowed_block_lag_for_qos_sync": 2,
+                "shares": 1,
+                "min_stake_provider": {
+                    "denom": "ulava",
+                    "amount": "47500000000"
+                },
+                "api_collections": [
+                    {
+                        "enabled": true,
+                        "collection_data": {
+                            "api_interface": "jsonrpc",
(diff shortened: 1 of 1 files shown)


cookbook/specs/spec_add_manta_pacific.json
```diff
@@ -0,0 +1,118 @@
...
+                                    {
+                                        "expected_value": "0xa9"
+                                    }
+                                ]
+                            },
+                            {
+                                "name": "pruning",
+                                "values": [
+                                    {
+                                        "latest_distance": 1990000
```
omerlavanet, line 50: will require nodes to have 1,990,000 blocks to operate, quite a lot, and every request of bigger than 17040 will go to archive, rule of thumb is 24h of blocks, with 10s per block it's 24*3600/10 blocks

omerlavanet (changes requested):

BabyScope:
Hi Nexus,

For the pruning and archive distance you need to keep in mind.
Pruning blocks = 24 hours
Archive blocks = 24 hours - 15 minutes 

So with this in mind;
{
                  "name": "pruning",
                  "values": [
                    {
                      "latest_distance": 8640
                    },
                    {
                      "extension": "archive",
                      "expected_value": "0x0"
                    }
                  ]
                }
              ],
              "extensions": [
                {
                  "name": "archive",
                  "cu_multiplier": 5,
                  "rule": {
                    "block": 8550
                  }
                }

Results should look like this.

@omerlavanet  i would say if he resolve this you can merge this i will add side then don't need two seperate pulls from same chain

cookbook/specs/spec_add_manta_pacific.json
datanexus-vincent: Ah, sorry about that. I misunderstood that value. I thought it was the block height threshold for archive vs full, but I understand now. I'll get this changed and pushed!

datanexus-vincent:
Sorry about the confusion in regards to those values. I understand them now so it shouldn't be an issue in the future.

oren-lava (approved):

Yaroms:
looking good, thanks @datanexus-vincent

PR #1406, opened 2024-05-03
feat: union spec add

BabyScope:
Union spec with pruning

### DIFF
--- a/cookbook/specs/spec_add_union.json
+++ b/cookbook/specs/spec_add_union.json
@@ -1,0 +1,209 @@
+{
+    "proposal": {
+        "title": "Add Specs: Union",
+        "description": "Adding new specification support for relaying union data on Lava",
+        "specs": [
+            {
+                "index": "UNIONT",
+                "name": "union testnet",
+                "enabled": true,
+                "imports": [
+                    "COSMOSSDK"
+                ],
+                "reliability_threshold": 268435455,
+                "data_reliability_enabled": true,
+                "block_distance_for_finalized_data": 0,
+                "blocks_in_finalization_proof": 1,
+                "average_block_time": 6180,
+                "allowed_block_lag_for_qos_sync": 2,
+                "shares": 1,
+                "min_stake_provider": {
+                    "denom": "ulava",
+                    "amount": "47500000000"
+                },
+                "api_collections": [
+                    {
+                        "enabled": true,
+                        "collection_data": {
+                            "api_interface": "rest",
+                            "internal_path": "",
+                            "type": "GET",
+                            "add_on": ""
+                        },
+                        "apis": [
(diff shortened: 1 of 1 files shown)


cookbook/specs/spec_add_union.json
```diff
@@ -0,0 +1,209 @@
...
+                                    {
+                                        "expected_value": "union-testnet-8"
+                                    }
+                                ]
+                            },
+                            {
+                                "name": "pruning",
+                                "values": [
+                                    {
+                                        "latest_distance": 13980
```
oren-lava, line 105: I saw in Union explorer that the average block time is about 6 seconds. so the latest distance should 14400 (seconds_in_24h / block_time = 86400 / 6 = 14400)
Also update the `archive` rule as well

cookbook/specs/spec_add_union.json
```diff
@@ -0,0 +1,209 @@
...
+    "proposal": {
+        "title": "Add Specs: Union",
+        "description": "Adding new specification support for relaying union data on Lava",
+        "specs": [
+            {
+                "index": "UNIONT",
+                "name": "union testnet",
+                "enabled": true,
+                "imports": [
+                    "COSMOSSDK"
```
oren-lava, line 11: also import IBC

oren-lava (changes requested):

cookbook/specs/spec_add_union.json
BabyScope: I didn't import IBC because COSMOSSDK already imports it.

cookbook/specs/spec_add_union.json
BabyScope: Will do i looked it up on 3 different occasions it was from 6,7 seconds till 6,1 seconds whichone should I use?

oren-lava: use 6.1 and round it to 6 seconds

cookbook/specs/spec_add_union.json
oren-lava: you're right, my bad

cookbook/specs/spec_add_union.json
```diff
@@ -0,0 +1,209 @@
...
+                    },
+                    {
+                        "enabled": true,
+                        "collection_data": {
+                            "api_interface": "grpc",
+                            "internal_path": "",
+                            "type": "",
+                            "add_on": ""
+                        },
+                        "apis": [],
```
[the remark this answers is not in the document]
BabyScope: Oren, the grpc function is already mentioned in cosmossdk that's why i didn't include it

Yaroms (approved):

PR #1408, opened 2024-05-03
feat: side spec

BabyScope:
Side testnet spec sheet

[Side - Portal.txt](https://github.com/lavanet/lava/files/15204868/Side.-.Portal.txt)
[Side - Provider.txt](https://github.com/lavanet/lava/files/15204869/Side.-.Provider.txt)


cookbook/specs/spec_add_side.json
```diff
@@ -0,0 +1,489 @@
...
+                "apis": [],
+                "headers": [],
+                "inheritance_apis": [],
+                "parse_directives": [],
+                "verifications": [
+                    {
+                        "name": "chain-id",
+                        "values": [
+                            {
+                                "expected_value": "S2-testnet-1"
```
oren-lava, line 457: I saw that side chain id is "side-testnet-3" in their explorer.
Can you share where you saw this ID? if it's an old one, please update it to "side-testnet-3" (in all api interfaces)

cookbook/specs/spec_add_side.json
```diff
@@ -0,0 +1,489 @@
...
+            },
+              {
+                "enabled": true,
+                "collection_data": {
+                  "api_interface": "grpc",
+                  "internal_path": "",
+                  "type": "",
+                  "add_on": ""
+                },
+                "apis": [
```
oren-lava, line 245: using grpcurl, the following API is missing:
grpc.reflection.v1alpha.ServerReflection/ServerReflectionInfo

cookbook/specs/spec_add_side.json
```diff
@@ -0,0 +1,489 @@
...
+                    "category": {
+                        "deterministic": true,
+                        "local": false,
+                        "subscription": false,
+                        "stateful": 0
+                    },
+                    "extra_compute_units": 0
+                },
+                {
+                    "name": "side.interchainquery.v1.QueryService/PendingQueries",
```
oren-lava, line 373: using grpcurl and KJnodes endpoint, I didn't see this API supported. where did you get it from?
If it's old and deprecated, remove it here and in REST

oren-lava (changes requested):

cookbook/specs/spec_add_side.json
BabyScope: Side tesnet 3 is halted they are going to start S2-testnet-1 that's why already named it that way.

cookbook/specs/spec_add_side.json
BabyScope: I got it from there GitHub proto file so which one should I follow the ones from gcurl or the proto files?

oren-lava: i tried 3 different endpoints and none support this. I think you should remove it for now

cookbook/specs/spec_add_side.json
oren-lava: use the current chain id. If we deploy it tomorrow and providers would want to start giving service, their process will fail on this verification since S2-testnet-1 doesn't exist yet.
When side will start S2-testnet-1, we'll update the spec

BabyScope: the S2-testnet-1 started 03.05.2024 so that is the new one

cookbook/specs/spec_add_side.json
BabyScope: This changed with the S2-testnet-1

cookbook/specs/spec_add_side.json
oren-lava: ok, keep it as is

cookbook/specs/spec_add_side.json
oren-lava: not relevant (found in COSMOSSDK)

BabyScope:
Also changed the specs had statefull 1 no put 0 and deleted hanging API because the API dont need blocks

oren-lava (approved):

PR #1417, opened 2024-05-06
feat: dymension spec add

BabyScope:
[DYMENSION LOG.txt](https://github.com/lavanet/lava/files/15220389/DYMENSION.LOG.txt)
[DYMENSION PORTAL.txt](https://github.com/lavanet/lava/files/15220390/DYMENSION.PORTAL.txt)

I added ethermint api in Dymension cause evmos not using. Evmos is using ethermint in there grpc so the ones that are not in Evmos i added also in the grpc.

### DIFF
--- a/cookbook/projects/policy_all_chains_with_extension.yml
+++ b/cookbook/projects/policy_all_chains_with_extension.yml
@@ -56,4 +56,24 @@ Policy:
           extensions:
             - "archive"
           mixed: true
+    - chain_id: COSMOSHUB
+      requirements:
+        - collection:
+            api_interface: "rest"
+            type: "GET"
+          extensions:
+            - "archive"
+          mixed: true
+        - collection:
+            api_interface: "grpc"
+            type: ""
+          extensions:
+            - "archive"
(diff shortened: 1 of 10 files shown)


Yaroms:
@omerlavanet i think the best way is to create a spec specific for ethermint

omerlavanet:
yes would make a lot of sense, create ethermint base spec and inherit form it, @BabyScope want to give it a go?

BabyScope:
I will try it out

BabyScope:
Changed the spec of dymension added Ethermint to imports. Also changed the test script ethermint is added to the legacy proposal.

Yaroms:
@BabyScope for the tests to pass we need to set the ethermint spec as base spec in the following test:
https://github.com/lavanet/lava/blob/4daa8f49185d71a46efa28bbbf319bdecef1596c/x/spec/keeper/spec_test.go#L811

BabyScope:
@Yaroms i added it to the spec_test but he is still failling? do i something wrong here?

PR #1418, opened 2024-05-06
feat: Add Injective Spec

datanexus-vincent:
The full test works just fine with the right public RPC providers.

```
./scripts/test_spec_full.sh cookbook/specs/spec_add_injective.json tendermintrpc https://injective-rpc.publicnode.com:443 grpc sentry.chain.grpc.injective.network:443 rest https://sentry.lcd.injective.network:443 tendermintrpc https://testnet.sentry.tm.injective.network:443 grpc testnet.sentry.chain.grpc.injective.network:443 rest https://testnet.sentry.lcd.injective.network:443
```


cookbook/specs/spec_add_injective.json
```diff
@@ -0,0 +1,4530 @@
...
+                                    {
+                                        "expected_value": "injective-1"
+                                    }
+                                ]
+                            },
+                            {
+                                "name": "pruning",
+                                "values": [
+                                    {
+                                        "latest_distance": 144000
```
oren-lava, line 4416: how did you calculate this number?
I saw that Injective block time is 1sec, so it should be 86400 (number of seconds in 24H). Also, you should fix the archive rule to be 23H and 45M

oren-lava (changes requested):

cookbook/specs/spec_add_injective.json
datanexus-vincent: I saw on the official explorer that the block time was 600ms, but now it seems it is reporting 800ms, and I'm also seeing some places say 900ms and 1s. Seems like it might be somewhat variable. If that's the case, would it better to be on the upper end of the block time range? I'm guessing so considering it's a verification.

oren-lava: yeah, let's use 1sec

Yaroms (approved):

PR #1421, opened 2024-05-07
feat: CNS-952: provider pairing chance query & provider query

oren-lava:
In this PR, I implemented a new query for the pairing module that gets a provider address and chain ID and calculates the chance of said provider to be picked as part of the pairing.

Currently, the chance is influenced by stake and geolocation (the cluster was also added by it's not influencing the pairing process in general ATM).

Note that it's only an estimation. the query calculates the chance of being picked on the first pairing slot. In the full pairing process, the chance of getting picked also changes for different slot. Moreover, this query doesn't consider addons, extensions, or selected providers modes that can heavily influence the pairing chance. 

Also, I implemented the provider query which gets a Lava address and chain ID and returns the corresponding stake entry, if it exists


x/plans/types/geolocation.go
```diff
@@ -11,8 +11,9 @@ import (
 
 // for convenience (calculate once only)
 var (
-	allGeoEnumRegionsList []Geolocation
-	allGeoEnumRegions     int32
+	allGeoEnumRegionsList      []Geolocation
+	allGeoEnumRegionsListInt32 []int32
```
Yaroms, line 15: remove this

x/subscription/types/cluster.go
```diff
@@ -23,6 +23,8 @@ import "strconv"
 
 const FREE_PLAN = "free" // gets its own const because it's treated differently
 
+var SUB_USAGES = []int{0, 6, 12} // sub usages that are treated differemtly when constructing a cluster key
```
Yaroms, line 26: make sure to use that also in the GetSubUsageCriterion

oren-lava: done. see https://github.com/lavanet/lava/commit/a1547529b3cefd06a28a6dd754a9f3970c40496b

Yaroms (approved):

PR #1440, opened 2024-05-14
feat: PRT-add-grpc-compression-for-provider-consumer-communication

ranlavanet:


protocol/lavasession/consumer_session_manager.go
```diff
@@ -961,13 +961,13 @@ func (csm *ConsumerSessionManager) OnSessionDone(
 	// calculate QoS
 	consumerSession.CalculateQoS(currentLatency, expectedLatency, expectedBH-latestServicedBlock, numOfProviders, int64(providersCount))
 	go csm.providerOptimizer.AppendRelayData(consumerSession.Parent.PublicLavaAddress, currentLatency, isHangingApi, specComputeUnits, uint64(latestServicedBlock))
-	csm.updateMetricsManager(consumerSession)
+	csm.updateMetricsManager(consumerSession, currentLatency, !isHangingApi) // apply latency only for non hanging apis
```
omerlavanet, line 964: nice!

protocol/metrics/metrics_consumer_manager.go
```diff
@@ -5,13 +5,26 @@ import (
...
+type LatencyTracker struct {
+	AverageLatency time.Duration
+	TotalRequests  int
+}
+
+func (lt *LatencyTracker) AddLatency(latency time.Duration) {
+	lt.TotalRequests++
+	weight := 1.0 / float64(lt.TotalRequests)
+	// Calculate the weighted average of the current average latency and the new latency
+	lt.AverageLatency = time.Duration(float64(lt.AverageLatency)*(1-weight) + float64(latency.Milliseconds())*weight)
```
omerlavanet, line 25: if you turn it into milliseconds then cast back into duration it will divide the time and all the usages of averageLatenncy as a time.Duration will be incorrect, so don't use .Milliseconds for the calculation

omerlavanet (changes requested):

omerlavanet (approved):

PR #1441, opened 2024-05-16
feat: ethereaum beacon chain 

BabyScope:
Hi Omer, Here you have it somehow the parsing is not working


cookbook/specs/spec_add_beacon.json
```diff
@@ -0,0 +1,1366 @@
...
+                                "category": {
+                                    "deterministic": true,
+                                    "local": false,
+                                    "subscription": false,
+                                    "stateful": 0
+                                },
+                                "extra_compute_units": 0
+                            },
+                            {
+                                "name": "/eth/v1/beacon/pool/attestations",
```
Yaroms, line 446: this is a tx, when an api is tx stateful should be 1 and deterministic should be false

cookbook/specs/spec_add_beacon.json
```diff
@@ -0,0 +1,1366 @@
...
+                                "block_parsing": {
+                                    "parser_arg": [
+                                        "latest"
+                                    ],
+                                    "parser_func": "DEFAULT"
+                                },
+                                "compute_units": 10,
+                                "enabled": true,
+                                "category": {
+                                    "deterministic": true,
```
Yaroms, line 474: this is a tx, when an api is tx stateful should be 1 and deterministic should be false

cookbook/specs/spec_add_beacon.json
```diff
@@ -0,0 +1,1366 @@
...
+                                "name": "/eth/v1/beacon/pool/proposer_slashings",
+                                "block_parsing": {
+                                    "parser_arg": [
+                                        "latest"
+                                    ],
+                                    "parser_func": "DEFAULT"
+                                },
+                                "compute_units": 10,
+                                "enabled": true,
+                                "category": {
```
Yaroms, line 491: this is a tx, when an api is tx stateful should be 1 and deterministic should be false

cookbook/specs/spec_add_beacon.json
```diff
@@ -0,0 +1,1366 @@
...
+                                    "deterministic": true,
+                                    "local": false,
+                                    "subscription": false,
+                                    "stateful": 0
+                                },
+                                "extra_compute_units": 0
+                            },
+                            {
+                                "name": "/eth/v1/beacon/pool/voluntary_exits",
```
Yaroms, line 500: this is a tx, when an api is tx stateful should be 1 and deterministic should be false

cookbook/specs/spec_add_beacon.json
```diff
@@ -0,0 +1,1366 @@
...
+                                "category": {
+                                    "deterministic": true,
+                                    "local": false,
+                                    "subscription": false,
+                                    "stateful": 0
+                                },
+                                "extra_compute_units": 0
+                            },
+                            {
+                                "name": "/eth/v1/node/peers",
```
Yaroms, line 734: all queries that are node specific should be deterministic false.
if the same query from different nodes can yield different results the query is not deterministic

Yaroms:
@BabyScope i left some comments, aside from that looking good

BabyScope:
Hi @Yaroms ,

I made changes to it but have a question.

This one is true right?
because you search on peer id so it need to be the same all time
"name": "/eth/v1/node/peers/{peer_id}",

Yaroms:
> Hi @Yaroms ,
> 
> I made changes to it but have a question.
> 
> This one is true right? because you search on peer id so it need to be the same all time "name": "/eth/v1/node/peers/{peer_id}",

i would say that a node is not connected to all peers at all times, so this query is specific for each node.
here is a quote from the docs:
"/eth/v1/node/peers
Get node network peers.

Brings up information about the node's network neighbors. This retrieves all peers by default. AND conditions are used to combine multiple query parameters."

Yaroms (approved):

PR #1445, opened 2024-05-21
fix: PRT-1205: Support null ID in jsonrpc

shleikes:


ecosystem/cache/cache_test.go
```diff
@@ -466,9 +475,15 @@ func TestCacheSetGetJsonRPCWithID(t *testing.T) {
 			id := rand.Int63()
 
 			formatIDInJson := func(idNum int64) []byte {
+				if tt.nullId {
```
omerlavanet, line 478: you need to check two more scenarios:
1. you set id (not null) and then do a get with id:null - needs to work
2. you set id (null) and then do a get id(not null) - needs to work
in your function you both set and get the same id that is null

omerlavanet (changes requested):

omerlavanet (approved):

PR #1452, opened 2024-05-30
feat: IPRPC over IBC: Part 3 - CNS-965: Pending IPRPC IBC fund

oren-lava:
In this PR, I implemented a new scaffolded map which holds PendingIprpcIbcFund objects with increasing int index.

This map allows us to keep data of IPRPC over IBC fund requests for future funding. To clarify, all IPRPC over IBC fund requests are automatically turning to PendingIprpcIbcFund since they do not cover the minimum cost of funding the IPRPC pool. The funds received in the ibc-transfer packet are sent to a constant module address derived from the string "iprpc".

The PendingIprpcIbcFund has an expiration timestamp determined by a new parameter of the rewards module : IbcIprpcExpiration (default = 3 months). Expired objects are removed every BeginBlock and their funds are transferred to the community pool.

Covering the PendingIprpcIbcFund will be implemented in the next parts.


proto/lavanet/lava/rewards/params.proto
```diff
@@ -40,4 +40,10 @@ message Params {
     (gogoproto.customtype) = "github.com/cosmos/cosmos-sdk/types.Dec",
     (gogoproto.nullable)   = false
   ];
+
+  google.protobuf.Duration ibc_iprpc_expiration = 7 [
```
Yaroms, line 44: when adding a new param you must set it in a migrator or we will get a panic

x/rewards/keeper/ibc_iprpc.go
```diff
@@ -99,7 +101,191 @@ func printInvalidMemoWarning(iprpcData map[string]interface{}, description strin
...
+				utils.LogAttr("spec", spec),
+				utils.LogAttr("duration", duration),
+				utils.LogAttr("funds", fund),
+				utils.LogAttr("leftovers", leftovers),
+			)
+		}
+	}
+
+	// get index for the new object
+	latestPendingIbcIprpcFund := k.GetLatestPendingIbcIprpcFund(ctx)
```
Yaroms, line 159: is reusing indicis a problem?

x/rewards/ibc_middleware.go
```diff
@@ -112,25 +113,50 @@ func (im IBCMiddleware) OnRecvPacket(ctx sdk.Context, packet channeltypes.Packet
...
+		utils.LavaFormatError("rewards module IBC middleware processing failed", fmt.Errorf("ibcFundCoins has more than one type of coins"),
+			utils.LogAttr("data", data),
+		)
+		return channeltypes.NewErrorAcknowledgement(err)
+	}
+
+	// set pending IPRPC over IBC requests on-chain
+	piif, err := im.keeper.NewPendingIbcIprpcFund(ctx, memo.Creator, memo.Spec, memo.Duration, ibcFundCoins[0])
 	if err != nil {
 		return channeltypes.NewErrorAcknowledgement(err)
```
Yaroms, line 147: we need to make sure that if we fail here the ibc tokens dont stay on chain

oren-lava: done. see https://github.com/lavanet/lava/commit/53009f19b9398625b095096d63e07c0bf80f0b81

oren-lava: Ignore the previous commit, see fix here: https://github.com/lavanet/lava/commit/752017792d92ca6a28573f198d96585654273395
What I did is: 
1. reorder the code so we cannot fail after the sending the IBC tokens. Note that I deleted the len(ibcIprpcFunds) check because it's redundant (the packet's data struct only supports one denom, so it'll always succeed). Also I deleted the balance check because I added an amount check before.
2. created a pending IPRPC pool that holds the pending IPRPC requests funds

proto/lavanet/lava/rewards/params.proto
oren-lava: done. see https://github.com/lavanet/lava/commit/12b0582b29f896f95c15c770300717651fddb92c

migrator was tested both locally and with state sync

x/rewards/ibc_middleware.go
oren-lava: also look at this: https://github.com/lavanet/lava/commit/5b040c72f64ee9f2bb6696e70886601bb0793eb9
this handles deletion of expired pending iprpc requests

PR #1453, opened 2024-05-30
feat: PRT - adding useful headers 

ranlavanet:


ranlavanet:
add another metadata to the user that did the provider send to his extension or not

protocol/chainlib/chainproxy/connector_test.go
```diff
@@ -2,26 +2,29 @@ package chainproxy
...
 	"google.golang.org/grpc/metadata"
 )
 
 const (
-	listenerAddress    = "localhost:1234"
-	port               = "1234"
-	listenerAddressTcp = "http://localhost:1234"
-	numberOfClients    = 5
+	listenerAddress     = "localhost:1234"
+	listenerAddressGrpc = "localhost:1235"
```
omerlavanet, line 25: let's use a generic port :0 like we do in our more advanced tests, it's easy to do a listen to port 0 then get the port and save it in a global variable, continue form there

omerlavanet (approved):

omerlavanet (approved):

PR #1464, opened 2024-06-02
feat: docker compose support

amitza:
- [x] docker compose for state sync operation


omerlavanet (changes requested):
please add a readme explaining how to modify the compose to adapt to a different version/chain(testnet-mainnet etc)

docker/docker-compose.state-sync.yml
```diff
@@ -0,0 +1,76 @@
+
+services:
+
+  lava-node-init:
+    # Update with the most recent supported version of lavad
+    image: ghcr.io/lavanet/lava/lavad:${LAVAD_VERSION:-v2.0.0}
+    environment:
+      - CHAIN_ID=lava-testnet-2
```
omerlavanet: let's allow the user to set these variables too via an export

omerlavanet (changes requested):

omerlavanet (approved):

PR #1666, opened 2024-09-01
feat: PRT - Support StarkNet pathfinder

shleikes:


protocol/chainlib/jsonRPC.go
```diff
@@ -517,57 +525,20 @@ func NewJrpcChainProxy(ctx context.Context, nConns uint, rpcProviderEndpoint lav
...
+	if err != nil {
+		return err
 	}
-	return nil
-}
 
-func (cp *JrpcChainProxy) start(ctx context.Context, nConns uint, nodeUrl common.NodeUrl, internalPaths map[string]struct{}) error {
-	if len(internalPaths) == 0 {
-		internalPaths = map[string]struct{}{"": {}} // add default path
+	cp.conn[nodeUrl.InternalPath] = conn
```
ranlavanet, line 537: weren't we supposed to deprecate conn map and have 1 conn for each proxy now that we create chain router per internal path

protocol/chainlib/jsonRPC.go
```diff
@@ -61,7 +61,12 @@ func (apip *JsonRPCChainParser) getSupportedApi(name, connectionType string, int
 
 func (apip *JsonRPCChainParser) CraftMessage(parsing *spectypes.ParseDirective, connectionType string, craftData *CraftData, metadata []pairingtypes.Metadata) (ChainMessageForSend, error) {
 	if craftData != nil {
-		chainMessage, err := apip.ParseMsg("", craftData.Data, craftData.ConnectionType, metadata, extensionslib.ExtensionInfo{LatestBlock: 0})
+		path := craftData.Path
```
ranlavanet, line 64: not sure why path is here can you explain why we moved from "" to path / internal path ? 
(internal path part makes sense but path doesnt )

protocol/lavasession/router_key.go
```diff
@@ -4,34 +4,78 @@ import (
...
+
+func (rk *RouterKey) ApplyInternalPath(internalPath string) {
+	rk.internalPath = internalPath
+}
+
+func (rk RouterKey) HasExtension(extension string) bool {
+	return lavaslices.Contains(rk.uniqueExtensions, extension)
+}
+
+func (rk RouterKey) String() string {
```
ranlavanet, line 66: This one doesn't change, I dont want us to calculate it every time string is called. it would be better to just store the key string in the struct once and return it. 
if rk.key != "" {
    return rk.key 
}

else calculate it and store.

ranlavanet (changes requested):
Very nice overall. minor changes please

protocol/lavasession/router_key.go
shleikes: It is dangerous because we might add some logic that will actually change it in the run, and then we will get unexpected behavior. There is currently only one place in the code that creates the router key and call its String function more than once. I fixed it there in this commit: b3982570d647634e91812c6cb049ca497f0d30de

protocol/chainlib/jsonRPC.go
shleikes: You are right sir, the fix is here: eba62e6063f38f5f9cec6d259435d295a25342a3

protocol/chainlib/jsonRPC.go
shleikes: That's because when we send node messages on init to each node in the node_urls, we want to use the right internal path. 
You can look the use here:

https://github.com/lavanet/lava/blob/eba62e6063f38f5f9cec6d259435d295a25342a3/protocol/chainlib/chain_fetcher.go#L170

ranlavanet (dismissed):

ranlavanet (approved):

protocol/chainlib/chain_router_test.go
omerlavanet, line 1: let's verify we have a test for partial support:
one chain proxy is an addon, one is extension and addon,
let's make sure:
1. regular calls go to the non extension endpoint
2. extension only calls go to the extension endpoint
3. the order of defining them doesnt matter

protocol/chainlib/chain_router.go
```diff
@@ -114,10 +117,51 @@ func (cri *chainRouterImpl) BatchNodeUrlsByServices(rpcProviderEndpoint lavasess
...
 			}
-			routerKey = routerKey.ApplyMethodsRoute(existing)
+			routerKey.ApplyMethodsRoute(existing)
+		}
+		routerKey.ApplyInternalPath(nodeUrl.InternalPath)
+		isWs, err := IsUrlWebSocket(nodeUrl.Url)
+		// Some parsing may fail because of gRPC
+		if err == nil && isWs {
+			// save the first ws router key and nodeUrl for later use
+			firstWsRouterKey = routerKey
```
omerlavanet, line 127: you are overwriting this every time, not sure if this is intended, if so maybe call this lastWsRouterKey?

protocol/chainlib/chain_router.go
```diff
@@ -234,8 +263,8 @@ func newChainRouter(ctx context.Context, nConns uint, rpcProviderEndpoint lavase
 	_, apiCollection, hasSubscriptionInSpec := chainParser.GetParsingByTag(spectypes.FUNCTION_TAG_SUBSCRIBE)
 	// validating we have websocket support for subscription supported specs.
 	webSocketSupported := false
-	for key := range supportedMap {
-		if key.IsRequirementMet(WebSocketExtension) {
+	for _, requirement := range supportedMap {
+		if requirement.IsRequirementMet(WebSocketExtension) {
```
omerlavanet, line 267: can break if true

protocol/chainlib/jsonRPC.go
```diff
@@ -61,7 +61,12 @@ func (apip *JsonRPCChainParser) getSupportedApi(name, connectionType string, int
...
+		if craftData.InternalPath != "" {
+			path = craftData.InternalPath
```
omerlavanet, line 66: we overwrite path here is that intended or we need to use path+/+internal_path?

protocol/chainlib/jsonRPC.go
```diff
@@ -496,88 +501,49 @@ func (apil *JsonRPCChainListener) GetListeningAddress() string {
...
-func (cp *JrpcChainProxy) start(ctx context.Context, nConns uint, nodeUrl common.NodeUrl, internalPaths map[string]struct{}) error {
-	if len(internalPaths) == 0 {
-		internalPaths = map[string]struct{}{"": {}} // add default path
+func (cp *JrpcChainProxy) start(ctx context.Context, nConns uint, nodeUrl common.NodeUrl) error {
+	conn, err := chainproxy.NewConnector(ctx, nConns, nodeUrl)
+	if err != nil {
+		return err
 	}
-	basePath := nodeUrl.Url
-	for path := range internalPaths {
```
omerlavanet, line 559: this functionality allowed providers to set only the base path, all internal paths were generated automatically, is that no longer supported? because it must be

protocol/chainlib/tendermintRPC.go
```diff
@@ -61,7 +61,12 @@ func (apip *TendermintChainParser) getSupportedApi(name, connectionType string)
 
 func (apip *TendermintChainParser) CraftMessage(parsing *spectypes.ParseDirective, connectionType string, craftData *CraftData, metadata []pairingtypes.Metadata) (ChainMessageForSend, error) {
 	if craftData != nil {
-		chainMessage, err := apip.ParseMsg("", craftData.Data, craftData.ConnectionType, metadata, extensionslib.ExtensionInfo{LatestBlock: 0})
+		path := craftData.Path
+		if craftData.InternalPath != "" {
+			path = craftData.InternalPath
```
omerlavanet, line 66: same Q, do we overwrite with internal path or append it to the path

protocol/rpcprovider/rpcprovider.go
```diff
@@ -361,6 +371,27 @@ func (rpcp *RPCProvider) SetupEndpoint(ctx context.Context, rpcProviderEndpoint
 		return utils.LavaFormatError("[PANIC] failed to RegisterForSpecUpdates, panic severity critical error, aborting support for chain api due to invalid chain parser, continuing with others", err, utils.Attribute{Key: "endpoint", Value: rpcProviderEndpoint.String()})
 	}
 
+	// warn if not all internal paths are configured
+	configuredInternalPaths := GetAllNodeUrlsInternalPaths(rpcProviderEndpoint.NodeUrls)
+	chainInternalPaths := chainParser.GetAllInternalPaths()
+	nonConfiguredInternalPaths := lavaslices.Difference(chainInternalPaths, configuredInternalPaths)
```
omerlavanet, line 377: if the provider defines the base path, we should be able to add all internal paths (because they are all relative to the base)
we need to maintain the auto generation of urls we had so far

protocol/chainlib/jsonRPC.go
shleikes: Yes, this is intended. You can look at the use of this field here:
https://github.com/lavanet/lava/blob/dcffaf73574f46cec25b511a02b12a2188e0ffa1/protocol/chainlib/jsonRPC.go#L117C4-L117C37

shleikes: Yes, this is intended. You can look at the use of this field here:

https://github.com/lavanet/lava/blob/6591537984b4e0e305ab4c781b7a6a1fad5ce5ce/protocol/chainlib/jsonRPC.go#L112

protocol/chainlib/jsonRPC.go
shleikes: Providers will still be able to configure only the base path. The chain router is now responsible to create several chain proxies, for each internal path. If the internal paths are not configured, the chain proxy won't be created.

protocol/chainlib/tendermintRPC.go
shleikes: Overwrite, read previous comment

protocol/chainlib/chainproxy/rpcclient/handler.go
```diff
@@ -303,6 +334,16 @@ func (h *handler) handleResponse(msg *JsonrpcMessage) {
 		go op.sub.run()
 		h.clientSubs[op.sub.subid] = op.sub
 	}
+
+	// This is because StarkNet Pathfinder is returning an integer instead of a string in the result
+	var integerSubId int
+	err := json.Unmarshal(msg.Result, &integerSubId)
```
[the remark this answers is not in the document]
shleikes: Fixed in 5768e9ce7e6c9ceb9c6202b952a9287f821841bf

protocol/chainlib/common.go
```diff
@@ -50,7 +51,13 @@ type VerificationKey struct {
 	Addon     string
 }
 
+type VerificationCollectionKey struct {
+	ApiInterface string
```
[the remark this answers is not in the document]
shleikes: Removed in 0fc7f381004a1717180f2891bfcb7ee0ae68578a

protocol/chainlib/chain_router.go
shleikes: Fixed in a8f7367d813366f78fbd459bd1d54ee3fe2eb53e

protocol/chainlib/chain_router.go
shleikes: Fixed in df21e04521b181134ddcb5c44f9b57da270b98db

ranlavanet (approved):

omerlavanet (approved):

PR #1688, opened 2024-09-11
fix: PRT: Fix all sync.Map uses

shleikes:


ranlavanet (approved):

protocol/chainlib/grpc.go
```diff
@@ -488,7 +468,7 @@ func (cp *GrpcChainProxy) SendNodeMsg(ctx context.Context, ch chan interface{},
 		}
 
 		// add the descriptor to the chainProxy cache
-		cp.descriptorsCache.setDescriptor(methodName, methodDescriptor)
+		cp.descriptorsCache.Store(methodName, methodDescriptor)
```
omerlavanet, line 471: prone to races, let's use loadOrStore?

protocol/rpcconsumer/rpcconsumer.go
```diff
@@ -229,46 +236,37 @@ func (rpcc *RPCConsumer) Start(ctx context.Context, options *rpcConsumerStartOpt
...
-					consumerConsistency = NewConsumerConsistency(chainID)
-					consumerConsistencies.Store(chainID, consumerConsistency)
-				} else {
-					var ok bool
-					consumerConsistency, ok = value.(*ConsumerConsistency)
-					if !ok {
-						err = utils.LavaFormatError("failed loading consumer consistency, value is of the wrong type", err, utils.Attribute{Key: "endpoint", Value: rpcEndpoint.Key()})
-						return err
-					}
+				if !loaded { // stored, use the new one
```
omerlavanet, line 249: repeating pattern let's use a generics function here

protocol/rpcconsumer/rpcconsumer.go
```diff
@@ -327,7 +325,10 @@ func (rpcc *RPCConsumer) Start(ctx context.Context, options *rpcConsumerStartOpt
 
 	utils.LavaFormatDebug("Starting Policy Updaters for all chains")
 	for chainId := range chainMutexes {
-		policyUpdater, ok := policyUpdaters.Load(chainId)
+		policyUpdater, ok, err := policyUpdaters.Load(chainId)
+		if err != nil {
+			utils.LavaFormatFatal("could not load policy Updater for chain", err, utils.LogAttr("chain", chainId))
```
omerlavanet, line 330: one chain will crash all others, shouldnt be fatal

omerlavanet (changes requested):

protocol/chainlib/grpc.go
shleikes: On the worst case scenario, where 2 threads are accessing the map at the same time, the same descriptor will be stored twice. I think that the current solution better than always creating the descriptor, since the outcome is the same.

protocol/rpcconsumer/rpcconsumer.go
shleikes: After a second look at this, the first too are redundant. Fix is here: c4f31b4f0d1b89fb58ef69a5cd719e72691877f0

protocol/rpcconsumer/rpcconsumer.go
shleikes: In this PR I tried to maintain the flow as it was before as much as I could. This is one example - it was fatal in this case before my changes. What should it be instead?

protocol/chainlib/grpc.go
omerlavanet: awesome, let's comment this and move on

protocol/rpcconsumer/rpcconsumer.go
omerlavanet: on the next line it errors out and continues, lets do that

protocol/chainlib/grpc.go
shleikes: Fixed: 397917ff5ab5815a617a0944b4ab4f55aff40f16

protocol/rpcconsumer/rpcconsumer.go
shleikes: Fixed: 9c59e9ca81e33b17b0b310f4e25524d153c0483b

ranlavanet (approved):

omerlavanet (approved):

omerlavanet (approved):

PR #1709, opened 2024-09-22
feat: PRT  - Add support for arrays in the parsing function ParseCanonical

shleikes:


shleikes:
Closed in favor of generic parsers in #1711

PR #1711, opened 2024-09-24
feat: PRT - Result generic parser

shleikes:


protocol/parser/parser.go
```diff
@@ -122,102 +124,119 @@ func parseInputFromParamsWithGenericParsers(rpcInput RPCInput, genericParsers []
...
-	parsedBlockInfo.parsedBlock = func() int64 {
-		// first we try to parse the value with the block parser
-		result, err := parse(rpcInput, blockParser, PARSE_PARAMS)
-		if err != nil || result == nil {
-			utils.LavaFormatDebug("ParseBlockFromParams - parse failed",
-				utils.LogAttr("error", err),
-				utils.LogAttr("result", result),
-				utils.LogAttr("blockParser", blockParser),
-				utils.LogAttr("rpcInput", rpcInput),
+	parsedBlock, err := rpcInput.ParseBlock(rawBlock)
```
ranlavanet: add unitest to validate the fix

ranlavanet (changes requested):

protocol/parser/parser.go
shleikes: Fixed in 2ac9725a327a5cf345edbebcb52d34509af3bdbe

ranlavanet (dismissed):

protocol/parser/parser.go
```diff
@@ -122,105 +123,120 @@ func parseInputFromParamsWithGenericParsers(rpcInput RPCInput, genericParsers []
...
+		parsedBlockInfo = NewParsedInput()
 	}
 
-	return response, nil
+	parsedRawBlock, _ := parseInputWithLegacyBlockParser(rpcInput, blockParser, source)
+	parsedBlockInfo.parsedDataRaw = unquoteString(parsedRawBlock)
+	return parsedBlockInfo
+}
+
+func ParseRawBlockFromParams(rpcInput RPCInput, blockParser spectypes.BlockParser, genericParsers []spectypes.GenericParser) *ParsedInput {
```
omerlavanet, line 188: comments, how is that different from the other parse block, why do we need the parseRawBlock? didn't see it being used

protocol/parser/parser.go
```diff
@@ -122,105 +123,120 @@ func parseInputFromParamsWithGenericParsers(rpcInput RPCInput, genericParsers []
...
-	if strings.Contains(response, "\"") {
-		responseUnquoted, err := strconv.Unquote(response)
-		if err != nil {
-			return response, nil
-		}
-		return responseUnquoted, nil
+	return resString, nil
+}
+
+func parseBlock(rpcInput RPCInput, blockParser spectypes.BlockParser, genericParsers []spectypes.GenericParser, source int) *ParsedInput {
```
omerlavanet, line 173: comment what it does

protocol/parser/parser.go
```diff
@@ -122,105 +123,120 @@ func parseInputFromParamsWithGenericParsers(rpcInput RPCInput, genericParsers []
...
 	}
-	return rpcInput.ParseBlock(result)
+	return unquoted
 }
 
 // This returns the parsed response after decoding
-func ParseFromReplyAndDecode(rpcInput RPCInput, resultParser spectypes.BlockParser) (string, error) {
-	response, err := ParseFromReply(rpcInput, resultParser)
+func ParseBlockHashFromReplyAndDecode(rpcInput RPCInput, resultParser spectypes.BlockParser, genericParsers []spectypes.GenericParser) (string, error) {
+	parsedInput, _ := parseInputWithGenericParsers(rpcInput, genericParsers)
```
omerlavanet, line 218: why on hashes we don't make sure the parsedSuccessfuly is set and on regular parsing it does, thats a discrepancy in how it works

protocol/parser/parser.go
```diff
@@ -254,21 +270,26 @@ func parse(rpcInput RPCInput, blockParser spectypes.BlockParser, dataSource int)
...
-	parsedBlock  int64
-	parsedHashes []string
+	parsedDataRaw string
+	parsedBlock   int64
+	parsedHashes  []string
 }
 
 func NewParsedInput() *ParsedInput {
 	return &ParsedInput{
-		parsedBlock:  spectypes.NOT_APPLICABLE,
```
omerlavanet, line 263: keep the default not applicable, it's a health habit

protocol/parser/parser.go
```diff
@@ -281,7 +302,11 @@ func (p *ParsedInput) GetBlockHashes() ([]string, error) {
 }
 
 func getMapForParse(rpcInput RPCInput) map[string]interface{} {
-	return map[string]interface{}{"params": rpcInput.GetParams(), "result": rpcInput.GetResult()}
+	var result map[string]interface{}
+	if rpcInput.GetResult() != nil {
```
omerlavanet, line 306: no need to call getResult twice, save the value and then unmarshal from it

omerlavanet (changes requested):

protocol/parser/parser.go
```diff
@@ -108,10 +109,10 @@ func parseInputFromParamsWithGenericParsers(rpcInput RPCInput, genericParsers []
...
-	parsedBlock := genericParserResult.GetBlock()
-	if parsedBlock != spectypes.NOT_APPLICABLE {
+	parsedBlock := genericParserResult.GetRawParsedData()
+	if parsedBlock != "" {
 		parsedSuccessfully = true
-		parsed.parsedBlock = parsedBlock
+		parsed.parsedDataRaw = parsedBlock
 	}
 
 	parsedBlockHashes, err := genericParserResult.GetBlockHashes()
```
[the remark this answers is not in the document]
shleikes: That's right. It doesn't matter if we parsed hashes or not, `parsedSuccessfully` will be true only if we parsed a raw block successfully.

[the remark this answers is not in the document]
shleikes: This is the current behavior, only because we still using the old block parser. Once we remove the old block parser, there will be no need for `parsedSuccessfully`

protocol/parser/parser.go
```diff
@@ -108,10 +109,10 @@ func parseInputFromParamsWithGenericParsers(rpcInput RPCInput, genericParsers []
 	}
 
 	parsed := NewParsedInput()
-	parsedBlock := genericParserResult.GetBlock()
-	if parsedBlock != spectypes.NOT_APPLICABLE {
+	parsedBlock := genericParserResult.GetRawParsedData()
```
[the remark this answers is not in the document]
shleikes: Rename in Fix in a66645c87f3dd2566291e4ba042fb1c28bfebc0a

protocol/parser/parser.go
```diff
@@ -122,105 +123,120 @@ func parseInputFromParamsWithGenericParsers(rpcInput RPCInput, genericParsers []
...
-	parsedBlockInfo.parsedBlock = func() int64 {
-		// first we try to parse the value with the block parser
-		result, err := parse(rpcInput, blockParser, PARSE_PARAMS)
-		if err != nil || result == nil {
-			utils.LavaFormatDebug("ParseBlockFromParams - parse failed",
-				utils.LogAttr("error", err),
-				utils.LogAttr("result", result),
-				utils.LogAttr("blockParser", blockParser),
-				utils.LogAttr("rpcInput", rpcInput),
+func ParseRawBlock(rpcInput RPCInput, parsedInput *ParsedInput, defaultValue string) {
```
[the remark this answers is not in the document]
shleikes: Basically just a wrapper for `rpcInput.ParseBlock(rawBlock)`.
The result is then stored in `ParsedInput`.
On a failed parse, it will parse the default value and store it in `ParsedInput`.

protocol/parser/parser.go
shleikes: Was originally used by the `Verify` function of the chain fetcher, but then changed it again and forgot to remove.
Removed in 3e69eaa719b0786f13c04d6362a0d566df004505

protocol/parser/parser.go
shleikes: Same as previous comment
> It doesn't matter if we parsed hashes or not, parsedSuccessfully will be true only if we parsed a raw block successfully.

protocol/parser/parser.go
shleikes: Added comment in 79e54af9705b33196af60d4bf41aed161b0b344a

protocol/parser/parser.go
shleikes: Fixed in 9d88cd7ab3cdde1f27be72c2cf857f06cfb5700e

protocol/parser/parser.go
shleikes: Fixed in b971e49be1a56fa62241b86af335e1c9362bc33f

protocol/parser/parser.go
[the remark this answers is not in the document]
ranlavanet: lets rename it to parsedBlock

protocol/parser/parser.go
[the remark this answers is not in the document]
shleikes: Added comment here a00402bde70748f08989ee266f91557a843d4e2d

protocol/parser/parser.go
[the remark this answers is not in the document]
shleikes: Renamed in cb26380c328d45aad01aa15fc27a86f962876786

protocol/parser/parser.go
shleikes: Added comment in 79e54af9705b33196af60d4bf41aed161b0b344a

protocol/parser/parser.go
```diff
@@ -275,9 +287,11 @@ type ParsedInput struct {
 	parsedHashes  []string
 }
 
+const RAW_NOT_APPLICABLE = "-1"
+
 func NewParsedInput() *ParsedInput {
 	return &ParsedInput{
```
omerlavanet, line 293: you didn't fix my comment here, add the parsed block with a default value of not applicable, the default value of 0 is wrong as it can be a valid block

protocol/parser/parser.go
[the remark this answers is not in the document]
omerlavanet: to be clear: add the parsed successfully to the hashes flow, i don't care if in our specific case right now the caller doesnt have to have it for now or is aware what is going on inside, currently this is just bad practice and this code will be relevant in the future

protocol/parser/parser.go
omerlavanet: you did a removel: [-] parsedBlock:  spectypes.NOT_APPLICABLE i commented to keep it

omerlavanet (changes requested):
fix me pls

protocol/parser/parser.go
shleikes: Fixed in 722362c25e9ed43f272989461697a7cc5d4ca993

protocol/parser/parser.go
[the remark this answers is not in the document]
shleikes: Fixed in 95e832b11a96e03393e9f4bef9c96d143617b3f9

omerlavanet (approved):

PR #1716, opened 2024-09-26
test: PRT - Attempt on fixing the statistical cache integration test failure

shleikes:


protocol/integration/protocol_test.go
```diff
@@ -1815,10 +1815,11 @@ func TestConsumerProviderWithProviderSideCache(t *testing.T) {
...
 
-	for i := 0; i < 5; i++ {
-		// Get block
-		sendMessage("block", []string{"1000"})
-	}
+	// Get block, this should be cached for next time
+	sendMessage("block", []string{"1000"})
+
+	// Get block again, this time it should be from cache
+	sendMessage("block", []string{"1000"})
```
ranlavanet, line 1822: it might fail because a routine will add to cache. so race here is gonna be constantly happening. verify its in cache before checking the result. 

do some thing like
for {
    if cache.inputFound("1000") { break} 
}

ranlavanet (changes requested):

protocol/integration/protocol_test.go
shleikes: Fixed in aacaa362ff440265da81f630258abf9cdec22b73

ranlavanet (approved):

omerlavanet (approved):

PR #1718, opened 2024-09-26
test: PRT - Attempt to fix statistical optimizer test

shleikes:


ranlavanet (approved):

shleikes:
Still happens after setting a delta of 10 to the requirement:
```
Max difference between 346 and 300 allowed is 10, but difference was 46
```

@omerlavanet Any ideas?

shleikes:
Close reason: inactivity

shleikes:
Close reason: inactivity

PR #1723, opened 2024-10-01
fix: PRT - Replace the method in the debug verification in EVMOS spec

shleikes:

PR #1725, opened 2024-10-01
fix: PRT - Add EVMOS missing debug APIs + remove unsupported

shleikes:
Depends on / can replace PR #1723


cookbook/specs/evmos.json
```diff
@@ -1590,6 +1590,367 @@
...
+                "category": {
+                  "deterministic": false,
+                  "local": false,
+                  "subscription": false,
+                  "stateful": 0
+                },
+                "extra_compute_units": 0
+              },
+              {
+                "name": "debug_setGCPercent",
```
omerlavanet, line 1756: things that affect the node shouldn't be in the spec, we don't add all apis blindly only those that we want exposed, please go over all the debug apis and only leave those that give information about the chain
for example a query that returns the node cpu shouldn't be here
imagine an api that can shut down the service for others or cause memory or cpu consumption costs or disturbance to others, we don't allow these

omerlavanet (changes requested):

cookbook/specs/evmos.json
shleikes: Removed in ff36877946494ba331925133a51720baeeed5987

omerlavanet (approved):

omerlavanet (approved):

PR #1739, opened 2024-10-13
fix: PRT - Fix the bug that prevented existing addons to be enabled after spec  update

shleikes:
Co-authored with @Yaroms

### DIFF
--- a/protocol/chainlib/base_chain_parser.go
+++ b/protocol/chainlib/base_chain_parser.go
@@ -242,11 +242,12 @@ func (bcp *BaseChainParser) Construct(spec spectypes.Spec, internalPaths map[str
 	bcp.verifications = verifications
 	allowedAddons := map[string]bool{}
 	allowedExtensions := map[string]struct{}{}
-	for _, apoCollection := range apiCollections {
-		for _, extension := range apoCollection.Extensions {
+	for _, apiCollection := range apiCollections {
+		for _, extension := range apiCollection.Extensions {
 			allowedExtensions[extension.Name] = struct{}{}
 		}
-		allowedAddons[apoCollection.CollectionData.AddOn] = false
+		// if addon was already existing (happens on spec update), use the existing policy, otherwise set it to false by default
+		allowedAddons[apiCollection.CollectionData.AddOn] = bcp.allowedAddons[apiCollection.CollectionData.AddOn]
 	}
 	bcp.allowedAddons = allowedAddons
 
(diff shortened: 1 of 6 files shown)


protocol/chainlib/base_chain_parser.go
```diff
@@ -242,11 +242,12 @@ func (bcp *BaseChainParser) Construct(spec spectypes.Spec, internalPaths map[str
...
 	allowedExtensions := map[string]struct{}{}
-	for _, apoCollection := range apiCollections {
-		for _, extension := range apoCollection.Extensions {
+	for _, apiCollection := range apiCollections {
+		for _, extension := range apiCollection.Extensions {
 			allowedExtensions[extension.Name] = struct{}{}
 		}
-		allowedAddons[apoCollection.CollectionData.AddOn] = false
+		// if addon was already existing (happens on spec update), use the existing policy, otherwise set it to false by default
+		allowedAddons[apiCollection.CollectionData.AddOn] = bcp.allowedAddons[apiCollection.CollectionData.AddOn]
```
ranlavanet, line 250: can be bcp.allowedAddons can be nil. 

you need to iterate over bcp.allowedAddons and for any true value that also exists in allowedAddons set it to true.

ranlavanet (changes requested):

protocol/chainlib/base_chain_parser.go
```diff
@@ -242,11 +242,12 @@ func (bcp *BaseChainParser) Construct(spec spectypes.Spec, internalPaths map[str
 	bcp.verifications = verifications
 	allowedAddons := map[string]bool{}
 	allowedExtensions := map[string]struct{}{}
-	for _, apoCollection := range apiCollections {
-		for _, extension := range apoCollection.Extensions {
+	for _, apiCollection := range apiCollections {
+		for _, extension := range apiCollection.Extensions {
 			allowedExtensions[extension.Name] = struct{}{}
```
ranlavanet, line 247: also you need a similar behavior to extensions.

call getConfiguredExtensions from the current bcp.extensionParser 
and set these extensions using 

func (ep *ExtensionParser) SetConfiguredExtensions(configuredExtensions map[ExtensionKey]*spectypes.Extension) {
	ep.configuredExtensions = configuredExtensions
}

ranlavanet (changes requested):

protocol/chainlib/base_chain_parser.go
shleikes: No need, already happening. Made a small fix here:

ranlavanet (approved):

omerlavanet (approved):
this PR is 10/10 what a nice code

PR #1751, opened 2024-10-20
feat: PRT - Check the lava chain id in provider

shleikes:


protocol/rpcprovider/rpcprovider.go
```diff
@@ -654,6 +654,18 @@ rpcprovider 127.0.0.1:3333 OSMOSIS tendermintrpc "wss://www.node-path.com:80,htt
 			}
 			utils.LavaFormatInfo("Running with chain-id:" + networkChainId)
 
+			resultStatus, err := clientCtx.Client.Status(ctx)
+			if err != nil {
+				utils.LavaFormatFatal("failed to get status from node", err)
+			}
+
+			if networkChainId != resultStatus.NodeInfo.Network {
+				utils.LavaFormatFatal("lava chain-id from flag does not match the lava chain-id of the node", nil,
```
ranlavanet, line 663: we already have this check for both consumer and provider on state tracker 
https://github.com/lavanet/lava/blob/1962ab0381ddd80048d400d8d83a5460781cf577/protocol/statetracker/state_tracker.go#L102

ranlavanet (changes requested):

protocol/rpcprovider/rpcprovider.go
shleikes: You are right, removed in a170e5b3ddca3966fe8ff4354bbf5e767fddd2d9

ranlavanet (approved):

PR #1783, opened 2024-11-20
feat: PRT - Near shard spec verification

shleikes:


omerlavanet (changes requested):
please add unitests for the new result parsing

PR #1881, opened 2025-01-02
feat: PRT - Add provider endpoint to metrics by option

shleikes:


protocol/metrics/mapped_labels_gauge_vec.go
```diff
@@ -0,0 +1,40 @@
+package metrics
+
+import "github.com/prometheus/client_golang/prometheus"
+
```
ranlavanet, line 4: add description why we need this, 
explain run time labels etc..

ranlavanet (changes requested):

ranlavanet (approved):

PR #1897, opened 2025-01-07
fix: PRT - Fix a bug where static provider won't use relay metrics

shleikes:


protocol/rpcprovider/rpcprovider_server.go
```diff
@@ -247,51 +247,58 @@ func (rpcps *RPCProviderServer) Relay(ctx context.Context, request *pairingtypes
...
-
 	if err != nil || common.ContextOutOfTime(ctx) {
-		// failed to send relay. we need to adjust session state. cuSum and relayNumber.
-		relayFailureError := rpcps.providerSessionManager.OnSessionFailure(relaySession, request.RelaySession.RelayNum)
-		if relayFailureError != nil {
-			var extraInfo string
-			if err != nil {
-				extraInfo = err.Error()
+		// static provider doesn't handle sessions
+		if !rpcps.StaticProvider {
```
ranlavanet, line 252: This is really ugly. 

lets just keep the previous return early code and just write to mertics as well.

ranlavanet (changes requested):

protocol/rpcprovider/rpcprovider_server.go
shleikes: That's even more ugly... I'll think of something else

shleikes: Fixed in 1df9f1b86c1c4940c3cd5af0c837cdc719ad3aca

protocol/rpcprovider/rpcprovider_server.go
```diff
@@ -312,6 +273,51 @@ func (rpcps *RPCProviderServer) Relay(ctx context.Context, request *pairingtypes
 	return reply, rpcps.handleRelayErrorStatus(err)
 }
 
+func (rpcps *RPCProviderServer) logSession(isRelayError bool, ctx context.Context, consumerAddress sdk.AccAddress, reply *pairingtypes.RelayReply, chainMessage chainlib.ChainMessage, relaySession *lavasession.SingleProviderSession, request *pairingtypes.RelayRequest, err error) error {
```
ranlavanet, line 276: Lets make it finalizeSession

ranlavanet (changes requested):

ranlavanet (approved):

PR #1914, opened 2025-01-19
refactor: PRT - QoS Manager Refactor - Part 2

shleikes:


ranlavanet:
forked too much from main. safer to reprogram straight from main.

PR #2006, opened 2025-04-25
feat: FB - Added support for pass_nullify

fbjohnny:
* Added support for `pass_nullify` in specs, allowing the removal of headers from being sent to the rpc node.
* This fixes Stellar (and possibly other nodes) which cannot process POST requests with application/json as content type.


protocol/chainlib/rest.go
```diff
@@ -521,7 +521,11 @@ func (rcp *RestChainProxy) SendNodeMsg(ctx context.Context, ch chan interface{},
 
 	if len(nodeMessage.GetHeaders()) > 0 {
 		for _, metadata := range nodeMessage.GetHeaders() {
-			req.Header.Set(metadata.Name, metadata.Value)
+			if metadata.Value == "" {
```
oren-lava, line 524: handle this also on other api interfaces as well. In other words, delete empty value header also in jsonRPC, tendermintRPC, and gRPC (not only REST)

proto/lavanet/lava/spec/api_collection.proto
```diff
@@ -61,6 +61,7 @@ message Header {
     pass_reply = 1;
     pass_both = 2;
     pass_ignore = 3; // allows it to pass around but is not signed
+    pass_nullify = 4; // removes a header from being sent
```
oren-lava, line 64: being sent to where? elaborate more for documentation

oren-lava (changes requested):

protocol/chainlib/rest.go
fbjohnny: done

proto/lavanet/lava/spec/api_collection.proto
fbjohnny: done

PR #2014, opened 2025-05-07
feat: pass_override support

fbjohnny:
* Support for pass_override for headers in spec. Needed to override headers before sending them to the RPC (for example, Stellar requires its content-type to be of specific value, not application/json)


protocol/chainlib/chainproxy/rpcclient/client.go
```diff
@@ -284,6 +284,22 @@ func (c *Client) SetHeader(key, value string) {
 	conn.mu.Unlock()
 }
 
+// SetHeader deletes a HTTP header to the client's requests.
```
Yaroms, line 287: fix comment

Yaroms (approved):

PR #2057, opened 2025-08-26
fix: The following has been fixed in the quorum feature

AnnaR-prog:
Fixes are done in the following:
1. Retry condition when the quorum is not met
2. Calculation of actualQuorumNeeded number after every Retry
3. Error represented to the user after quorum is not met

### DIFF
--- a/protocol/rpcconsumer/relay_processor.go
+++ b/protocol/rpcconsumer/relay_processor.go
@@ -222,7 +222,7 @@ func (rp *RelayProcessor) HasRequiredNodeResults(tries int) (bool, int) {
 	resultsCount, nodeErrors, specialNodeErrors, _ := rp.GetResults()
 
 	hash, hashErr := rp.getInputMsgInfoHashString()
-	neededForQuorum := int(math.Ceil(rp.quorumParams.Rate * float64(max(rp.quorumParams.Min, tries))))
+	neededForQuorum := rp.getActualQuorumSize(resultsCount)
 	if rp.quorumParams.Enabled() && neededForQuorum <= rp.highestQourum ||
 		!rp.quorumParams.Enabled() && resultsCount >= rp.quorumParams.Min {
 		if hashErr == nil { // Incase we had a successful relay we can remove the hash from our relay retries map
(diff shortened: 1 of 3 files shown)


protocol/rpcconsumer/relay_processor.go
```diff
@@ -237,20 +237,83 @@ func (rp *RelayProcessor) HasRequiredNodeResults(tries int) (bool, int) {
...
+		var retryForQuorumNeeded bool
+
+		if rp.quorumParams.Enabled() {
+			// Quorum feature enabled: check if quorum is still mathematically possible and quorum is not yet reached
+			maxRemainingProviders := rp.quorumParams.Max - (nodeErrors + resultsCount)
+			// The following line checks if, after accounting for the maximum possible additional responses (maxRemainingProviders)
+			// and the current highest number of matching responses (rp.highestQourum), it is still mathematically possible
+			// to reach the required quorum threshold (calculated as quorum rate * max providers).
+			// If not, then retrying is not needed because quorum cannot be achieved anymore.
+			retryForQuorumNeeded = maxRemainingProviders+rp.highestQourum >= int(math.Ceil(rp.quorumParams.Rate*float64(rp.quorumParams.Max)))
```
avitenzer: Lets change the param name from highestQourum to currentEquivalentResults

protocol/rpcconsumer/relay_processor.go
```diff
@@ -380,11 +460,24 @@ func (rp *RelayProcessor) responsesQuorum(results []common.RelayResult, quorumSi
...
+			// Quorum feature disabled - return original error message
+			return nil, utils.LavaFormatInfo("majority count is less than quorumSize",
+				utils.LogAttr("nilReplies", nilReplies),
+				utils.LogAttr("results", len(results)),
+				utils.LogAttr("maxCount", maxCount),
+				utils.LogAttr("quorumSize", quorumSize))
+		}
 	}
 
 	mostCommonResult.Quorum = maxCount
```
avitenzer, line 482: Let's change maxCount to equivalentResults

AnnaR-prog: Done: changed to currentQourumEqualResults (as we agreed)

protocol/rpcconsumer/relay_processor.go
AnnaR-prog: Done: changed to currentQourumEqualResults (as we agreed)

protocol/rpcconsumer/relay_processor.go
AnnaR-prog: Done: sorry, changed to quorumEqualResults

avitenzer (approved):

PR #2260, opened 2026-04-07
feat: Added support of Open Telemetry tracing for Smart Router

sotskov-do:
# Description

Closes: #XXXX

---

## Author Checklist

I have...

## Reviewers Checklist

I have...


protocol/rpcsmartrouter/rpcsmartrouter.go
```diff
@@ -1585,6 +1602,19 @@ rpcsmartrouter smartrouter_examples/full_smartrouter_example.yml --cache-be "127
...
+	// come from OTEL_* environment variables per the SDK spec — see the
+	// tracing.TraceBodyFlag doc-comment for the full list. Tracing is enabled
+	// unless OTEL_SDK_DISABLED=true or OTEL_TRACES_EXPORTER=none; when no OTLP
+	// endpoint is configured the SDK falls back to the spec default
+	// (localhost:4317 for gRPC, localhost:4318 for HTTP).
+	// --otel-trace-body is the only Lava-specific knob, exposed as a CLI flag
+	// because it's a per-invocation debug toggle rather than deployment
+	// configuration. Body size is delegated to the SDK via
+	// OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT (SDK default: unlimited).
+	cmdRPCSmartRouter.Flags().Bool(tracing.TraceBodyFlag, false, "record request/response bodies on trace spans (size limit delegated to OTel SDK via OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT)")
```
NadavLevi, line 1616: `--otel-trace-body` is registered on Cobra here but read via `viper.GetBool(tracing.TraceBodyFlag)` at line 1477 without a corresponding `viper.BindPFlag`. Other flags in this file explicitly bind to viper (lines 1515, 1531–1543, 1593). As-is, viper returns the default (`false`) regardless of CLI input, so the flag is a no-op.

Suggested fix — add right after the `Flags().Bool(...)` registration:

```go
if err := viper.BindPFlag(tracing.TraceBodyFlag, cmdRPCSmartRouter.Flags().Lookup(tracing.TraceBodyFlag)); err != nil {
    log.Fatalln("unable to bind otel-trace-body flag", err)
}
```

protocol/rpcsmartrouter/rpcsmartrouter_server.go
```diff
@@ -770,16 +821,27 @@ func (rpcss *RPCSmartRouterServer) sendRelayToDirectEndpoints(
...
 		go func(endpointAddress string, sessionInfo *lavasession.SessionInfo) {
 			// Derive from ctx so IP forwarding metadata (and other values) are preserved.
 			goroutineCtx, goroutineCtxCancel := context.WithCancel(ctx)
 
-			guid, found := utils.GetUniqueIdentifier(ctx)
-			if found {
-				goroutineCtx = utils.WithUniqueIdentifier(goroutineCtx, guid)
-			}
+			guid, _ := utils.GetUniqueIdentifier(ctx)
+			goroutineCtx = utils.WithUniqueIdentifier(goroutineCtx, guid)
```
NadavLevi, line 837: Behavior change in this PR: `utils.WithUniqueIdentifier` (utils/uniqueIdentifier.go:14) writes unconditionally, so when `GetUniqueIdentifier` returns `(0, false)` this line stores `uint64(0)` into `goroutineCtx`.

Downstream, `utils.AppendUniqueIdentifier` (utils/uniqueIdentifier.go:19) skips when `ctx.Value(...) != nil` — `uint64(0)` satisfies that — so real GUID generation later in the pipeline is blocked. Previously the write was gated by `if found`.

Suggested fix — restore the guard and still capture `guid` for the span (0 is fine as 'unknown'):

```go
guid, found := utils.GetUniqueIdentifier(ctx)
if found {
    goroutineCtx = utils.WithUniqueIdentifier(goroutineCtx, guid)
}
// ...
tracing.RecordProviderAttributes(provSpan, guid, endpointAddress)
```

NadavLevi (commented):
Review of the OTel tracing changes — 5 inline comments. Two are functional bugs (body flag dead, GUID forced to 0); the other three are smaller correctness/perf notes.

protocol/rpcsmartrouter/rpcsmartrouter_server.go
```diff
@@ -517,7 +547,12 @@ func (rpcss *RPCSmartRouterServer) SendParsedRelay(
 		return nil, err
 	}
 
+	ctx, processSpan := tracing.StartInternalSpan(ctx, tracing.SpanProcessingResult)
+	defer processSpan.End()
```
NadavLevi, line 551: `processSpan` is named `SpanProcessingResult` but `defer processSpan.End()` at this scope keeps it open until `SendParsedRelay` returns — so the measured duration includes `appendHeadersToRelayResult`, analytics assignment, and `LogRelay`, not just `ProcessingResult()`. The span's timing no longer matches its name.

Suggested fix — scope the span to just the ProcessingResult call:

```go
returnedResult, err := func() (*common.RelayResult, error) {
    _, processSpan := tracing.StartInternalSpan(ctx, tracing.SpanProcessingResult)
    defer processSpan.End()
    r, e := relayProcessor.ProcessingResult()
    if r != nil && r.Reply != nil {
        tracing.RecordBody(processSpan, tracing.AttrRelayResponseBody, r.Reply.Data)
    }
    return r, e
}()
```

protocol/rpcsmartrouter/rpcsmartrouter_server.go
```diff
@@ -1521,6 +1595,9 @@ func (rpcss *RPCSmartRouterServer) sendRelayToEndpoint(
 					}) // caching in the consumer doesn't care about hashes, and we don't have data on finalization yet
 					cancel()
 					cacheLatencyMs := float64(time.Since(cacheStart).Milliseconds())
+					cacheHit := cacheError == nil && cacheReply != nil && cacheReply.GetReply() != nil
+					tracing.RecordCacheResult(ctx, cacheSpan, cacheHit, cacheLatencyMs)
+					cacheSpan.End()
```
NadavLevi, line 1600: Minor: `cacheSpan.End()` is called explicitly rather than deferred. If anything between line 1584 and here panics, the span leaks. Defensive nit — prefer `defer cacheSpan.End()` right after the span is started (ideally wrapping this whole block in a helper or IIFE so the defer fires on any exit).

protocol/tracing/span_helpers.go
```diff
@@ -0,0 +1,291 @@
...
+	span.RecordError(err)
+	span.SetStatus(otelcodes.Error, err.Error())
+}
+
+// RecordBody conditionally records a body attribute on the span when
+// --otel-trace-body is enabled. Truncation of the recorded value is handled
+// by the SDK's SpanLimits, which reads OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT
+// (default: unlimited). Operators who want a body cap should set that env var.
+func RecordBody(span trace.Span, attrKey string, body []byte) {
+	if traceBodyEnabled && span.IsRecording() {
```
NadavLevi, line 96: Perf note: `string(body)` copies the full byte slice before `SetAttributes` applies the SDK's `OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT` truncation. For a large response body with a small configured limit, the full payload is still allocated only to be thrown away.

Suggested fix — read the env var once during `New()` and pre-truncate here:

```go
var bodyAttrLimit int // set from OTEL_SPAN_ATTRIBUTE_VALUE_LENGTH_LIMIT, 0 = unlimited

func RecordBody(span trace.Span, attrKey string, body []byte) {
    if !traceBodyEnabled || !span.IsRecording() {
        return
    }
    if bodyAttrLimit > 0 && len(body) > bodyAttrLimit {
        body = body[:bodyAttrLimit]
    }
    span.SetAttributes(attribute.String(attrKey, string(body)))
}
```

protocol/rpcsmartrouter/rpcsmartrouter.go
sotskov-do: Done

protocol/rpcsmartrouter/rpcsmartrouter_server.go
sotskov-do: Done

protocol/rpcsmartrouter/rpcsmartrouter_server.go
sotskov-do: Done

protocol/rpcsmartrouter/rpcsmartrouter_server.go
sotskov-do: Done

protocol/tracing/span_helpers.go
sotskov-do: Done

NadavLevi (dismissed):

protocol/tracing/propagation.go
avitenzer, line 1: **Issue:** If the inbound metadata carries both "traceparent" and "Traceparent" (case mismatch), only the first is updated; the second remains with whatever stale value it    
  had. The whole point of replace-in-place was to avoid duplicate traceparent downstream — this case slips through.
                                                                                                                                                                             
  **Suggestion:** track which positions matched and either overwrite both with the new value or null out the duplicates. Or, since this is a defensive measure for an unlikely   
  upstream bug, just collapse case-folded duplicates before injecting.

protocol/rpcsmartrouter/rpcsmartrouter.go
avitenzer, line 1: **Issue: **[]byte(req) allocates a fresh copy of the entire request string before RecordBody is even called. RecordBody's early-return on traceBodyEnabled happens after the   
  conversion has already cost an allocation + memcpy of the full body. For high-RPS deployments with non-trivial request sizes, this is a measurable hit purely from
  instrumentation that the operator hasn't opted into.                                                                                                                       
                             
  tracing.IsTraceBodyEnabled() already exists for exactly this reason — it's documented as the "instrumentation-site check" — but no call site uses it.                      
   
  Suggestion (option A — gate at call site):                                                                                                                                 
  if tracing.IsTraceBodyEnabled() {
      tracing.RecordBody(span, tracing.AttrRelayRequestBody, []byte(req))                                                                                                    
  }                                                                      
                                                                                                                                                                             
  Suggestion (option B — preferred, also avoids the inner string(body) round-trip): add a string-typed variant in span_helpers.go:
  func RecordBodyString(span trace.Span, attrKey, body string) {                                                                                                             
      if !traceBodyEnabled || !span.IsRecording() {             
          return                                                                                                                                                             
      }                      
      if bodyAttrLimit > 0 && len(body) > bodyAttrLimit {                                                                                                                    
          body = body[:bodyAttrLimit]                    
      }                                                                                                                                                                      
      span.SetAttributes(attribute.String(attrKey, body))
  }                                                                                                                                                                          
  and use it here: tracing.RecordBodyString(span, tracing.AttrRelayRequestBody, req). Keeps the existing RecordBody([]byte) for callers (e.g. r.Reply.Data) that genuinely   
  have bytes.

protocol/tracing/propagation.go
sotskov-do: Done

protocol/rpcsmartrouter/rpcsmartrouter.go
sotskov-do: Done

avitenzer (approved):

PR #2269, opened 2026-04-14
perf(chainlib): zero-copy passthrough in checkUTXOResponseAndFixReply…

NadavLevi:
… for non-UTXO chains

The function unconditionally did `string(replyData)` on entry, which allocated a full copy of every upstream reply body — even for non-UTXO chains that then returned immediately without using it. A live pprof capture on the eth router attributed ~4.5 GB of inuse memory over a 12-minute window to this single allocation.

Return []byte and only reformat on the UTXO path; non-UTXO chains (ETH, Cosmos, tendermint, etc.) now pass through the original slice with no copy. Test coverage adds a pointer-identity assertion so any future regression that reintroduces an eager copy fails loudly.

# Description

Closes: #XXXX

---

## Author Checklist

I have...

## Reviewers Checklist

I have...

### DIFF
--- a/protocol/chainlib/common.go
+++ b/protocol/chainlib/common.go
@@ -161,20 +161,22 @@ func isUTXOFamily(chainID string) bool {
 	return chainID == "BTC" || chainID == "BTCT" || chainID == "LTC" || chainID == "LTCT" || chainID == "DOGE" || chainID == "DOGET" || chainID == "BCH" || chainID == "BCHT"
 }
 
-func checkUTXOResponseAndFixReply(chainID string, replyData []byte) string {
-	response := string(replyData)
+// checkUTXOResponseAndFixReply returns the reply body for UTXO-family chains after
+// reformatting it for JSON-RPC 1.0 semantics, and returns replyData unchanged for all
+// other chains. The non-UTXO path is zero-copy: hot chains (ETH, Cosmos, etc.) no
+// longer pay a full string(replyData) allocation per response.
+func checkUTXOResponseAndFixReply(chainID string, replyData []byte) []byte {
 	if !isUTXOFamily(chainID) {
-		return response
+		return replyData
 	}
 
 	// Try single response first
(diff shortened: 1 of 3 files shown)


protocol/chainlib/jsonRPC.go
avitenzer, line 1: ▎ **Issue:** string(response) is called twice unconditionally. For the ETH hot path (the one cited in the pprof), this is 2 allocations vs the old 1. The function-level saving is real but cancelled here.
  ▎                                                                                                                                                                                                                                                          
  ▎ **Suggestion:** Change addHeadersAndSendString to accept []byte (Fiber supports this natively) and gate the log conversion:                                                                                                                                  
  ▎                                       
  ▎ // Only convert to string if the log level will actually emit                                                                                                                                                                                            
  ▎ if utils.LavaFormatLogger.Level() <= logLevel {                                                                                                                                                                                                          
  ▎     utils.LavaFormatInfo("...", string(response), ...)                                                                                                                                                                                                   
  ▎ }                                                                                                                                                                                                                                                        
  ▎ err = addHeadersAndSendBytes(fiberCtx, reply.GetMetadata(), response)
  ▎ This is where the real 4.5 GB win lives.

protocol/chainlib/common.go
avitenzer, line 1: ▎ **Issue:** The return replyData at the bottom of the function is the fallback for "batch unmarshal failed OR batch marshal failed," but it sits after two closing braces with no comment, making it look like dead code at a glance.                         
  ▎                                                         
  ▎ **Suggestion:**                                                                                                                                                                                                                                              
  ▎     if marshaledRes, err := json.Marshal(btcBatch); err == nil {
  ▎         return marshaledRes                                                                                                                                                                                                                              
  ▎     }                                                   
  ▎ }                                                                                                                                                                                                                                                        
  ▎ // Neither single nor batch JSON-RPC decode succeeded; return original bytes unchanged.
  ▎ return replyData

avitenzer (approved):

PR #2270, opened 2026-04-14
perf(chainlib): send response bodies as bytes via fiber.Ctx.Send

NadavLevi:
addHeadersAndSendString called fiber.Ctx.SendString, which copies the string into fasthttp's response bytebufferpool. All callers in the relay path already hold the body as []byte (reply.Data) or as JSON marshal output (also []byte) and were round-tripping through string() at the call site. That string() conversion is a full per-request copy on top of the unavoidable copy into fasthttp's buffer — visible in pprof as 96 GB inuse under bytebufferpool.(*ByteBuffer).WriteString on the eth router.

Rename to addHeadersAndSendBytes, take []byte, and use fiber.Ctx.Send. Update every caller (jsonRPC, tendermintRPC, rest) to drop the string() round-trip on success paths where reply.Data is already []byte; error paths keep []byte(s) on the small error strings since those aren't on the hot path. Add focused tests covering body + metadata headers, the empty-body case, and a binary payload round-trip to guard against a regression that reintroduces the string() conversion.

# Description

Closes: #XXXX

---

## Author Checklist

I have...

## Reviewers Checklist

I have...

### DIFF
--- a/protocol/chainlib/common.go
+++ b/protocol/chainlib/common.go
@@ -212,18 +212,10 @@ func convertToUTXOResponse(msg *rpcclient.JsonrpcMessage) *rpcclient.BTCResponse
 	}
 }
 
-func addHeadersAndSendString(c *fiber.Ctx, metaData []pairingtypes.Metadata, data string) error {
-	for _, value := range metaData {
-		c.Set(value.Name, value.Value)
-	}
-
-	return c.SendString(data)
-}
-
-// addHeadersAndSendBytes is the []byte counterpart of addHeadersAndSendString.
-// Used on the ETH/Cosmos hot path so the upstream reply body flows straight to
-// fiber.Ctx.Send without a string(response) detour — which would otherwise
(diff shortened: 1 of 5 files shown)


protocol/chainlib/common.go
avitenzer, line 1: **Issue:** Every error path now has []byte(convertToJsonError(errMasking)). There are 4 such sites across 3 files. Each adds an allocation (though error paths are not hot).
  ▎                                                                                                                                                                                                                                                          
  ▎ **Suggestion:** Change convertToJsonError to return []byte. This eliminates all 4 conversions in one place:                                                                                                                                                  
  ▎                                       
  ▎ func convertToJsonError(errorMsg string) []byte {                                                                                                                                                                                                        
  ▎     // existing body, just return []byte instead of string                                                                                                                                                                                               
  ▎ }

avitenzer (dismissed):

PR #2272, opened 2026-04-15
perf(router): skip upstream gzip auto-decode on smart-router HTTP req…

NadavLevi:
…uests

Go's default http.Transport auto-adds `Accept-Encoding: gzip` on outbound requests when no Accept-Encoding is set, and transparently decodes responses via net/http.(*http2gzipReader).Read → compress/gzip.Reader → compress/flate. Production pprof on the eth router attributed ~30-39% of total CPU to that inflate chain — CPU we spend decoding bytes we then either pass through, re-encode with gzip for the client, cache as a blob, or peek at with CheckResponseError.

Two coordinated changes remove that work *only on the smart router*:

  1. Switch lavasession.HTTPDirectRPCConnection from its own fresh &http.Client{} (default transport — no pooling, no TLS session cache, no HTTP/2) to common.OptimizedHttpClient(). Every smart-router HTTP connection now shares common.SharedHttpTransport() and inherits connection pooling, TLS session resumption, and ForceAttemptHTTP2.

  2. Set `Accept-Encoding: identity` on every outbound request from HTTPDirectRPCConnection.SendRequest and DoHTTPRequest. Go only auto-adds `gzip` when the caller left Accept-Encoding empty; once identity is set, both the auto-add and auto-decode are skipped.

The scoping matters: the shared transport keeps its default compression behavior, so provider chain proxies (chainlib/rest.go, tendermintRPC.go, chainproxy/rpcclient/http.go) continue to auto-gzip as before. The opt-out lives exclusively on the smart-router request path.

Measured impact on the eth router (same 900s pprof window, back-to-back before/after the deploy):

  CPU cores (avg):       2.47 → 1.23    (-50%)
  Memory inuse:          277 GB → 186 GB (-33%)
  Gzip-decode CPU path:  38.9% → 0% (eliminated from the top 25)
  RelayCacheSet.Marshal: 30 GB → 19 GB  (-37%)

Tradeoff: upstream bandwidth goes up because nodes now send plain JSON on the wire for smart-router traffic. For co-located upstreams this is free; geographically-distant deployments trade CPU for bytes. Per-request deadlines are unchanged (caller context via NewRequestWithContext); the client's 5-minute DefaultHTTPTimeout is a safety backstop above any realistic upstream timeout.

Covered by:
  - TestHTTPDirectRPCConnection_UsesSharedOptimizedTransport: regression lock-in that two HTTPDirectRPCConnection instances share the singleton common.SharedHttpTransport() for pooling and TLS session reuse.
  - TestHTTPDirectRPCConnection_AdvertisesAcceptEncodingIdentity: httptest end-to-end assertion that both SendRequest and DoHTTPRequest send Accept-Encoding: identity on outbound requests and that response bodies are returned untransformed.

# Description

Closes: #XXXX

---

## Author Checklist

I have...

## Reviewers Checklist

I have...

### DIFF
--- a/protocol/lavasession/direct_rpc_connection_test.go
+++ b/protocol/lavasession/direct_rpc_connection_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"net/http"
 	"net/http/httptest"
+	"sync"
 	"testing"
 	"time"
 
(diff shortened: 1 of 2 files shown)


protocol/lavasession/direct_rpc_connection_test.go
avitenzer, line 1: ▎ **Issue:** sendRequestAE and doHTTPRequestAE are written inside the httptest.Server handler (a separate goroutine) and read from the test body with no synchronization. go test -race will report a data race.
  ▎
  ▎ **Suggestion:**
  ▎ var mu sync.Mutex
  ▎ var sendRequestAE, doHTTPRequestAE string
  ▎
  ▎ srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
  ▎     ae := r.Header.Get("Accept-Encoding")
  ▎     mu.Lock()
  ▎     if r.Method == http.MethodPost {
  ▎         sendRequestAE = ae
  ▎     } else {
  ▎         doHTTPRequestAE = ae
  ▎     }
  ▎     mu.Unlock()
  ▎     ...
  ▎ }))
  ▎ // ...
  ▎ mu.Lock()
  ▎ sae := sendRequestAE
  ▎ dae := doHTTPRequestAE
  ▎ mu.Unlock()
  ▎ require.Equal(t, "identity", sae, ...)
  ▎ require.Equal(t, "identity", dae, ...)

avitenzer (approved):

PR #2281, opened 2026-04-28
feat: Added support of backup providers to WS/gRPC

sotskov-do:
# Description

- Adds a two-tier (primary → backup) endpoint cascade to `DirectWSSubscriptionManager` and `DirectGRPCSubscriptionManager`, matching the HTTP backup-provider model already used by `ConsumerSessionManager`.
- Backup endpoints are consulted only when the primary tier is exhausted; sticky-session lookups span both tiers via a unified URL map.
- `CreateSmartRouterEndpoint` now harvests WS/gRPC URLs from `healthyBackupProviders` in addition to `healthyStaticProviders` and wires them into the new manager constructors. Endpoint collection is factored into shared `collectWSEndpoints` / `collectGRPCEndpoints` helpers that log the tier each endpoint came from.

---

## Author Checklist

I have...

## Reviewers Checklist

I have...

### DIFF
--- a/protocol/rpcsmartrouter/direct_grpc_subscription_manager.go
+++ b/protocol/rpcsmartrouter/direct_grpc_subscription_manager.go
@@ -77,9 +77,13 @@ type DirectGRPCSubscriptionManager struct {
 	chainID        string
 	apiInterface   string
 
-	// Upstream gRPC endpoints
-	grpcEndpoints  []*common.NodeUrl
-	endpointsByURL map[string]*common.NodeUrl
+	// Upstream gRPC endpoints — two-tier separation matches HTTP backup model.
+	// Primary tier serves all selections; backup tier is only consulted when
+	// primary is exhausted (analogous to ConsumerSessionManager's
+	// pairing/backupProviders split).
+	grpcEndpoints       []*common.NodeUrl          // Primary tier — selected first
+	grpcBackupEndpoints []*common.NodeUrl          // Backup tier — used only when primary exhausted
+	endpointsByURL      map[string]*common.NodeUrl // Lookup across both tiers (sticky-session uniformity)
 
 	// Endpoint selection (can be nil)
 	optimizer WebSocketEndpointOptimizer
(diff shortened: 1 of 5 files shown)


protocol/rpcsmartrouter/direct_grpc_subscription_manager.go
avitenzer, line 1: **Issue:** Sticky session is written inside selectEndpoint before the connection is verified, and is never cleared on endpoint failure. Once a client has a sticky entry pointing to a primary URL, the
  backup tier cannot be reached even if that primary endpoint dies. This invalidates the cascade for any client beyond their first call.                                                               
  **Suggestion:** Move the dgm.stickySessions[clientKey] = endpoint.Url writes out of selectEndpoint and into createNewSubscription after pool.GetConnectionForStream(ctx) succeeds — same pattern as
  DirectWSSubscriptionManager.startUpstreamSubscription (line 616). Then add a code path that clears dgm.stickySessions[clientKey] when a stream error is observed, so the next selectEndpoint         
  re-cascades.    
                                                                                                                                                                                                       
  // In createNewSubscription, after the stream is successfully established:
  dgm.lock.Lock()                                                                                                                                                                                      
  dgm.stickySessions[clientKey] = endpoint.Url
  dgm.lock.Unlock()                                                                                                                                                                                    
                  
  // On stream error, before retry:                                                                                                                                                                    
  dgm.lock.Lock() 
  delete(dgm.stickySessions, clientKey)                                                                                                                                                                
  dgm.lock.Unlock()

avitenzer, line 1: **Issue:** selectEndpoint takes no ignoredEndpoints param, while WS does. There's no way to express "this primary URL is misbehaving — try another tier." Same shape needed in selectFromTier (line 917)
  and the optimizer call (line 933 currently passes nil unconditionally).                                                                                                                              
  **Suggestion:** Mirror the WS signature: selectEndpoint(ctx, clientKey, ignoredEndpoints map[string]struct{}), plumb through selectFromTier, and pass to optimizer.ChooseProvider. Even if no caller uses
   it today, the asymmetry between WS and gRPC is confusing and blocks future failover work.

avitenzer (commented):
Please update the PR description.

protocol/rpcsmartrouter/direct_grpc_subscription_manager.go
sotskov-do: Done. Replaced with `lavasession.StickySessionStore`.

sotskov-do: This is exactly what I emphasize on daily: currently `ignoredEndpoints` isn't working at all. It always passed to function as `nil` . I made `gRPC` have the same signature of function like `WS`.
First of all, it's better to decide cases when endpoint will be put in ignored and when excluded from it. I think it's better to do in separate task.

avitenzer: Please open a ticket and assign it to Omer

sotskov-do: Done - https://magmadevs.atlassian.net/browse/MAG-1691

PR #2285, opened 2026-05-04
refactor: OTel usage pipeline + project-id rename + reporter-flow removal

nimrod-teich:
## Scope

This PR bundles three independent refactors that all originate from the same consumer-telemetry cutover. Each is reviewable standalone; they ship together because they all touch the same call sites and there's no in-tree dependency that crosses them.

1. **OTel-based usage pipeline** — replaces the Kafka producer and the push-style optimizer-QoS server with a single OTLP/HTTP log sink (two event types: `relay_usage`, `optimizer_qos`). Off by default. Bulk of the diff.
2. **`dapp-id` → `project-id` wire-key rename** — the canonical project-attribution wire key changes in chainlib (HTTP header, gRPC metadata, fiber Locals). Includes a fasthttp buffer-aliasing fix at the same extraction sites.
3. **`Reporter` / `ReportsAddressFlag` removal** — the off-host reporter flow (already disabled in production) is dropped along with its CSV of plumbing (`errorsForReport`, `metrics.Reporter` interface, `consumer_reports_client.go`, `consumer_relayserver_client.go`, two-field signature shrink on `blockProvider` / `ReportProvider`).

---

### Summary
Replaces the custom Kafka producer with an **OpenTelemetry log-based usage emitter** that handles two event types:

- `relay_usage` — one log record per relay, emitted by `RPCConsumerLogs.AddMetricFor*`
- `optimizer_qos` — one log record per (chain, provider) per sampling tick, emitted by `ConsumerOptimizerQoSClient`

Each rpcconsumer / rpcsmartrouter process emits OTLP/HTTP log records to a local OTel collector; the collector handles fan-out to whatever backend(s) the operator chooses (Kafka, S3, ClickHouse, …) via exporter YAML — no consumer code changes to swap destinations.

**Off by default.** With `--usage-otel-enabled=false` (the default), both relay and QoS paths pay one virtual call per event (`NoopUsageSink.{Emit,EmitOptimizerQoS}`, both empty). No SDK setup, no goroutines, no batching, no resource cost. Flip the flag on once your collector is ready.

### How it works

```
┌──────────────────────────────────────────────────────────────────────┐
│  Bare-metal host (one of N)                                          │
│                                                                      │
│  ┌──────────────┐   ┌──────────────┐   ┌──────────────┐              │
│  │ rpcconsumer  │   │ rpcconsumer  │   │ rpcconsumer  │              │
│  │  (chain A)   │   │  (chain B)   │   │  (chain C)   │   ...        │
│  └──────┬───────┘   └──────┬───────┘   └──────┬───────┘              │
│         │ OTLP/HTTP        │ OTLP/HTTP        │ OTLP/HTTP            │
│         │ (relay_usage +   │ (relay_usage +   │ (relay_usage +       │
│         │  optimizer_qos)  │  optimizer_qos)  │  optimizer_qos)      │
│         └──────────────────┼──────────────────┘                      │
│                            ▼                                         │
│                    ┌──────────────────┐                              │
│                    │  OTel Collector  │  one per host                │
│                    │  (you have it)   │  filter by body, route to    │
│                    │                  │  per-event-type exporter     │
│                    └────────┬─────────┘                              │
└─────────────────────────────┼────────────────────────────────────────┘
                              │
            ┌─────────────────┴─────────────────┐
            ▼                 ▼                 ▼
      ┌──────────┐      ┌──────────┐      ┌──────────┐
      │  Kafka   │      │ S3 +     │      │ Click-   │
      │ exporter │      │ Athena   │      │ House    │
      │          │      │ (one     │      │ exporter │
      │          │      │ bucket   │      │          │
      │          │      │ per ev.) │      │          │
      └──────────┘      └──────────┘      └──────────┘
            ▲                 ▲                 ▲
            │                 │                 │
        choose any 1, 2, or all 3 by editing the
        collector's exporters: YAML — no code change
```

The relay and QoS paths' only contact with telemetry is `NoopUsageSink.{Emit,EmitOptimizerQoS}` (off) or `OTelUsageSink.{Emit,EmitOptimizerQoS}` → `BatchLogProcessor.OnEmit` (non-blocking enqueue). Backpressure shows up as dropped events accounted for in the SDK's internal counters.

### Per-event code path when enabled

```
relay_usage:                            optimizer_qos:

relay completes                         sampling tick fires
     │                                       │
     ▼                                       ▼
chainlib.AddMetricFor*                  ConsumerOptimizerQoSClient
(already `go ...` launched)             .sampleAndEmit
     │                                       │
     ▼                                       ▼
RPCConsumerLogs.usageSink.Emit          for each registered optimizer:
     │                                    optimizer.CalculateQoSScores
     ▼                                    ForMetrics → for each
OTelUsageSink.Emit                       (chain, provider):
     │                                    usageSink.EmitOptimizerQoS
     ▼                                       │
BatchLogProcessor.OnEmit                     ▼
(non-blocking enqueue;                  OTelUsageSink.EmitOptimizerQoS
 full queue → SDK drops)                     │
                                             ▼
                                        BatchLogProcessor.OnEmit

       (background, shared by both event types)
                            │
                            ▼
                   BatchProcessor flushes on
                   BatchSize / FlushInterval / Shutdown
                            │
                            ▼
                   OTLP/HTTP → localhost collector
                            │
                            ▼
                   collector exporters (operator's choice)
```

#### Sink layer (new)
- **`UsageEventSink` interface** in `protocol/metrics/usage_sink.go` — `Emit / EmitOptimizerQoS / Stats / Close`. Implementations must be non-blocking.
- **`NoopUsageSink`** — zero-cost default for both event types; the relay and QoS paths each pay one inlinable empty call when telemetry is off.
- **`OTelUsageSink`** — single OTLP/HTTP exporter, `BatchLogProcessor` with configurable queue / batch / flush / timeout. Resource attributes: `service.name` + `service.instance.id` (default `hostname-pid`). Two log bodies routed by the collector: `relay_usage` and `optimizer_qos`.

#### relay_usage
- **`RPCConsumerLogs`** holds a `UsageEventSink` (was `*ConsumerKafkaClient`); `AddMetricFor{Http,WebSocket,Grpc}` calls `usageSink.Emit(NewRelayUsageEvent(data))`.
- Per-relay attributes (flat in the OTel log body): `project, chain, api_interface, api_method, cu, latency_ms, success, cache_hit, is_write, is_archive, is_batch, is_debug_trace, hedge_count, provider, origin`.

#### optimizer_qos
- **`ConsumerOptimizerQoSClient`** is now a pure periodic sampler — drops every per-(chain, provider) counter, running sum, and aggregation helper (NodeErrorRate, SelectionRate, AverageSelectionQoSScore, etc.). Downstream computes rates / averages by `GROUP BY` at query time.
- `UpdatePairingListStake` stays as a side channel: it learns the current chain → provider set, the current epoch, and the current pairing-list stake per provider. No per-epoch history retained.
- On each tick, `sampleAndEmit` asks every optimizer for `CalculateQoSScoresForMetrics` and forwards one OTel log record per (chain, provider). Nil reports from the optimizer are skipped so the in-memory `latest` cache (read by the optional Prometheus scrape endpoint) only ever holds real records.
- Per-sample attributes (flat): `ts, provider, consumer_hostname, consumer_pub_address, chain_id, geo_location, epoch, entry_index, provider_stake`, the WRS-normalized scores (`selection_availability/_latency/_sync/_stake/_composite`), and the weighted contributions (`availability_contribution/latency_contribution/sync_contribution/stake_contribution`).
- The legacy raw EWMA fields (`sync_score, availability_score, latency_score, generic_score`) are not on the wire — they were the optimizer's internal state; the WRS-normalized scores are what actually drive provider selection. Saved 4 floats per record. `OptimizerQoSReport` (the optimizer's output struct) is unchanged.
- Construction trigger: the QoS client is now created when `--optimizer-qos-listen=true` (Prometheus scrape endpoint) **OR** `--usage-otel-enabled=true` (OTel emission), in either order. The two toggles are independent.

### Lifecycle wiring
`usageSink.Close()` is deferred immediately after construction in both `rpcconsumer.Start` and `rpcsmartrouter.Start`. On process exit the `BatchProcessor` drains its pending queue (up to `ExportTimeout`) instead of dropping it.

#### Kafka pipeline (relay_usage)
- `consumer_kafka_client.go` and its test
- All `--relay-kafka-*` flags
- The Kafka-specific fields on `AnalyticsServerAddresses`

#### Push pipeline (optimizer_qos)
- `--optimizer-qos-server-address` flag and `OptimizerQosServerAddressFlag`
- `--optimizer-qos-push-interval` flag (push cadence is now controlled by the OTel BatchProcessor)
- `AnalyticsServerAddresses.OptimizerQoSAddress`
- `endpointAddress` / `interval` parameters on `NewConsumerOptimizerQoSClient`
- `handleOptimizerQoS` HTTP handler (dead — never wired into a route)

### New flags

| Flag | Default | Purpose |
|---|---|---|
| `--usage-otel-enabled` | `false` | master switch (covers both event types) |
| `--usage-otel-endpoint` | `""` | OTLP/HTTP endpoint; empty defers to `OTEL_EXPORTER_OTLP_ENDPOINT` or `localhost:4318` |
| `--usage-otel-insecure` | `true` | skip TLS (expected target is a sidecar collector) |
| `--usage-otel-queue-size` | `50000` | in-memory queue capacity; full → drop |
| `--usage-otel-batch-size` | `1000` | flush trigger by record count |
| `--usage-otel-flush-interval` | `500ms` | flush trigger by elapsed time |
| `--usage-otel-export-timeout` | `10s` | per-batch OTLP export timeout |
| `--usage-otel-service-name` | `lava-rpcconsumer` / `lava-rpcsmartrouter` | OTel service.name resource attribute |
| `--usage-otel-service-instance-id` | `hostname-pid` | OTel service.instance.id; useful when running multiple processes per host |

### Multi-process-per-host deployments

The expected shape — one rpcconsumer process per chain on a bare-metal host, all sharing one local OTel collector — is supported out of the box. The `hostname-pid` default for `service.instance.id` keeps each process individually identifiable downstream without conflating process identity with billing dimensions. **Chain remains an event attribute on both event types**, so aggregation by `chain × project × method` (relay_usage) or by `chain × provider` (optimizer_qos) is just `GROUP BY` at query time, regardless of which process emitted.

If you'd rather encode chain in the instance ID for ops legibility, the flag override is available: `--usage-otel-service-instance-id="$HOSTNAME-eth"`.

### Sample collector exporter config (what the operator owns)

One pipeline per event type — filter by body, route to its own bucket. Each bucket holds one event shape so no `event` discriminator column is needed downstream.

```yaml
receivers:
  otlp:
    protocols:
      http:
        endpoint: 127.0.0.1:4318

processors:
  filter/relay_usage:
    error_mode: ignore
    logs:
      log_record: ['body != "relay_usage"']
  filter/optimizer_qos:
    error_mode: ignore
    logs:
      log_record: ['body != "optimizer_qos"']
  transform/usage:
    error_mode: ignore
    log_statements:
      - context: log
        statements:
          - set(attributes["ts"], UnixNano(time))
          - set(attributes["service_name"], resource.attributes["service.name"])
          - set(attributes["service_instance_id"], resource.attributes["service.instance.id"])
          - set(body, attributes)
  batch:
    send_batch_size: 1000
    timeout: 500ms

exporters:
  awss3/relay_usage:
    s3uploader:
      region: us-east-1
      s3_bucket: lava-relay-usage-mainnet
      s3_partition: hour
    marshaler: body
  awss3/optimizer_qos:
    s3uploader:
      region: us-east-1
      s3_bucket: lava-optimizer-qos-mainnet
      s3_partition: hour
    marshaler: body

service:
  pipelines:
    logs/relay_usage:
      receivers: [otlp]
      processors: [filter/relay_usage, transform/usage, batch]
      exporters: [awss3/relay_usage]
    logs/optimizer_qos:
      receivers: [otlp]
      processors: [filter/optimizer_qos, transform/usage, batch]
      exporters: [awss3/optimizer_qos]
```

The cloud-side counterpart (S3 buckets + Glue tables + Athena workgroup) is in **Magma-Devs/devops#21**. The collector wiring on the bare-metal hosts is in **Magma-Devs/ansible-internal#268**.

### Dependency / collateral changes
- OTel bumped v1.21 → v1.30 (logs API stable in v1.27+).
- gRPC indirectly bumped 1.62 → 1.66.
- Replaced the deprecated `grpc.DialContext` + `WithBlock` in `upstream_grpc_pool.go` with `grpc.NewClient` + a state-walk loop on `Connect()` / `WaitForStateChange` that only returns success at `connectivity.Ready` — preserves the original "fast-fail on unreachable target" semantics (a single `WaitForStateChange(_, Idle)` would have accepted `Idle → TransientFailure` as success).

### API breaking changes (in-tree callers updated)
- `NewRPCConsumerLogs(...)` now takes `UsageEventSink` instead of `*ConsumerKafkaClient`.
- `NewConsumerOptimizerQoSClient(...)` now takes `UsageEventSink` instead of `(endpointAddress, interval)`.
- `AnalyticsServerAddresses.RelayKafka*` fields → `AnalyticsServerAddresses.UsageOTel*`. `AnalyticsServerAddresses.OptimizerQoSAddress` removed.
- `ConsumerOptimizerQoSClient` no longer exposes `Set*` mutator methods (SetRelaySentToProvider, SetNodeErrorToProvider, SetProviderSelected) — dead callers in `consumer_metrics_manager` and `rpcconsumer_logs` removed.

---

## Scope 2: `dapp-id` → `project-id` wire rename

The consumer reads a tenant identifier from every relay request and stamps it onto the per-relay usage event as the `project` attribute. The wire key was historically `dapp-id`; this renames it to `project-id` everywhere:

- HTTP header name: `Project-Id` (case-insensitive on the wire)
- gRPC metadata key: `project-id` (lower-cased per gRPC convention)
- Fiber Locals key for the HTTP-to-websocket handoff: `project-id`
- New constant `chainlib.ProjectIDHeader = "project-id"` is the single source of truth at every call site.

### fasthttp buffer-aliasing fix (same extraction sites)
`fiber.Ctx.Get` returns strings that alias fasthttp's per-request header buffer (zero-copy via unsafe). The buffer is recycled when the request completes and reused for subsequent requests on the same worker. `RelayMetrics.ProjectHash` is enqueued asynchronously into the OTel `BatchLogProcessor` and serialized after the request returns — by then the backing array may already hold another request's headers, prefix-overwriting the project hash. Both `extractDappIDFromFiberContext` and `extractDappIDFromGrpcHeader` now call `strings.Clone` on the extracted value to detach it from the per-request pool.

### Migration
Clients sending the old `dapp-id` header will be attributed under `DefaultDappID` until they update to `project-id`. There is no header-name back-compat shim.

---

## Scope 3: `Reporter` / `ReportsAddressFlag` removal

The off-host reporter (POST JSON to a configured URL on every block) has been disabled in production for some time. This drops the dead plumbing:

- `--reports-server-address` flag and the `ReportsAddressFlag` constant
- `metrics.Reporter` interface, `consumer_reports_client.go`, and its test
- `consumer_relayserver_client.go` (the only `Reporter` implementation)
- `errorsForReport` / `errorsForConsumerSession` plumbing through `ConsumerSessionManager.blockProvider` and `ReportedProviders.ReportProvider` — both signatures shrunk; in-tree call sites updated
- `ConsumerSessionManager.reporter` field, `RPCConsumerServer.reporter` field, and the matching constructor parameters

Provider unresponsiveness is still surfaced through the existing prometheus counters and the on-chain reporting path — this only removes the off-host push side.

---

## Billing semantics — what reaches the OTel pipeline

Every emitted `relay_usage` event carries `cu` from the spec's `api.ComputeUnits` (set in `rpcconsumer_server.go` / `rpcsmartrouter_server.go` immediately before the relay returns), so the downstream pipeline can bill per-request by summing `cu` grouped by `(project, chain, api_method, provider, time_bucket)`. Coverage and policy notes below.

### Coverage by interface × transport

| Listener | File | Emit point |
|---|---|---|
| REST | `chainlib/rest.go` | every HTTP request (lines 306, 376) |
| JSON-RPC over HTTP | `chainlib/jsonRPC.go` | every HTTP request (line 465) |
| TendermintRPC over HTTP (batch + non-batch) | `chainlib/tendermintRPC.go` | every HTTP request (lines 475, 547) |
| JSON-RPC / TendermintRPC over WebSocket — normal relay | `chainlib/consumer_websocket_manager.go` | every relay (added in this PR) |
| WebSocket — subscription START | same file | every `*_subscribe` call |
| WebSocket — subscription delivery (per streamed message) | same file | every message (added in this PR) |
| WebSocket — unsubscribe / unsubscribe_all | `chainlib/consumer_ws_subscription_manager.go` | not emitted (no CU at provider) |
| gRPC (any spec method) | `chainlib/grpc.go` | every call — `grpcproxy.makeProxyFunc` collapses streaming RPCs to one-shot unary semantics, so each gRPC call lands on this emit path regardless of the spec's `stream` annotation |
| UTXO chains (BTC/LTC/DOGE/BCH) | rides the JSON-RPC HTTP path | every HTTP request — `isUTXOFamily` only reshapes responses, not the emit |

### Policy notes for the downstream pipeline

- **Cache hits.** Events emit with `cache_hit=true` AND the full `cu` value. If the billing model excludes cached responses, filter on `cache_hit=false` in the collector / Athena query.

- **Failure attribution.** Relays that fail before reaching the analytics-fill block in `SendRelay` / `SendParsedRelay` emit with `cu=0, success=false`. Relays that fail *after* — e.g., provider returned an error response — emit with the spec's `cu` and `success=false`. If the model is "bill anything that hit a provider", group on `(success=true OR (success=false AND cu>0))`; if it's "bill successes only", filter on `success=true`.

- **Per-subscription-message billing.** Each delivered subscription event records one `relay_usage` with the same `project / chain / api_method / provider / cu` as the original `*_subscribe`, just with a fresh `ts`. A chatty subscription (e.g. `eth_subscribe newPendingTransactions` on Ethereum mainnet) produces one event per delivered message, each charged at the subscribe API's CU. If the billing model is "subscribe once, deliveries free", deduplicate downstream by `(project, chain, api_method, provider, hour_bucket)` and bill the first event only; the per-message emit gives that filter a population to work over rather than hiding subscription cost in a single record.

- **Pre-emit failures.** Requests that fail at parse / rate-limit / auth before `NewRelayAnalytics` runs do NOT emit any `relay_usage`. They represent no CU consumption at the provider, so they're invisible to billing by design. If you need request-count metrics for unparseable traffic, that's a separate Prometheus counter rather than an OTel event.

- **`success=false` events still carry latency and provider attribution.** Useful for SLO dashboards even if they're excluded from the billing aggregate.

---

## Test plan
- [x] `go build ./...` clean
- [x] `go test ./protocol/metrics/ -count=1` passes
- [x] `go test ./protocol/rpcconsumer/ -count=1` passes
- [x] `go test ./protocol/rpcsmartrouter/ -count=1` passes
- [x] `golangci-lint run --config .golangci.yml ./protocol/...` clean
- [x] `usage_sink_test.go` and `otel_usage_sink_test.go` cover: NoopUsageSink zero-cost on both event types, `RelayUsageEvent` fidelity, defaults application, hostname error fallback, nil-safety on every method, `Emit` + `EmitOptimizerQoS` don't panic or block under unreachable endpoint, `Sent` counter advances for both
- [x] `optimizer_selection_score_test.go` updated to the new constructor signature
- [x] Consumer + smart-router lifecycle: `usageSink.Close()` deferred at construction so the BatchProcessor drains pending records on shutdown
- [x] gRPC dial-state loop preserves `WithBlock` semantics (only `Ready` is success; `Shutdown` fails fast; intermediate `Connecting` / `TransientFailure` keep waiting under the connect timeout)
- [x] `sampleAndEmit` skips nil reports from the optimizer (no zero-valued records leaking into the Prometheus scrape cache)


protocol/chainlib/consumer_websocket_manager.go
avitenzer, line 1: **Issue:** Per-subscription-message events carry the full subscribe-API cu. For chatty subscriptions like eth_subscribe newPendingTransactions, this produces tens to hundreds of events per second per subscriber, each
  charged at the same CU. Without a discriminator attribute, downstream dedup must guess based on (project, chain, api_method, provider, hour_bucket), which is fragile if any of those legitimately repeat for
  non-subscription traffic.
  **Suggestion:** Add an explicit attribute so downstream can filter precisely.
  perMessage := subscriptionFields
  perMessage.Timestamp = time.Now()
  perMessage.IsSubscriptionDelivery = true   // new field on RelayMetrics
  go logger.AddMetricForWebSocket(&perMessage, nil, websocketConn)
  Plumb the new flag into RelayUsageEvent and emit as is_subscription_delivery OTel attribute. The first/initial subscribe event keeps is_subscription_delivery=false, so downstream can use WHERE 
  is_subscription_delivery=false for the "subscribe once" billing model.

protocol/metrics/consumer_optimizer_qos_client_test.go
avitenzer, line 1: **Issue:** cu := uint64(10) and requestedBlock := spectypes.LATEST_BLOCK are passed to CalculateQoSScoresForMetrics, which ignores both. Misleading magic numbers.
  **Suggestion:** Either drop the parameters from the interface (preferred — easy because there's only one in-tree caller) or add a comment:
  // cu and requestedBlock are unused by the current CalculateQoSScoresForMetrics
  // implementation; the interface predates this sampler. Kept here only to
  // satisfy the signature.

avitenzer (commented):
otel/log v0.4.0 and sdk/log v0.4.0 are old versions of the packages.

protocol/chainlib/consumer_websocket_manager.go
nimrod-teich: What we do now is -> subscribe event you pay the spec amount (1000CU) and any subsequent websocket push event coming back from the node will cost 10CU

nimrod-teich:
> otel/log v0.4.0 and sdk/log v0.4.0 are old versions of the packages.

@avitenzer I upgraded to the latest possible without touching the GO version.

sotskov-do (approved):

PR #2306, opened 2026-05-28
fix(chainlib): raise fasthttp ReadBufferSize to 128 KiB

nimrod-teich:
## Summary

- Set Fiber's `ReadBufferSize` / `WriteBufferSize` to 128 KiB at the single Fiber init site (`createAndSetupBaseAppListener` in `protocol/chainlib/common.go`) — every chain listener (JSON-RPC, REST, Tendermint, gRPC HTTP) flows through this.
- Add a network-level regression test that goes over a real TCP socket and verifies an 8 KiB XFCC header gets 200 instead of 431.

## Why

Fiber's default `ReadBufferSize` is 4096 bytes ([fiber/v2 app.go](https://github.com/gofiber/fiber/blob/v2.52.12/app.go#L458)). It maps directly to fasthttp's `Server.ReadBufferSize`, which caps the combined request line + headers. Above that, fasthttp emits **HTTP 431** before the request reaches any handler.

In mTLS deployments, the upstream proxy (Envoy in our case) forwards the client cert via `X-Forwarded-Client-Cert` (XFCC). The PEM-encoded cert plus any chain routinely runs 4–80 KiB, so every mTLS-forwarded relay hits the 4 KiB ceiling and gets a 431.

Repro before the fix:

```bash
# XFCC ~4101 bytes
curl -s -o /dev/null -w "HTTP %{http_code}\n" -X POST \
  -H 'X-Forwarded-Client-Cert: ...' --data '{...}' \
  http://127.0.0.1:3000/
HTTP 431
```

After the fix, the same request returns 200.

128 KiB was chosen to match the ceiling on the corresponding Envoy Gateway `ClientTrafficPolicy` (separate, downstream change), so the two hops agree. Same number is also being applied in the sibling smart-router repo.

## Test plan

- [x] `go build ./protocol/chainlib/...` — passes
- [x] `go test ./protocol/chainlib/ -count=1` — full chainlib suite passes (10s)
- [x] New test `TestCreateAndSetupBaseAppListener_HandlesLargeHeaders` — both subtests (`small_headers_succeed`, `large_xfcc_header_succeeds_after_buffer_bump`) pass
- [x] Verified the regression test actually fails when the production fix is reverted (returns 431, with a diagnostic message pointing at `ReadBufferSize`)
- [ ] Smoke-test against a chain listener with a real >4 KiB XFCC after rollout

### DIFF
--- a/protocol/chainlib/common.go
+++ b/protocol/chainlib/common.go
@@ -460,6 +460,11 @@ func createAndSetupBaseAppListener(cmdFlags common.ConsumerCmdFlags, healthCheck
 	app := fiber.New(fiber.Config{
 		JSONEncoder: json.Marshal,
 		JSONDecoder: json.Unmarshal,
+		// Fiber's default is 4 KiB, which mTLS X-Forwarded-Client-Cert headers (PEM-encoded
+		// certs, often with a chain) routinely exceed — fasthttp then 431s before any handler
+		// runs. 128 KiB matches the corresponding Envoy Gateway ClientTrafficPolicy ceiling.
+		ReadBufferSize:  128 * 1024,
+		WriteBufferSize: 128 * 1024,
 	})
 	app.Use(favicon.New())
 	applyResponseCompression(app, cmdFlags.ResponseCompression)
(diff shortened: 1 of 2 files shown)


nimrod-teich:
Closing — wrong repo; the fix belongs in Magma-Devs/smart-router (PR #59), not lavanet/lava.

PR #2307, opened 2026-05-31
feat: Added host.name to OTel traces

sotskov-do:
### What
Attach the host's name as a resource attribute (`host.name`) to every OTel span emitted by `lavap` (consumer, provider, and smart router). This lets us map collected traces back to
the region a consumer/provider runs in.

### How
- `protocol/tracing/trace_manager.go`: factored resource construction into `newResource()` and added OTel's built-in host detector (`resource.WithHost()`), which sets the standard
`host.name` attribute from `os.Hostname()`. Enabled unconditionally — no flag, no opt-in.
- Ordered `WithFromEnv()` **last** so operator config wins on conflicts: `OTEL_SERVICE_NAME` overrides the in-code service name, and `OTEL_RESOURCE_ATTRIBUTES=host.name=...`
overrides the detected hostname (e.g. to label a clean region instead of a pod hostname).

### Override example
OTEL_SERVICE_NAME=lava-consumer
OTEL_RESOURCE_ATTRIBUTES=host.name=eu-west-1-consumer-3,deployment.environment=production

### Tests
- `TestNewResource_HostName` — default resource carries `os.Hostname()`.
- `TestNewResource_HostNameOverridableViaEnv` — `OTEL_RESOURCE_ATTRIBUTES` overrides the detected hostname.
- Added `OTEL_RESOURCE_ATTRIBUTES` to `clearTracingEnv` so the suite stays hermetic.

No protocol-surface or config changes; standard OTel SDK behavior only.

### Example
image

### DIFF
--- a/protocol/tracing/trace_manager.go
+++ b/protocol/tracing/trace_manager.go
@@ -135,10 +135,7 @@ func New(ctx context.Context, cfg TraceConfig) (*TraceManager, error) {
 	// limit on the same spans, so we cannot disagree with the SDK.
 	bodyAttrLimit = sdktrace.NewSpanLimits().AttributeValueLengthLimit
 
-	res, err := resource.New(ctx,
-		resource.WithAttributes(semconv.ServiceName(cfg.DefaultServiceName)),
-		resource.WithFromEnv(), // OTEL_SERVICE_NAME, OTEL_RESOURCE_ATTRIBUTES
-	)
+	res, err := newResource(ctx, cfg)
 	if err != nil {
 		return nil, utils.LavaFormatError("failed to create OTel resource", err)
 	}
@@ -178,6 +175,21 @@ func New(ctx context.Context, cfg TraceConfig) (*TraceManager, error) {
 		provider: tp,
 		shutdown: tp.Shutdown,
 	}, nil
+}
+
+// newResource builds the OTel resource describing this service. WithHost() adds
+// the standard `host.name` attribute (the OS hostname) so traces can be mapped
+// to the host's region. WithFromEnv() is applied LAST so operator-supplied env
+// config wins on conflicts: OTEL_SERVICE_NAME overrides DefaultServiceName, and
+// OTEL_RESOURCE_ATTRIBUTES (e.g. host.name=eu-west-1) overrides the detected
+// hostname. resource.New merges detectors in order, with later ones taking
+// precedence on key collisions.
+func newResource(ctx context.Context, cfg TraceConfig) (*resource.Resource, error) {
+	return resource.New(ctx,
+		resource.WithAttributes(semconv.ServiceName(cfg.DefaultServiceName)),
+		resource.WithHost(),    // host.name from os.Hostname()
+		resource.WithFromEnv(), // OTEL_SERVICE_NAME, OTEL_RESOURCE_ATTRIBUTES (override)
+	)
 }
 
 // newNoop installs a noop TracerProvider as the global and returns a TraceManager
--- a/protocol/tracing/trace_manager_test.go
+++ b/protocol/tracing/trace_manager_test.go
@@ -217,6 +217,7 @@ func clearTracingEnv(t *testing.T) {
 		"OTEL_EXPORTER_OTLP_ENDPOINT",
 		"OTEL_EXPORTER_OTLP_TRACES_ENDPOINT",
 		"OTEL_SERVICE_NAME",
+		"OTEL_RESOURCE_ATTRIBUTES",
 		"OTEL_TRACES_SAMPLER",
 		"OTEL_TRACES_SAMPLER_ARG",
 		"OTEL_TRACES_EXPORTER",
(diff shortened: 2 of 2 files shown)


avitenzer (commented):
## Review: `feat: Added host.name to OTel traces`

Reviewed the change end-to-end and verified the two load-bearing OTel-SDK behaviors (detector merge precedence and schema-URL handling) against the `go.opentelemetry.io/otel/sdk v1.38.0` source. The full `protocol/tracing` suite passes, and `go vet` / `gofmt` are clean.

**Verdict:** clean, well-scoped, well-documented. The mechanism is correct. One robustness edge case is worth considering plus a few trivial notes — none blocking.

### What's done well

- **The precedence comment is accurate and earns its place.** Detector merge-ordering in `resource.New` is genuinely non-obvious. Confirmed: `detect` folds each detector via `Merge(res, r)` and `Merge(a, b)` lets `b` win on key collisions, so later detectors win — `WithFromEnv()` being last is what gives operators the override. `TestNewResource_HostNameOverridableViaEnv` exercises this.
- **No schema-URL conflict — checked, because it's the classic footgun here.** Only `WithHost()` contributes a non-empty schema URL (`semconv.SchemaURL`); `WithAttributes` and `WithFromEnv` produce schemaless resources, and merging empty-with-non-empty never triggers `ErrSchemaURLConflict`. Verified empirically (`newResource` returns no error in the tests).
- **The hermeticity fix is correct and important.** Adding `OTEL_RESOURCE_ATTRIBUTES` to `clearTracingEnv` stops the override test from leaking `host.name=eu-west-1-consumer` into sibling tests. Easy to forget.
- **The `newResource` extraction is the right move** — it makes resource construction unit-testable without standing up the full exporter/SDK pipeline.

### Potential issue (worth considering, non-blocking)

**Adding `WithHost()` introduces a new, environmental trigger for a fatal startup path.**

`resource.WithHost()` calls `os.Hostname()`. On error, the SDK's `stringDetector.Detect` returns a plain error (not `ErrPartialResource`); `detect` then joins it and continues **with the error still set**, so `resource.New` returns a *usable partial resource* **plus a non-nil error**. That error flows up through `newResource` → `tracing.New`, and all three call sites abort startup:

```go
// rpcconsumer.go:684, rpcprovider.go:900, rpcsmartrouter.go:1589
traceManager, err := tracing.New(ctx, traceCfg)
if err != nil {
    return err   // returned from cobra RunE → the lavap binary fails to start
}
```

The point isn't probability — `os.Hostname()` essentially never fails on Linux (realistically only `ENAMETOOLONG`, a >64-byte hostname, which normal pod/host names won't hit). The point is **asymmetry**: before this PR, `newResource` could only fail on operator-malformed `OTEL_RESOURCE_ATTRIBUTES` (a config error). This adds the first *environmental, non-config* way for it to fail — and the consequence is a node that won't boot, with a `failed to create OTel resource` message that gives no hint the culprit is a hostname quirk. Since the PR makes this unconditional, the blast radius is every binary, all to protect an observability label that should never be able to take a node down.

**Suggested fix (preferred): log-and-continue.** This also matches OTel's guidance to treat `resource.New` errors as non-fatal. You already get a usable partial resource back, so in `New()`:

```go
res, err := newResource(ctx, cfg)
if err != nil {
    // resource.New returns a usable partial resource even on error (e.g. if
    // os.Hostname() fails). Warn and proceed rather than aborting node startup
    // over a missing observability attribute.
    utils.LavaFormatWarning("OTel resource detection incomplete; continuing", err)
}
```

(`res` is always non-nil here — `resource.New` allocates it before detection and mutates in place.) Keeps `WithHost()`, no new imports.

**Narrower alternative:** detect the hostname yourself with `os.Hostname()` and add it via `WithAttributes` only on success, dropping `WithHost()` — a failed lookup is then silently skipped and `newResource` keeps its old config-error-only failure profile. Touches only the new path, but the log-and-continue version is more robust since it also covers any future detector.

### Minor / informational

- **Schema-URL side-effect (benign):** before this PR the resource's schema URL was empty; after, it becomes the SDK's bundled semconv version (from `WithHost()`). Harmless — the attribute keys are stable across semconv versions — but backends keying on schema URL will now see one. Worth a note, not a change.
- **semconv version mix (cosmetic):** the code pins `semconv/v1.21.0` for `ServiceName` while `WithHost()` uses the SDK's newer internal semconv for the host attribute. Not a bug, just slightly inconsistent.
- **Test DRY:** both new tests repeat the same `for … range res.Attributes()` scan; a small `getAttr(res, "host.name") (string, bool)` helper would tighten them. `TestNewResource_HostNameOverridableViaEnv` also asserts only the value, not presence — a dropped attr would give a slightly less clear `"" != "eu-west-1-consumer"` failure.
- **Coverage gap (optional):** no test asserts that `service.name` and `host.name` coexist, nor that `OTEL_SERVICE_NAME` override still works now that a second detector is in the chain. Low value if service-name override is covered elsewhere.

### Summary

| Area | Assessment |
|---|---|
| Correctness | ✅ Precedence + schema-URL behavior verified against SDK source & tests |
| Conventions | ✅ Matches the file's env-driven, well-commented style |
| Tests | ✅ Pass; hermetic; minor DRY/coverage polish possible |
| Risk | ⚠️ One new (low-probability, high-asymmetry) fatal startup path via `os.Hostname()` |

I'd suggest the log-and-continue tweak before merge, but it's a judgment call — the change is otherwise solid.

protocol/tracing/trace_manager.go
```diff
@@ -135,10 +135,7 @@ func New(ctx context.Context, cfg TraceConfig) (*TraceManager, error) {
...
 	bodyAttrLimit = sdktrace.NewSpanLimits().AttributeValueLengthLimit
 
-	res, err := resource.New(ctx,
-		resource.WithAttributes(semconv.ServiceName(cfg.DefaultServiceName)),
-		resource.WithFromEnv(), // OTEL_SERVICE_NAME, OTEL_RESOURCE_ATTRIBUTES
-	)
+	res, err := newResource(ctx, cfg)
 	if err != nil {
 		return nil, utils.LavaFormatError("failed to create OTel resource", err)
 	}
```
avitenzer, line 141: **`os.Hostname()` failure is now fatal to node startup.**

`newResource`'s new `WithHost()` call reads `os.Hostname()`. If that fails, the OTel SDK returns a *usable partial resource* together with a non-nil error — but this block turns that into `return nil, err`, which propagates through `tracing.New` to the `return err` in all three cobra `RunE` callers (`rpcconsumer.go:684`, `rpcprovider.go:900`, `rpcsmartrouter.go:1589`), so the binary fails to start.

Low-probability — on Linux `os.Hostname()` realistically only fails with `ENAMETOOLONG` — but it's a new *environmental* trigger for a path that was previously config-error-only, and a node refusing to boot over a missing observability label is disproportionate. OTel's own guidance is to treat `resource.New` errors as non-fatal. Since `res` is non-nil even on error, log-and-continue with the partial resource:

```suggestion
	if err != nil {
		// resource.New returns a usable partial resource even on error (e.g. if
		// os.Hostname() fails). Warn and proceed rather than aborting startup
		// over a missing observability attribute.
		utils.LavaFormatWarning("OTel resource detection incomplete; continuing", err)
	}
```

sotskov-do: Fatal path here is intentional and consistent with the exporter init directly above: `autoexport.NewSpanExporter` already aborts startup on failure. 
Contract is binary: `tracing enabled -> must fully initialize`.
Supported way to make it optional is `OTEL_SDK_DISABLED=true` / `OTEL_TRACES_EXPORTER=none`. Log-and-continue would also silently swallow malformed `OTEL_RESOURCE_ATTRIBUTES`, turning a fail-loud config contract into best-effort, which I assume isn't expected. 
Note also that `os.Hostname()` doesn't fail on long hostnames. Golang Linux impl falls back to a 512-byte read from `/proc/sys/kernel/hostname`: so the only real trigger is `/proc` being inaccessible, an environmental condition where failing loud (with tracing explicitly on) is the consistent behavior.

avitenzer (approved):

PR #2309, opened 2026-06-01
fix: Avalanche chains names capital case

sotskov-do:


nimrod-teich (approved):

sotskov-do:
Changes couldn't pass `ValidateSpec`, because `spec name must contain lowercase characters only`.

PR #2310, opened 2026-06-01
feat(rpcconsumer): provider whitelist restricting relays per chain and geolocation

avitenzer:
## Summary

Adds an operator-controlled **provider whitelist** to `rpcconsumer` that restricts which providers the consumer is willing to relay to, keyed by `(provider, chain, geolocation)`.

- **List present (configured + loaded):** the consumer relays **only** to whitelisted providers for its chain **and its own region** (`--geolocation`).
- **List absent (flag empty, or not yet loaded):** the consumer keeps its **current relay behavior** (relay to any paired provider).

The list is a JSON document, refreshed **hourly** (interval configurable) in the background, sourced from **GitHub/GitLab or a local file**. When the flag is empty, no refresh loop runs at all.

### Why geolocation (MAG-1987)

The Cosmos pilot needs per-region control — *"1 spec per geolocation and per chain"* — so a top-performing provider can be whitelisted for one region (e.g. EU) and blocked from others. Because Lava's on-chain geolocation pairing filter is disabled, a consumer's pairing already spans all regions; this client-side filter is what pins each per-region gateway to its chosen providers.

## Configuration

```
--providers-whitelist-config            local JSON file path OR a GitHub/GitLab directory URL
--providers-whitelist-refresh-interval  default 1h
--providers-whitelist-token             dedicated token for a remote source in a different repo;
                                         falls back to --github-token / --gitlab-token when empty
```

GitHub/GitLab fetching reuses the **exact same machinery as specs** (URL parsing, contents-API directory listing, authenticated raw download, timeouts). A remote source lists **one directory** and loads only the **top-level `.json` files** in it (subdirectories and non-json files ignored); their `chains` maps are unioned.

### JSON schema

Keyed **chain → geolocation → providers**:

```json
{
  "chains": {
    "ETH1": { "EU": ["lava@1abc...", "lava@1def..."], "AS": ["lava@1ghi...", "lava@1abc..."] },
    "LAV1": { "EU": ["lava@1abc..."] }
  }
}
```

Region keys are the on-chain codes parsed by `planstypes.ParseGeoEnum`: a name (`EU`, `AS`, `AF`, `AU`, `USC`/`USE`/`USW`), a raw bitmask (`2`), a comma-combined set (`EU,AS`), or `GL` (any region). Addresses are matched exactly (on-chain bech32). The same provider may appear under multiple regions.

## Geolocation matching

Each per-chain CSM queries the whitelist with its own `ChainID` **and the consumer's `--geolocation`**. A `GL` bucket matches any region; otherwise the consumer's geolocation must overlap the region bitmask (which reduces to equality for a single-region gateway). This is a **per-region-gateway** model — production runs one consumer per region (e.g. EU = `--geolocation 2`), each relaying only to the providers listed for its region.

## Design notes

- **`ProviderWhitelist`** (`lavasession`): immutable index (`chain → geo → provider set`) behind an `atomic.Pointer`, so the per-relay `IsAllowed` hot path is lock-free; the hourly refresh swaps the pointer atomically. Passthrough until the first successful load.
- **Selection paths guarded (4):** optimizer, header-select, and sticky session (all via the single `validAddresses` filter), plus blocked-provider recovery (per-candidate guard — the list refreshes on its own clock, so a provider whitelisted when blocked can be de-listed before recovery). *Note: the emergency backup-provider fallback is not whitelist-filtered.*
- Filtering is applied at the **selection call site**, not inside `CalculateAddonValidAddresses`, so an empty intersection degrades to a clean `PairingListEmptyError` rather than triggering `validAddresses` reset loops (covered by a regression test).
- **specfetcher refactor** is behavior-preserving: `FetchAllSpecs` layers on `FetchAllRawFiles`; existing spec behavior is unchanged.
- **Token resolution:** `--providers-whitelist-token` wins when set; otherwise falls back to `--github-token` / `--gitlab-token` by provider (testable `resolveTokenForSource`).
- **Fail-closed semantics:** startup fetch failure retries on a short interval until the first success; transient/malformed refreshes keep the last-known-good list; an empty (but loaded) list is logged loudly and relays to nobody.
- **Old-schema rejection:** an old top-level `{"providers":[...]}` file is now rejected with a **hard error** (not silently skipped), so an un-migrated file can't drop the consumer into passthrough.

## Migration

The schema changed from `{"providers":[...]}` (chain-only) to `{"chains":{...}}` (chain → geolocation → providers). Old-format files are rejected loudly — the whitelist file must be converted to the new structure and rolled out **together** with this build.

## Scope

`rpcconsumer` only. `rpcsmartrouter` shares the selection code but is injected no whitelist, so it stays in current behavior.

## Docs

`docs/provider-whitelist.md` covers purpose, flags, JSON schema, region codes, the per-region-gateway model, repo layout, GitHub-like-specs fetching, the guarded selection paths, migration, and fail modes.

## Testing

- **Value type:** parse/empty/malformed, hit/miss, per-chain **and per-geolocation** isolation, same-provider-in-multiple-regions, `GL` wildcard, named-vs-numeric region keys, invalid-region rejection, old-schema rejection, union + skip non-conforming across files.
- **CSM filter:** exclusion across the guarded selection paths, filter-by-consumer-geolocation, a `GetSessions` end-to-end test, and an empty-intersection "no reset storm" regression test.
- **Fetcher:** local-file load, last-known-good on malformed refresh, passthrough when the source is missing, and token resolution.
- **specfetcher:** `FetchAllRawFiles` over a mocked GitHub API, and a spec-path parity test.
- `go build ./...`, `go vet`, gofmt, and the `lavasession` / `rpcconsumer` / `specfetcher` suites all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -645,7 +645,7 @@ func New(
 	// If evidence needs to be handled for the app, set routes in router here and seal
 	app.EvidenceKeeper = *evidenceKeeper
 
-	govConfig := govtypes.Config{MaxMetadataLen: 4000} // 1640 TODO fix it from spec test proposal
+	govConfig := govtypes.Config{MaxMetadataLen: 8000} // 1640 TODO fix it from spec test proposal
 	govKeeper := govkeeper.NewKeeper(
 		appCodec, keys[govtypes.StoreKey], app.AccountKeeper, app.BankKeeper,
 		app.StakingKeeper, app.MsgServiceRouter(), govConfig,
--- a/protocol/lavasession/provider_whitelist.go
+++ b/protocol/lavasession/provider_whitelist.go
@@ -1,0 +1,185 @@
+package lavasession
+
+import (
+	"encoding/json"
+	"fmt"
+	"sync/atomic"
+
+	"github.com/lavanet/lava/v5/utils"
+)
+
+// providerWhitelistJSON is the JSON schema of the consumer provider whitelist, as fetched from
+// GitHub or a local file. It is intentionally unrelated to the spec schema
+// (specutils.SpecAddProposalJSON): the whitelist is a flat list of providers and the chains
+// each provider is allowed to serve.
+//
+//	{
+//	  "providers": [
+//	    { "address": "lava@1abc...", "chains": ["ETH1", "LAV1"] },
+//	    { "address": "lava@1def...", "chains": ["NEAR"] }
+//	  ]
+//	}
+//
+// Providers is a pointer so we can distinguish a document that omits the "providers" key (not a
+// whitelist file at all -> rejected) from one that sets it to an empty array (an intentional
+// "allow nobody" whitelist -> accepted).
+type providerWhitelistJSON struct {
+	Providers *[]providerWhitelistEntry `json:"providers"`
+}
+
+type providerWhitelistEntry struct {
+	Address string   `json:"address"`
+	Chains  []string `json:"chains"`
+}
+
+// whitelistData is the immutable, lookup-optimized index built from the parsed JSON:
+// chainID -> set of allowed provider addresses.
+type whitelistData struct {
+	byChain map[string]map[string]struct{}
+}
+
+func (wd *whitelistData) isAllowed(chainID, providerAddr string) bool {
+	addrs, ok := wd.byChain[chainID]
+	if !ok {
+		return false
+	}
+	_, ok = addrs[providerAddr]
+	return ok
+}
+
+// pairCount returns the number of distinct (chain, provider) pairs, used for logging.
+func (wd *whitelistData) pairCount() int {
+	count := 0
+	for _, addrs := range wd.byChain {
+		count += len(addrs)
+	}
+	return count
+}
+
+// ProviderWhitelist is a thread-safe, hourly-refreshed allowlist that restricts which providers
+// the consumer is willing to relay to, per chain. A single instance is shared across all
+// per-chain ConsumerSessionManagers; each manager queries it with its own chainID.
+//
+// Semantics:
+//   - Until the first successful load it is "not loaded": IsAllowed returns true (passthrough),
+//     preserving the consumer's current behavior when no whitelist exists.
+//   - Once loaded, IsAllowed returns true only for (chainID, providerAddr) pairs present in the
+//     list. A loaded-but-empty list therefore allows nobody (intentional whitelist semantics).
+//
+// The active snapshot is held in an atomic.Pointer so the read hot path (IsAllowed, called
+// per-candidate-provider per-relay) takes no locks, and the hourly refresh swaps it atomically.
+type ProviderWhitelist struct {
+	data atomic.Pointer[whitelistData]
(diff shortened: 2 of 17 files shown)


app/app.go
```diff
@@ -645,7 +645,7 @@ func New(
 	// If evidence needs to be handled for the app, set routes in router here and seal
 	app.EvidenceKeeper = *evidenceKeeper
 
-	govConfig := govtypes.Config{MaxMetadataLen: 4000} // 1640 TODO fix it from spec test proposal
+	govConfig := govtypes.Config{MaxMetadataLen: 8000} // 1640 TODO fix it from spec test proposal
```
sotskov-do, line 648: What is the idea behind this change?

protocol/lavasession/provider_whitelist.go
```diff
@@ -0,0 +1,204 @@
...
+	return nil
+}
+
+// store atomically swaps in a new snapshot and logs the result. An empty (but loaded) whitelist
+// is logged loudly because it causes the consumer to relay to no providers.
+func (pw *ProviderWhitelist) store(data *whitelistData) {
+	pw.data.Store(data)
+	pairs := data.pairCount()
+	if pairs == 0 {
+		utils.LavaFormatWarning("provider whitelist loaded but EMPTY - the consumer will relay to NO providers", nil)
```
sotskov-do, line 203: Is this an expected case? For what reason `consumer` could configure empty whitelist?

protocol/rpcconsumer/provider_whitelist_fetcher.go
```diff
@@ -0,0 +1,169 @@
...
+
+	// Initial load: retry quickly until the first success so the consumer doesn't relay in
+	// passthrough for a whole refresh interval if the source is briefly unavailable at startup.
+	if !f.loadOnce(ctx) {
+		retryInterval := providerWhitelistRetryInterval
+		if f.refreshInterval > 0 && f.refreshInterval < retryInterval {
+			retryInterval = f.refreshInterval
+		}
+		retry := time.NewTicker(retryInterval)
+		defer retry.Stop()
```
sotskov-do: This ticker could be stopped earlier, when `initialLoad` loop finished. No need to keep it live after this.

app/app.go
avitenzer: I had to increase it for the testing to work because the script loads all the specs.

protocol/lavasession/provider_whitelist.go
avitenzer: It make sense that if we have an allow list with no entries we deny everyone. If we don't have a list than we allow everyone.

protocol/rpcconsumer/provider_whitelist_fetcher.go
avitenzer: fixed

sotskov-do (dismissed):

sotskov-do (approved):

PR #2314, opened 2026-06-18
feat(consumer): per-chain latency cutoff for random provider selection

EliasiOfir:
## Summary
Adds an optional per-chain latency cutoff for the consumer's random provider
selection. Providers whose measured QoS latency exceeds a configurable
`max-provider-latency` (seconds) are put aside during general random selection,
so slow providers stop receiving traffic while faster ones absorb it. Opt-in:
`max-provider-latency: 0` (or omitted) preserves current behaviour exactly.

### Behaviour
- Enabled (`> 0`): providers measured above the threshold are excluded from the general random pool.
- Safety fallback: if every provider is above the threshold, the full pool is kept (we never strand a chain).
- Cold start: providers with no latency measurement yet are not filtered.
- Unaffected: stickiness, stateful, and explicitly-selected-provider paths.

## Changes
- `protocol/lavasession/consumer_session_manager.go` — latency filtering in the general selection path.
- `protocol/lavasession/consumer_types.go` — `RPCEndpoint.MaxProviderLatency` field.
- `protocol/rpcconsumer` — parse `max-provider-latency` from the endpoints YAML (defaults to 0).
- `config/` — example consumer configs (enabled / disabled variants) + `rpcconsumer.yml` field.
- `protocol/rpcconsumer/README.md` — docs.
- Tests: unit + integration tests, plus a live E2E harness under `scripts/`.

## Testing
- New unit/integration tests: 14/14 pass (13 lavasession + 1 rpcconsumer).
- E2E harness (`scripts/test/e2e_latency_cutoff.sh all`): all 3 modes pass
  - `cutoff`: slow provider put aside (deltas `p1=+46 p2=+54 p3(slow)=+0`)
  - `regression-disabled`: cutoff off → slow provider still selected (`p3=+24`)
  - `regression-fallback`: all slow → fallback keeps full pool (`+100/+100/+100`)
- Full suite: `go build ./...` clean; no regressions across `x/...` and `protocol/...`.

### DIFF
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -1153,7 +1153,7 @@ func (csm *ConsumerSessionManager) filterHighLatencyProviders(ctx context.Contex
 		// safety fallback: every candidate is over the cutoff - keep the full pool
 		utils.LavaFormatDebug("latency cutoff skipped, all providers above threshold",
 			utils.LogAttr("maxProviderLatency", maxLatency),
-			utils.LogAttr("filteredCount", len(slowProviders)),
+			utils.LogAttr("slowProviders", slowProviders),
 			utils.LogAttr("chainId", csm.rpcEndpoint.ChainID),
 			utils.LogAttr("GUID", ctx),
 		)
@@ -1167,7 +1167,7 @@ func (csm *ConsumerSessionManager) filterHighLatencyProviders(ctx context.Contex
 	if len(slowProviders) > 0 {
 		utils.LavaFormatTrace("latency cutoff applied",
 			utils.LogAttr("maxProviderLatency", maxLatency),
-			utils.LogAttr("filtered", slowProviders),
+			utils.LogAttr("slowProviders", slowProviders),
 			utils.LogAttr("chainId", csm.rpcEndpoint.ChainID),
 			utils.LogAttr("GUID", ctx),
 		)
--- a/protocol/lavasession/consumer_session_manager.go
+++ b/protocol/lavasession/consumer_session_manager.go
@@ -1032,6 +1032,9 @@ func (csm *ConsumerSessionManager) getValidProviderAddresses(ctx context.Context
 		for k, v := range ignoredProvidersList {
 			ignoredProvidersListCopy[k] = v
 		}
+		// Put aside providers whose QoS latency exceeds the per-chain cutoff (general random selection only).
+		// Rides on the ignored-providers set so the optimizer never picks them.
+		csm.filterHighLatencyProviders(ctx, validAddresses, ignoredProvidersListCopy)
 		for i := 0; i < wantedProviders; i++ {
 			provider, selectionStats := csm.providerOptimizer.ChooseProviderWithStats(ctx, validAddresses, ignoredProvidersListCopy, cu, requestedBlock)
 			if len(provider) == 0 {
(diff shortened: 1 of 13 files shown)


protocol/lavasession/consumer_types.go
```diff
@@ -252,6 +252,10 @@ type RPCEndpoint struct {
 	TLSEnabled      bool   `yaml:"tls-enabled,omitempty" json:"tls-enabled,omitempty" mapstructure:"tls-enabled"`
 	HealthCheckPath string `yaml:"health-check-path,omitempty" json:"health-check-path,omitempty" mapstructure:"health-check-path"` // health check status code 200 path, default is "/"
 	Geolocation     uint64 `yaml:"geolocation,omitempty" json:"geolocation,omitempty" mapstructure:"geolocation"`
+	// MaxProviderLatency is the per-chain QoS latency cutoff (in seconds) used during the general random
+	// provider selection. Providers whose EWMA latency exceeds this value are put aside and not picked.
+	// 0 (default) disables the cutoff. Static, header-selected, sticky and stateful selections are unaffected.
+	MaxProviderLatency float64 `yaml:"max-provider-latency,omitempty" json:"max-provider-latency,omitempty" mapstructure:"max-provider-latency"`
```
avitenzer, line 258: The doc says the cutoff is "in seconds," but the value compared in `filterHighLatencyProviders` (`qos.Latency`) isn't wall-clock seconds — it's the QoS **EWMA latency score**, which weights samples by a CU factor (`WithLatencyCuFactor`, `provider_optimizer.go:683`): `1.0` for low-CU relays down to `0.01` for high-CU ones (`score_config.go:44-46`).

Practical effect: a ~2s-real-latency provider on a high-CU method stores an EWMA around `~0.02`, so `max-provider-latency: 1.0` would never fire there — while on low-CU `jsonrpc` (factor `1.0`, like the `eth_blockNumber` E2E) it behaves as ~real seconds. So the same config value means very different things across chains/methods, which an operator reading "seconds" won't expect.

**Requested change:** update this comment and the README to state it's the CU-weighted QoS latency score (not wall-clock), note it's capped at `30s` (`WorstLatencyScore`), and advise tuning per chain rather than as a universal time bound.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -1107,6 +1110,70 @@ func (csm *ConsumerSessionManager) getValidProviderAddresses(ctx context.Context
...
+		}
+		if latency > maxLatency {
+			slowProviders[providerAddress] = struct{}{}
+			continue
+		}
+		remaining++
+	}
+
+	if remaining == 0 {
+		// safety fallback: every candidate is over the cutoff - keep the full pool
```
avitenzer, line 1153: This fallback only triggers when `remaining == 0`, and cold-start providers (no QoS yet) increment `remaining`. So if all *measured* providers are slow but one unmeasured provider exists, the fallback does **not** engage — the slow ones are filtered and that single cold-start provider absorbs the entire batch until it earns a measurement. `TestFilterHighLatencyProviders_MixedColdStartSlowFast` locks this in.

That may be acceptable, but the README implies the fallback protects against pool collapse generally, which isn't quite true here.

**Requested change:** add one line to the README documenting this case (all-measured-slow + ≥1 cold-start → traffic funnels to the cold-start provider, no fallback). If that's *not* intended, let's discuss counting cold-start providers separately so the fallback can still engage.

protocol/lavasession/consumer_session_manager.go
```diff
@@ -1107,6 +1110,70 @@ func (csm *ConsumerSessionManager) getValidProviderAddresses(ctx context.Context
...
+		utils.LavaFormatDebug("latency cutoff skipped, all providers above threshold",
+			utils.LogAttr("maxProviderLatency", maxLatency),
+			utils.LogAttr("filteredCount", len(slowProviders)),
```
avitenzer: In this branch nothing is filtered (pool kept), but the log emits `filteredCount: len(slowProviders)`, which reads as if those providers were excluded.

**Requested change:** rename to `wouldHaveFiltered` (or drop the field) so the log doesn't imply filtering that didn't happen.

avitenzer (commented):
A few semantics/docs notes from review (no correctness blockers). One overarching point that didn't fit a single line:

**On "parse `max-provider-latency`" in the PR description:** there's no parsing code — `UnmarshalKey` + the `mapstructure` tag wire it up automatically (`rpcconsumer.go:567`), which is fine. But note the CLI positional-endpoint form (`ParseEndpointArgs` with `Yaml_config_properties`) likely doesn't support this field, only the config-file form does. Could you confirm whether CLI-arg endpoints need to support it, and if not, document that `max-provider-latency` is config-file-only?

Inline comments below.

protocol/lavasession/consumer_session_manager.go
EliasiOfir: default latency is 0.01 sec.
so it seems very reasonable for new providers to be considerd as "fast" by default.

protocol/lavasession/consumer_session_manager.go
EliasiOfir: fair, changed

avitenzer (approved):

PR #2321, opened 2026-08-06
feat: add grpcwebcompat package for canonical trailer encoding

EliasiOfir:
# Description

Adds a new `grpcwebcompat` package that wraps gRPC servers to serve gRPC-Web requests with canonically encoded trailer frames, fixing compatibility issues with strict gRPC-Web clients.

## Problem

The `github.com/improbable-eng/grpc-web` library encodes gRPC-Web trailer frames using `http.Header.Write()`, which emits headers as `"key: value\r\n"` with optional whitespace after the colon. While RFC 7230 allows this optional whitespace (OWS), receivers that naively split on `:` without trimming read `grpc-status` as `" 0"` (with leading space), causing parsing failures:

```
transport: malformed grpc-status: strconv.ParseInt: parsing " 0": invalid syntax
```

This surfaces as a Trailers-Only response with an Internal error and no message for the caller.

## Solution

The new `grpcwebcompat` package provides:

1. **`WrapServer()`** - Wraps a gRPC server for gRPC-Web with canonical trailer encoding
2. **`WrapHandler()`** - Routes gRPC-Web requests through a custom response writer while passing other requests to a fallback handler
3. **Canonical trailer encoding** - Emits trailers as `"key:value\r\n"` (no whitespace after colon) matching Envoy's reference implementation
4. **Value sanitization** - Strips surrounding whitespace and replaces CR/LF characters that would corrupt frame boundaries
5. **gRPC-Web-text support** - Handles base64-encoded payloads for text format requests
6. **Comprehensive test coverage** - Tests for both binary and text formats, error cases, and native gRPC passthrough

## Changes

- **New file**: `protocol/grpcwebcompat/grpcwebcompat.go` - Core implementation with request/response handling
- **New file**: `protocol/grpcwebcompat/grpcwebcompat_test.go` - Comprehensive test suite (255 lines)
- **Modified files**: Updated imports in `grpcproxy.go`, `chain_tracker.go`, `connection_cmd.go`, and `provider_listener.go` to use the new package instead of direct `grpcweb` dependency

The implementation maintains full compatibility with native gRPC requests and CORS pre-flights by delegating non-gRPC-Web requests to the wrapped handler.

## Test Coverage

- Trailer frame encoding without optional whitespace
- Error status encoding in trailer frames
- Trailers-Only responses (failures before body write)
- Native gRPC request passthrough
- Both `application/grpc-web` and `application/grpc-web-text` formats
- Field value sanitization (whitespace trimming, CR/LF replacement)

https://claude.ai/code/session_01ThUAjPPzmZ4wqSJcoyXPRm

### DIFF
--- a/protocol/chainlib/grpcproxy/grpcproxy.go
+++ b/protocol/chainlib/grpcproxy/grpcproxy.go
@@ -6,8 +6,8 @@ import (
 	"strings"
 
 	"github.com/gofiber/fiber/v2"
-	"github.com/improbable-eng/grpc-web/go/grpcweb"
 	"github.com/lavanet/lava/v5/protocol/common"
+	"github.com/lavanet/lava/v5/protocol/grpcwebcompat"
 	"github.com/lavanet/lava/v5/utils"
 	"go.opentelemetry.io/contrib/instrumentation/google.golang.org/grpc/otelgrpc"
 	"golang.org/x/net/http2"
@@ -47,16 +47,16 @@ func NewGRPCProxyWithReflection(cb ProxyCallBack, healthCheckPath string, cmdFla
 	s := grpc.NewServer(serverOpts...)
 	grpc_health_v1.RegisterHealthServer(s, health.NewServer())
 
-	wrappedServer := grpcweb.WrapServer(s)
+	wrappedServer := grpcwebcompat.WrapServer(s)
 
 	// Create a separate gRPC server for reflection with standard protobuf codec
 	// This is needed because the main proxy server uses RawBytesCodec which breaks
 	// the reflection service's protobuf message serialization
-	var wrappedReflectionServer *grpcweb.WrappedGrpcServer
+	var wrappedReflectionServer http.Handler
 	if reflectionCallback != nil {
 		reflectionGrpcServer := grpc.NewServer(serverReceiveMaxMessageSize)
 		RegisterReflectionProxy(reflectionGrpcServer, reflectionCallback)
-		wrappedReflectionServer = grpcweb.WrapServer(reflectionGrpcServer)
+		wrappedReflectionServer = grpcwebcompat.WrapServer(reflectionGrpcServer)
 	}
 
 	handler := func(resp http.ResponseWriter, req *http.Request) {
--- a/protocol/chaintracker/chain_tracker.go
+++ b/protocol/chaintracker/chain_tracker.go
@@ -14,8 +14,8 @@ import (
 	"github.com/dgraph-io/ristretto/v2"
 	rand "github.com/lavanet/lava/v5/utils/rand"
 
-	"github.com/improbable-eng/grpc-web/go/grpcweb"
 	"github.com/lavanet/lava/v5/protocol/common"
+	"github.com/lavanet/lava/v5/protocol/grpcwebcompat"
 	"github.com/lavanet/lava/v5/protocol/lavasession"
 	"github.com/lavanet/lava/v5/protocol/metrics"
 	"github.com/lavanet/lava/v5/utils"
@@ -582,7 +582,7 @@ func (ct *ChainTracker) serve(ctx context.Context, listenAddr string) error {
 	}
 	s := grpc.NewServer()
 
-	wrappedServer := grpcweb.WrapServer(s)
+	wrappedServer := grpcwebcompat.WrapServer(s)
 	handler := func(resp http.ResponseWriter, req *http.Request) {
 		// Set CORS headers
 		resp.Header().Set("Access-Control-Allow-Origin", "*")
--- a/protocol/grpcwebcompat/grpcwebcompat.go
+++ b/protocol/grpcwebcompat/grpcwebcompat.go
@@ -50,15 +50,15 @@ const trailerFrameFlag = 1 << 7
 // WrapServer wraps a gRPC server for gRPC-Web the same way grpcweb.WrapServer does,
 // except that gRPC-Web responses get a canonically encoded trailer frame.
 func WrapServer(server *grpc.Server, options ...grpcweb.Option) http.Handler {
-	return WrapHandler(server, grpcweb.WrapServer(server, options...))
+	return wrapHandler(server, grpcweb.WrapServer(server, options...))
 }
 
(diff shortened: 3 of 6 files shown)


EliasiOfir:
Pushed 009fcc8 addressing the CI failures and the review findings.

**Fixed**

- **lint** — all three findings: `rawCodec`'s two forced type assertions (`forcetypeassert`) now check `, ok`, and `http.CanonicalHeaderKey` is gone from the key of `Header.Set`, which already canonicalises (`gosimple` S1035).
- **CodeQL `go/reflected-xss`** — the response `Content-Type` was built by rewriting the value grpc-go derived from the request, so the request could steer what landed in the response header. It is now assembled from this package's constants, with the request only *selecting* between them, and set via `Header.Set` rather than by assigning into the header map. A sub-type outside `proto`/`json` collapses to the base content type instead of being echoed. Covered by `TestResponseContentType_NotReflectedFromRequest`.
- **Test naming / exported surface** — tests renamed to the `TestComponent_Scenario` form AGENTS.md asks for and the remaining ones made table-driven; `WrapHandler` and `IsGrpcWebRequest` unexported, since `WrapServer` is the only API used outside the package.

**Not changing, with reasons**

**Case-sensitive gRPC-Web detection.** This is deliberate, and making it case-insensitive on its own would break the requests it aims to fix rather than fix them. `isGrpcWebRequest` is character-for-character the predicate in `grpcweb.WrappedGrpcServer.IsGrpcWebRequest`, and that is load-bearing: whatever this predicate does *not* claim is served by the fallback. The two must agree exactly. If this one matched `Application/Grpc-Web+Proto` but the rewrite in `toNativeGrpcRequest` (a case-sensitive `strings.Replace`) did not, the content type would reach grpc-go unrewritten and be rejected as an invalid gRPC content type — a request that currently just falls through to the same rejection. Making mixed-case gRPC-Web actually *work* means normalising detection and rewrite together, which newly accepts requests that neither this code nor the library accepts today. That is a behaviour expansion beyond this PR's scope. I've added a `mixed case` row to `TestIsGrpcWebRequest_MethodAndContentType` pinning the current behaviour and a comment explaining why.

**Trailer values lose whitespace.** This one is incorrect — trimming is parity with the previous behaviour, not a change. The library serialises the trailer block with `http.Header.Write`, and Go's implementation already applies `textproto.TrimString` to every value ([`net/http/header.go`](https://github.com/golang/go/blob/master/src/net/http/header.go)). Verified directly:

```go
h := make(http.Header)
h["x-padded"] = []string{"  padded  "}
h.Write(&buf)   // => "x-padded: padded\r\n"
```

So a gRPC-Web caller never received the surrounding whitespace before this PR either. The difference the review spotted between the gRPC-Web and native-gRPC assertions in the tests is real, but it predates this change and is inherent to gRPC-Web putting trailers through an HTTP/1 header block. The CR/LF replacement is not optional: an unescaped CRLF in a trailer value would terminate one trailer early and inject another, corrupting the frame.

---

EliasiOfir:
**CodeQL is still red on 009fcc8, and I don't think the remaining alert should be fixed in code. Flagging for a maintainer decision rather than pushing another change.**

**Where it points.** `go/reflected-xss`, sink `w.wrapped.Write(b)` in `responseWriter.Write` — the gRPC-Web response body.

**Why my last commit didn't clear it.** I removed the genuinely attacker-influenced part (the response `Content-Type` was being rewritten from the request's, so a request could steer it — worth fixing on its own merits, and done). But the query's sanitiser is "the content type is statically provable as non-HTML," and it still can't prove that here: the value is assigned from a struct field (`w.contentType`) rather than a literal, and the `Header().Set` happens on `w.wrapped` while the sink is `w.wrapped.Write`, so the two aren't associated through the interface field. Safe value, unprovable to the query.

**Why I believe it's a false positive.** XSS requires the browser to interpret the response as HTML. This body is length-prefixed protobuf frames served as `application/grpc-web` / `application/grpc-web-text` — never HTML, and content sniffing does not upgrade a well-formed non-HTML type. The tainted data the query is tracking is real (the URL path reaches the body via grpc-go's `grpc-message` on an unknown-method error, and handler responses echo request data), but it lands in a frame payload that only a gRPC-Web client parses.

**This dataflow is not new.** It is exactly what `improbable-eng/grpc-web` has always done; CodeQL doesn't analyse module dependencies, so moving those ~130 lines into the repo is what made a pre-existing behaviour visible to the scanner. Nothing about the response changed here.

**What I'd suggest**, in order of preference:

1. **Dismiss the alert as a false positive** in Security → Code scanning. Needs maintainer permissions — I can't do it from here.
2. **Inline suppression** in the source, if you prefer the justification to live next to the code. The repo has no existing CodeQL suppression convention, so this would set one.
3. **Restructure to satisfy the query** — I'd rather not. It means writing literal content types at the sink and reshaping the writer purely to make a taint tracker happy, changing real behaviour (the `+proto` / `+json` sub-type would have to go) for no security benefit. I also can't run CodeQL locally, so each attempt costs a full CI cycle of guessing.

Separately, and independent of the alert: I'm happy to add `X-Content-Type-Options: nosniff` to these responses. That's the correct defence-in-depth for a proxy emitting a non-HTML body and it's one line — I just wouldn't expect it to change CodeQL's verdict.

Tell me which way you want to go and I'll implement it.

---

EliasiOfir:
Correcting my previous comment: **option 2 (inline suppression) isn't viable.** GitHub code scanning has no comment-based suppression mechanism — the [docs](https://docs.github.com/en/code-security/code-scanning/managing-code-scanning-alerts/resolving-code-scanning-alerts) describe dismissal only through the web UI or the REST API. A `// lgtm[...]` / `// codeql[...]` comment would be a no-op that leaves the check red while looking like a fix, so I'm not pushing one.

That leaves dismissal via Security → Code scanning (maintainer permissions, which I don't have) or restructuring the writer to make the content type statically provable — which I'd still advise against, since it trades real behaviour for a taint tracker's benefit.

Everything else on 009fcc8 is green: **lint now passes**, along with `go`, `Analyze (go)`, `Analyze (python)`, `Analyze (actions)`, `validate-check`, `labeler` and `main`. The `test-*` and `build_docker` jobs are still running. CodeQL is the only failure.

---

nimrod-teich (approved):

PR #2323, opened 2026-08-10
fix(consumer): reject implausible seen block advances

EliasiOfir:
# Description

The consumer takes `seenBlock` from whatever a provider reports as its latest block, ratchets it monotonically, and then sends it to every other provider — where `handleConsistency` makes them bail when they cannot reach it. Nothing validated that number anywhere in the consumer path.

So a single provider reporting a value from outside the chain's numeric domain — a different unit, a bug, or a deliberately inflated number — pins the seen block out of reach and makes every honest provider look hopelessly behind, for the lifetime of the entry. With cache-be in shared-state mode the poisoned value also propagates across consumer processes and survives an hour rather than five minutes.

## The fix

Bound the advance by what the chain can actually produce. A cache entry lives at most `EntryTTL`, so any legitimate advance fits inside the blocks produced in that window:

```
maxPlausibleAdvance = max(blocksWithinTTL * 4, 16)
    where blocksWithinTTL = EntryTTL / averageBlockTime
```

| chain | average block time | bound |
|---|---|---|
| SOLANA | 400ms | 3,000 |
| ETH1 | ~12s | 100 |
| BTC | 10min (fewer than one block per TTL) | 16 |

The bound is deliberately loose — its job is to reject the impossible, not to police normal provider spread or bursty block production. The minimum is deliberately *small*: an allowance sized for a fast chain would leave a slow one accepting jumps of days' worth of blocks, which is exactly what this guard exists to catch.

Two safety properties worth calling out:

* an unknown `averageBlockTime` **disables** the guard rather than gating on a number it cannot justify;
* the baseline is re-read after a cache flush when the first lookup misses. Cache writes are applied asynchronously, and treating a not-yet-visible baseline as absent would skip the guard for precisely the back-to-back updates it exists to police, since relay responses are handled one after another.

`averageBlockTime` was already in scope at the construction site (`rpcconsumer.go`), so this only widens `NewConsistency`.

## Most critical files to review

* `protocol/relaycore/consistency.go` — the guard, its bound, and the flush-and-recheck
* `protocol/relaycore/consistency_test.go` — new tests are table-driven and flush rather than sleep; the pre-existing tests pass unchanged

## Known limits

* **Cold entries.** This bounds *advances*, so an entry with no baseline still adopts whatever it is first told. Closing that needs a cross-provider reference; `LatestBlockEstimator` already keeps per-provider observations and a median, which is the natural source. Deliberately out of scope here.
* **The async test is not a regression lock.** `TestConsistency_GuardAppliesToBackToBackUpdates` asserts the ordering's outcome, but the write buffer drains quickly enough that it does not reliably reproduce the race and it passes without the flush. The flush is what closes the window; the test is documentation of the scenario, and its comment says so.

## Relationship to #2322

Independent and separately reviewable. #2322 (now merged) is provider-side (`chaintracker`, `rpcprovider`); this is consumer-side only. Neither depends on the other — this closes a standing exposure that exists regardless of the Solana work. Verified to merge cleanly on top of #2322.

---

## Author Checklist

* [x] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title (`fix`)
* [x] confirmed `!` in the type prefix if API or client breaking change — not needed; no wire or client behavior changes, and the guard is fail-open
* [x] provided a link to the relevant issue or specification — no issue; found while analysing the rollout risk in #2322
* [x] included the necessary unit and integration tests
* [x] updated the relevant documentation or specification, including comments for [documenting Go code](https://blog.golang.org/godoc)
* [x] confirmed all CI checks have passed — green; the one `test-protocol` failure was the pre-existing `protocol/lavasession` flake and passed on re-run of the identical commit

## Reviewers Checklist

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_014BqbitFAAqDiatYCs12prn

### DIFF
--- a/protocol/integration/protocol_test.go
+++ b/protocol/integration/protocol_test.go
@@ -271,7 +271,7 @@ func createRpcConsumer(t *testing.T, ctx context.Context, rpcConsumerOptions rpc
 		}
 	}
 
-	consumerConsistency := relaycore.NewConsistency(rpcConsumerOptions.specId)
+	consumerConsistency := relaycore.NewConsistency(rpcConsumerOptions.specId, 0)
 	consumerCmdFlags := common.ConsumerCmdFlags{}
 	rpcconsumerLogs, err := metrics.NewRPCConsumerLogs(nil, nil, nil)
 	require.NoError(t, err)
--- a/protocol/relaycore/consistency.go
+++ b/protocol/relaycore/consistency.go
@@ -20,10 +20,13 @@ const (
 	// bursty block production, clock skew and provider spread never trip the guard —
 	// it is meant to reject values that are impossible, not merely surprising.
 	seenBlockAdvanceSafetyFactor = 4
-	// seenBlockAdvanceFloor is always accepted regardless of block time, so chains
-	// slower than EntryTTL per block (where the computed bound rounds to zero) and
-	// freshly-started chains are never gated.
-	seenBlockAdvanceFloor = 1000
+	// minSeenBlockAdvance keeps the bound usable on chains that produce less than one
+	// block per EntryTTL, where the computed bound truncates to zero and would reject
+	// every advance. It is deliberately small: a blanket allowance large enough for a
+	// fast chain would leave a slow one — BTC produces a block every 10 minutes —
+	// accepting jumps of days' worth of blocks, which is exactly what this guard is
+	// supposed to catch.
+	minSeenBlockAdvance = 16
 )
 
 // Consistency interface for managing block consistency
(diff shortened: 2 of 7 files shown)


EliasiOfir:
`test-protocol` failed on `8eb9aeb`. It's the pre-existing `protocol/lavasession` flake, not this change.

Reproduced locally by running the protocol packages concurrently on this branch:

```
--- FAIL: TestCheckAndUnblock_BackupUnblockedWhenHealthy (0.32s)
FAIL	github.com/lavanet/lava/v5/protocol/lavasession	38.004s
ok  	github.com/lavanet/lava/v5/protocol/relaycore	4.469s
```

That is the **same test** that fails on `main` at `abba43e` under the same load, with none of this PR's changes present. `protocol/relaycore` — the only package this PR touches — passes.

I checked the one way this PR could plausibly have caused it: the new flush-and-recheck adds a `cache.Wait()` on the baseline-miss path, which could in principle slow the consumer down. It doesn't — `rpcconsumer` runs 77.9s with the change against 78.1s without, and `relaycore` passes in both isolated and loaded runs.

`lavasession` has at least two timing-sensitive tests that fail intermittently under CPU contention (`TestCheckAndUnblock_BackupUnblockedWhenHealthy` and `TestEndpointSortingFlow`); both pass in isolation. Fuller write-up with the cross-branch evidence is in #2322.

Re-run is currently rejected with `403 This workflow is already running` while other jobs are in flight; I'll retry once the run settles. Worth noting the failure detail never reaches the CI log — `test-protocol` pipes `go test` output to an artifact and the "Print failed test output" step tails 200 lines regardless of outcome, so it prints passing packages. Both that and the flaky tests are pre-existing and worth fixing separately.

---

nimrod-teich (approved):

PR #2325, opened 2026-08-11
fix(specs): correct NEAR/NEART block timing to match 600ms blocks

### LINKED ISSUE #2324: [Bug]: False consistency rejections and degraded QoS on NEAR/NEART
### Is there an existing issue for this?

- [x] I have searched the existing issues

### What happened?

We are observing frequent relay failures and degraded QoS on NEAR and especially NEART.
The dominant errors are:
```
Requested a block that is too new
requested block is too new
does not meet consistency requirements
```
Most failures occur with a small blockGap of 2–3.
Example:
```
requestedBlock: 263319664
latestBlock reported by Lava Chain Tracker: 263319662
blockGap: 2
```
At the same time, the connected NEAR node was already at approximately block 263319667. Therefore, the requested block already existed on the node, but Lava rejected the relay before sending it to the node because its internal Chain Tracker was behind.

The NEAR node was healthy:
```
syncing: false
peers: 35–36
RPC latency: 0.3–4 ms
nearcore errors/panics: none
OOM or I/O errors: none
NTP: synchronized
```
The NEAR/NEART specification currently contains:
```
"block_distance_for_finalized_data": 3,
"blocks_in_finalization_proof": 3,
"average_block_time": 1200,
"allowed_block_lag_for_qos_sync": 8
```
However, the observed NEART block interval is approximately 500–625 ms.

In `handleConsistency`, `average_block_time affects` the probability calculation, and the provider may reject a request when:
```
blockGap > 1 && probabilityBlockError > 0.4
```
Our hypothesis is that average_block_time: 1200 is significantly higher than the actual NEART block interval, causing the probabilistic consistency check to reject valid blocks too aggressively.

The hard block-lag limit does not appear to be responsible because:
```
allowed_block_lag_for_qos_sync × 2 = 16
```
while most failures occur at a gap of only 2–3 blocks.

### Lava Version

lavap v6.0.0

### How to reproduce?

1. Start a NEART provider connected to a fully synchronized NEAR testnet node.
2. Enable the standard Lava consistency checks.
3. Send relays where the consumer has already seen a block that is 2–3 blocks ahead of the provider’s internal Chain Tracker:
```
seenBlock: N+2
requestedBlock: N+2
Chain Tracker latestBlock: N
```
4. Confirm directly through the NEAR RPC that requestedBlock already exists.
5. Observe that the provider rejects the request before forwarding it to the node:
```
Requested a block that is too new
does not meet consistency requirements
```
6. Compare the configured block time with the actual block production rate:
```
Configured average_block_time: 1200 ms
Observed block interval: 500–625 ms
```
A possible specification-level fix would be to change the NEART value to approximately: 
```
"average_block_time": 600
```
It may also be worth avoiding the probabilistic early rejection while: 
```
blockGap <= allowed_block_lag_for_qos_sync
```

EliasiOfir:
# Description

NEAR mainnet moved to ~600ms block production in May 2025 ([optimistic blocks in Nightshade 2.0](https://pages.near.org/blog/blink-and-its-final-near-launches-600ms-blocks-and-1-2s-finality/)), but `average_block_time` in the NEAR/NEART specs has been `1200` since the spec was first added in Dec 2024 (`aacfac2`). The only edits to `near.json` since were shard verifications (Sep 2025) and an endpoint rename (Feb 2026) — neither touched timing. Operators report measuring 500–625ms, so the spec is ~2x off.

This came out of investigating NEAR/NEART operators seeing sustained `Requested a block that is too new` / `does not meet consistency requirements` errors at a `blockGap` of 2–3.

`average_block_time` feeds two paths, and both get more aggressive as the value grows relative to reality.

**1. ChainTracker polling cadence.** `start()`/`updateTimer()` derive the poll interval from `averageBlockTime` (`base/4`, `base/8`, `base/16`), so a 1200ms base polls at 300/150/75ms instead of 150/75/37.5ms. The tracker's latest block trails the node by up to a full extra block, which is what opens the gap in the first place.

**2. `handleConsistency`'s Poisson bail.** `eventRate` is `(timeSinceChange + halfTimeLeft) / averageBlockTime`, and the provider returns `ConsistencyError` when `P(X <= blockGap-1) > 0.4`. Understating the block rate inflates that probability. Evaluated with the repo's own `CumulativeProbabilityFunctionForPoissonDist`:

| average_block_time | blockGap | 10 CU (`block`, `chunk`, `status`) | 20 CU (`query`) |
|---|---|---|---|
| **1200 (current)** | 2 | 0.77–0.86 → bail | 0.62–0.71 → bail |
| **1200 (current)** | 3 | 0.93–0.97 → bail | 0.85–0.90 → bail |
| 600 | 2 | 0.45–0.62 → bail | **0.26–0.36 → pass** |
| 600 | 3 | 0.72–0.85 → bail | 0.50–0.63 → bail |

At 1200 the bail is unconditional for any `blockGap >= 2`, which makes the catch-up wait immediately below it unreachable on NEAR.

`allowed_block_lag_for_qos_sync` is doubled to 16 so the QoS sync tolerance stays equivalent in wall-clock terms (8 blocks x 1200ms == 16 x 600ms). Leaving it at 8 would silently halve the allowed lag from 9.6s to 4.8s as a side effect of the block-time correction.

## Scope

This corrects the spec to match the chain. It is necessary but **not sufficient** on its own — per the table above, `blockGap` of 2 still bails on 10-CU methods and 3 still bails everywhere. The remaining gap is a provider-side issue in `handleConsistency` (the Poisson term models block production, but the block being waited for is always `<= seenBlock` and therefore already exists on chain — the provider is waiting on its own tracker poll, not on block production). That is a separate, independently reviewable change and is deliberately not in this PR.

## Most critical files to review

* `specs/mainnet-1/specs/near.json` — four values, NEAR and NEART

## Impact on relay timeouts

None worth noting. `GetRelayTimeout` only uses `averageBlockTime` for hanging APIs (`extraRelayTimeout = averageBlockTime * 2`), and all four of NEAR's hanging APIs (`broadcast_tx_async`, `send_tx`, `broadcast_tx_commit`, `tx`) set an explicit `timeout_ms` of 5–20s, so the term is a small fraction of their budget.

## Deployment note

Changing the file here does not change the live spec — this needs a governance proposal to take effect on mainnet.

---

## Author Checklist

* [x] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title (`fix`)
* [x] confirmed `!` in the type prefix if API or client breaking change — not applicable; spec values only, no API or wire change
* [x] provided a link to the relevant issue or specification — no issue; raised by NEAR/NEART operators reporting degraded QoS, upstream cause linked above
* [x] included the necessary unit and integration tests — not applicable; no new code paths. `go test ./x/spec/...` passes with the changed values
* [x] updated the relevant documentation or specification — this PR is the specification change
* [ ] confirmed all CI checks have passed — CI has not run yet at the time of opening

## Reviewers Checklist

---
_Generated by [Claude Code](https://claude.ai/code/session_01YCMc3dBY9QkwHkX8H4iK1Q)_

### DIFF
--- a/specs/mainnet-1/specs/near.json
+++ b/specs/mainnet-1/specs/near.json
@@ -11,8 +11,8 @@
                 "data_reliability_enabled": true,
                 "block_distance_for_finalized_data": 3,
                 "blocks_in_finalization_proof": 3,
-                "average_block_time": 1200,
-                "allowed_block_lag_for_qos_sync": 8,
+                "average_block_time": 600,
+                "allowed_block_lag_for_qos_sync": 16,
                 "shares": 1,
                 "min_stake_provider": {
                     "denom": "ulava",
@@ -753,8 +753,8 @@
                 "data_reliability_enabled": true,
                 "block_distance_for_finalized_data": 3,
                 "blocks_in_finalization_proof": 3,
-                "average_block_time": 1200,
-                "allowed_block_lag_for_qos_sync": 8,
+                "average_block_time": 600,
+                "allowed_block_lag_for_qos_sync": 16,
                 "shares": 1,
                 "min_stake_provider": {
                     "denom": "ulava",
--- a/specs/testnet-2/specs/near.json
+++ b/specs/testnet-2/specs/near.json
@@ -1,0 +1,994 @@
+{
+    "proposal": {
+        "title": "Add Specs: Near",
+        "description": "Adding new specification support for relaying Near data on Lava",
+        "specs": [
+            {
+                "index": "NEAR",
+                "name": "near mainnet",
+                "enabled": true,
+                "reliability_threshold": 268435455,
+                "data_reliability_enabled": true,
+                "block_distance_for_finalized_data": 3,
+                "blocks_in_finalization_proof": 3,
+                "average_block_time": 600,
+                "allowed_block_lag_for_qos_sync": 16,
+                "shares": 1,
+                "min_stake_provider": {
+                    "denom": "ulava",
+                    "amount": "5000000000"
+                },
+                "api_collections": [
+                    {
+                        "enabled": true,
+                        "collection_data": {
+                            "api_interface": "jsonrpc",
+                            "internal_path": "",
+                            "type": "POST",
+                            "add_on": ""
+                        },
+                        "apis": [
+                            {
+                                "name": "query",
+                                "block_parsing": {
+                                    "parser_arg": [
+                                        "0",
+                                        "block_id"
+                                    ],
+                                    "parser_func": "PARSE_CANONICAL",
+                                    "default_value": "latest"
+                                },
+                                "compute_units": 20,
+                                "enabled": true,
+                                "category": {
+                                    "deterministic": true,
+                                    "local": false,
+                                    "subscription": false,
+                                    "stateful": 0
+                                },
+                                "extra_compute_units": 0
+                            },
+                            {
+                                "name": "EXPERIMENTAL_changes",
+                                "block_parsing": {
+                                    "parser_arg": [
+                                        "0",
+                                        "block_id"
+                                    ],
+                                    "parser_func": "PARSE_CANONICAL",
+                                    "default_value": "latest"
+                                },
+                                "compute_units": 10,
+                                "enabled": true,
+                                "category": {
+                                    "deterministic": true,
+                                    "local": false,
+                                    "subscription": false,
+                                    "stateful": 0
+                                },
+                                "extra_compute_units": 0
+                            },
+                            {
+                                "name": "block",
+                                "block_parsing": {
+                                    "parser_arg": [
+                                        "block_id",
+                                        "=",
+                                        "0"
+                                    ],
+                                    "parser_func": "PARSE_DICTIONARY_OR_ORDERED",
+                                    "default_value": "latest"
+                                },
+                                "compute_units": 10,
+                                "enabled": true,
+                                "category": {
+                                    "deterministic": true,
+                                    "local": false,
+                                    "subscription": false,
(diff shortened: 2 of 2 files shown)


nimrod-teich (changes requested):
You have to add this to testnet-2 folder as well, and follow-up with a proposal on testnet first and mainnet after.

EliasiOfir:
Done — added in `10ffcb6`.

`near.json` didn't exist under `specs/testnet-2/specs/` at all, so I added it as a byte-identical copy of the mainnet-1 file, matching how the other dual-deployed chains are carried (`bsc.json`, `iota.json`, `casper.json` etc. are the same file in both directories). Both indices it contains, `NEAR` and `NEART`, therefore carry `average_block_time: 600` and `allowed_block_lag_for_qos_sync: 16`.

`x/spec/keeper` `TestSpecs` loads and merges both directories, so the new file is exercised there — it passes.

On the rollout: understood, testnet proposal first and mainnet after. Flagging one thing for whoever runs them, since it affects what you'll see on testnet. This spec correction is necessary but not sufficient on its own — at 600ms a `blockGap` of 2 still trips the provider-side bail on 10-CU methods (`block`, `chunk`, `status`) and a gap of 3 still trips it everywhere, per the table in the description. The rest is #2326, which is provider-side and ships on its own schedule via a `lavap` release. So if the testnet proposal lands first and NEART operators still report some `Requested a block that is too new`, that's expected rather than a sign this didn't take — the error rate should drop, not go to zero.

---

nimrod-teich (approved):

PR #2326, opened 2026-08-11
fix: improve handleConsistency to wait for tracker polls instead of bailing

### LINKED ISSUE #2324: [Bug]: False consistency rejections and degraded QoS on NEAR/NEART

EliasiOfir:
# Description

Fixes #2324

Fixes the `handleConsistency` logic to distinguish between two failure modes:

1. **Provider lag beyond spec tolerance** (a node problem): bail immediately
2. **Tracker hasn't polled yet** (transient): wait for the next poll instead of bailing

The previous implementation used `probabilityBlockError` (which models block production) to decide whether to bail. This caused false positives when the spec's `average_block_time` overstates the chain's real block time, leading to premature bailouts even when the requested blocks already exist on-chain and the tracker would pick them up on its next poll.

The fix gates the early bail on whether there's enough time for at least one tracker poll (`trackerPollInterval`), which is much faster than block production. This allows the wait loop to run and catch up in cases like NEAR where small gaps (2-3 blocks) are transient.

The spec-level half of #2324 was fixed separately in #2325 (NEAR/NEART `average_block_time` 1200 -> 600). That change alone is not sufficient: at 600 a `blockGap` of 2 still bails on 10-CU methods and 3 still bails everywhere. This PR closes the remaining gap.

## Changes

- **rpcprovider_server.go**: Replace `probabilityBlockError > 0.4` condition with `noTimeForATrackerPoll` check that compares available time against `trackerPollInterval`
- **consistency_test.go**: Add two new test cases:
  - `TestHandleConsistency_SmallGapWaitsInsteadOfBailing`: Verifies that small gaps (2-3 blocks) wait instead of bail when the tracker can catch up
  - `TestHandleConsistency_NoTimeToPollBails`: Verifies that we still bail when there's insufficient time for even one tracker poll

---

## Author Checklist

* [x] included the correct [type prefix](https://github.com/commitizen/conventional-commit-types/blob/v3.0.0/index.json) in the PR title
* [x] included the necessary unit tests
* [x] updated comments for code documentation
* [x] confirmed all CI checks have passed

## Reviewers Checklist

https://claude.ai/code/session_01YCMc3dBY9QkwHkX8H4iK1Q


nimrod-teich (dismissed):
LGTM. Rebase to main and push.

EliasiOfir:
Rebased onto `main` (`0a7440f`) and pushed — `0d1fb2f` → `fa15ab3`. Note this was a force-push, so it will have reset the approval.

#2322 landed in the meantime, which made this a real rebase rather than a fast-forward. `rpcprovider_server.go` merged cleanly — that PR's edits to `handleConsistency` are at the top of the function (reading `latestBlock` from `GetWireLatestBlock`) and in the wait loop, while this change is the bail condition between them. `consistency_test.go` conflicted at the insertion point, where both PRs added new fixtures and tests in the same place. Resolved by keeping both; nothing was dropped or rewritten on either side.

The diff is unchanged at +85/−2, and `TestHandleConsistency_ComparesInPublishedDomain` from #2322 passes against the new gate — it exercises a 500-block gap against a `blockLagForQosSync` of 25, so it bails on the far-behind term, which this change leaves untouched.

Verified locally: `go build ./...` clean; all seven tests in `protocol/rpcprovider` consistency pass; `protocol/rpcprovider/...`, `protocol/chaintracker/...` and `protocol/chainlib/...` all green.

One open item, flagging rather than silently carrying it: Qodo's `Conservative poll interval` finding above is unaddressed on purpose. It is correct that `trackerPollInterval` uses `averageBlockTime/4` — `updateTimer`'s slowest branch — while the tracker can poll as fast as `averageBlockTime/16`, so the gate is conservative and can still bail where waiting would have worked. I measured the practical span: for 10-CU methods it changes behaviour on chains with `average_block_time` between 2s and 8s, which is 31 specs (the COSMOSSDK/COSMOSHUB/TENDERMINT/IBC family, TRX, EVMOS, XRP, XLM, CELO, TON, TEZOS, CASPER and others). NEAR is unaffected either way at 600ms. Taking that fix would widen this PR's blast radius well beyond the NEAR-shaped case it was scoped to, so it seemed better as a follow-up than as a late change to an approved PR — happy to fold it in here instead if you'd prefer.

---

EliasiOfir:
`test-protocol` failed on `d3050ff`. Diagnosed as a flake, not this change — re-run triggered.

**What failed.** `protocol/integration` died with `panic: test timed out after 15m0s`, hung in `TestConsumerProviderBasic`. That is also why the Test Results reporter showed `6 files - 1` — the package never wrote its report. The failure detail doesn't reach the CI log, since `test-protocol` pipes `go test` into an artifact and the print step tails a fixed number of lines regardless of outcome, so the log ends on passing packages.

**Why it isn't this PR.**

* `TestConsumerProviderBasic` uses `LAV1`, not NEAR. Neither this PR nor #2325 touched `lava.json`.
* Even so, this change is a no-op for LAV1's timings: `average_block_time` is 15000ms, so `trackerPollInterval` is 3.75s, far above any `halfTimeLeft` in that test — the gate bails exactly where `probabilityBlockError > 0.4` did.
* Both parents of the merge are green in CI: `fa15ab3` (this branch, all 30 checks) and `c4c16fe` (main, post-#2325). The merge itself added only two spec JSON files on top of a green tree; the Go code at `d3050ff` is byte-identical to `fa15ab3`.

A package that is green on both parents and red on a merge that changes no Go code is a flake. `protocol/integration` is the heaviest package in the tree and it shares a single 15m budget with everything else `./protocol/...` runs, so it is the natural one to tip over under runner contention — the same shape as the `protocol/lavasession` flakes seen on #2323.

One caveat on my own evidence, in the interest of not overstating it: I could not reproduce this locally in a way that discriminates. `TestConsumerProviderBasic` hangs for the full 15 minutes in my sandbox at every commit I tried, including plain `main`, because the environment restricts the local servers it stands up. Since `main`'s own CI is green at those same commits, that hang is an artifact of my environment, not a signal about the code. The reasoning above rests on the CI results and the diff, not on the local runs.

If the re-run reproduces it, that changes the picture and I'll dig in properly rather than re-running again.

---

EliasiOfir:
Correcting my previous comment: **the re-run reproduced the failure, so this is not a flake.** `test-protocol` timed out again on `d3050ff` — the `go test` step ran ~16 minutes (05:50 → 06:06) and exited 1, which is the `-timeout 15m` firing. I was wrong to call it a flake and I'm not re-running again.

**What the evidence now says.** `test-protocol` is green on both parents and red only on their merge:

| commit | contents | `test-protocol` |
|---|---|---|
| `fa15ab3` | this change, NEAR spec at 1200/8 | green (30/30) |
| `c4c16fe` | main, NEAR spec at 600/16, without this change | green |
| `d3050ff` | both | **timeout, twice** |

The Go code at `d3050ff` is byte-identical to `fa15ab3`; the merge added only the two NEAR spec files. So the failure is an interaction between this change and #2325, and neither is faulty alone.

**The mechanism, as best I can establish it.** This change replaces an instant bail with entry into the existing wait loop, which sleeps up to `halfTimeLeft`. #2325 raised NEAR's `allowed_block_lag_for_qos_sync` from 8 to 16, which doubles the far-behind threshold `blockLagForQosSync*2` from 16 to 32 — so relays with a gap between 17 and 32 that previously failed instantly now sleep instead. Across an integration suite that runs many relays, added sleeps accumulate against a package budget that was already the tightest in the tree.

That is precisely the fail-fast trade-off noted when this PR was opened. CI is now putting a number on it rather than leaving it theoretical.

**What I have not established, and won't assert.** I cannot reproduce this locally in a way that discriminates — `protocol/integration` cannot run in my environment at any commit, including plain `main`, because it stands up local servers the sandbox blocks. So I know *that* the interaction is real from the CI matrix above, but I have not directly observed *which* test slows or by how much. Everything after "the mechanism" is inference from the diff.

**Where I think this should go.** The most promising fix is to bound the wait by what it is actually waiting for. We established the provider is waiting for one ChainTracker poll, not for block production — so sleeping up to `halfTimeLeft` (500ms on 10-CU methods) is far longer than needed, and is what makes the accumulated cost bite. Capping the sleep near a poll interval would keep the NEAR benefit while cutting the added latency substantially.

That is a design change to a PR that is already approved, and it interacts with Qodo's `base/4` vs `base/16` finding above, which pushes the opposite way. I'd rather not guess at it unverified — flagging for a decision on direction before I push anything further.

---

EliasiOfir:
Found the actual cause, and it is neither this change nor an interaction with #2325. My previous two comments were both wrong; disregard them. Fix pushed as `ba7cc72`.

**The hang is in the test helper, on the first line of the first test.** Goroutine dump from the timeout:

```
integration.isPortInUse(...)                 mocks.go:250
integration.(*uniqueAddressGenerator).GetAddress(...)  mocks.go:302
integration.TestConsumerProviderBasic(...)   protocol_test.go:465
```

Two defects in `protocol/integration/mocks.go` combine into an infinite loop:

1. `isPortInUse` treats **every** error from resolve/listen as "this port is taken", including errors that say nothing about the port. On a host without IPv6 loopback the `[::1]` bind fails with `EAFNOSUPPORT`, so every port reports busy while none of them are.
2. `GetAddress` loops on that, and its `currentPort > maxPort` guard is **after** the loop rather than inside it. When every probe says "in use" the loop never terminates and never reaches the guard — so the intended `panic("all ports have been exhausted")` is unreachable and the package spins until `go test -timeout 15m` kills it. No output, no report file, which is why the reporter showed `6 files - 1` and the log ends on passing packages.

`ba7cc72` fixes both: probe failures meaning the address family is unavailable (`EAFNOSUPPORT`, `EADDRNOTAVAIL`) skip that address rather than condemning the port, and the bound moves inside the loop so exhaustion panics with a usable message instead of hanging.

**Measured, on a host where this reproduces every time:**

| | before | after |
|---|---|---|
| `TestConsumerProviderBasic` | 15m timeout | passes in 2.09s |
| full `protocol/integration` | 15m timeout | green in 37s |

The 37s run was at `d3050ff` exactly — this change plus the merged NEAR spec, the configuration that was failing. So the interaction I hypothesised earlier does not exist: the suite is nowhere near its budget, and the fail-fast trade-off is not costing measurable time here.

**One thing I still cannot prove.** I reproduced deterministically on a host with no IPv6 loopback. I don't know the runner's IPv6 state, so I can't say for certain the CI hang enters the loop by that same door — the unbounded loop is reachable any time probes keep reporting "in use", and `currentPort` is package-global and only ever increments across the whole suite. Both halves of the fix are correct regardless, and the loop bound converts any remaining case from a silent 15m hang into an immediate, named panic. If CI still times out after this, it will at least say why.

Happy to split this into its own PR if you'd rather keep this one to the consistency change — it is independent, and it fixes a latent hang that can take down `test-protocol` on any PR.

---

EliasiOfir:
`test-protocol` is green on `ba7cc72` — 6m12s, against two consecutive 15m timeouts before the fix. So the port-probe loop was the cause on the runner too, not only on my host.

A different job went red on the same commit: `test-consensus`, and it is unrelated to this PR.

```
FAIL	github.com/lavanet/lava/v5/x/conflict/client/cli	600.099s
```

A 10-minute timeout, hung in a cosmos-sdk test-network gRPC `Accept` loop. Two checks that it cannot be this branch:

* the only new file in `ba7cc72` is `protocol/integration/mocks.go`, which is `package integration_test` — a test-only package that is never linked into another package's test binary;
* `go list -deps ./x/conflict/client/cli` returns **zero** dependencies on `lava/v5/protocol`.

`test-consensus` was also green on `fa15ab3`, two commits ago on this same branch.

Worth noting it is the same *class* of problem as the one just fixed — test infrastructure standing up listeners on a constrained runner and hanging rather than failing fast. Different package and different owner, so I'm not touching it here, but `x/conflict/client/cli` timing out at exactly its 10m budget is worth someone's attention independently of this PR.

Re-run of the failed job is currently rejected with `403 This workflow is already running` while the e2e and docker jobs are still in flight; I'll retry once the run settles and report the result either way.

---

